PR #30, opened 2019-01-14
Fixed inverse merge, improved logging during debugging

hohwille:
After refactoring inverse merge (`update-eclipse-workspace-settings`) was broken. Some variables were not substituted anymore. To trace down the error I improved the logging that was not yet helpful enough. Further I discovered that on linux and mac the system resolves links (which are especially required to make eclipse work under macos) and therefore the re-substitution of the paths does not work as expected. I extended the solution to resolve symbolic links and allow also the substitution of the results as far as symlinks are involved.

### DIFF
--- a/configurator/src/main/java/com/devonfw/ide/configurator/Configurator.java
+++ b/configurator/src/main/java/com/devonfw/ide/configurator/Configurator.java
@@ -100,8 +100,36 @@ public class Configurator {
   private VariableResolver createResolver(File variablesFile) {
 
     Properties variables = PropertiesMerger.loadIfExists(variablesFile);
-    variables.put(CLIENT_ENV_HOME_VARIABLE, CURRENT_WORKING_DIRECTORY);
+    putVariable(variables, CLIENT_ENV_HOME_VARIABLE, CURRENT_WORKING_DIRECTORY);
+    putSystemProperty(variables, "java.home");
+    putEnvironmentVariable(variables, "JAVA_HOME");
+    putEnvironmentVariable(variables, "ECLIPSE_HOME");
+    putEnvironmentVariable(variables, "SETTINGS_PATH");
     return new VariableResolverImpl(variables);
+  }
+
+  private static void putSystemProperty(Properties properties, String key) {
+
+    putVariable(properties, key, System.getProperty(key));
+  }
+
+  private static void putEnvironmentVariable(Properties properties, String key) {
+
+    putVariable(properties, key, System.getenv(key));
+  }
+
+  private static void putVariable(Properties properties, String key, String value) {
+
+    if ((value != null) && !value.isEmpty()) {
+      if (value.startsWith("file:")) {
+        value = value.substring(5);
+      }
+      value = VariableResolverImpl.normalizePath(value);
+      properties.put(key, value);
+      Log.LOGGER.fine("Variable '" + key + "' = " + value);
+    } else {
+      Log.LOGGER.info("Variable '" + key + "' is undefined");
+    }
   }
 
   /**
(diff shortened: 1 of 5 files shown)


configurator/src/main/java/com/devonfw/ide/configurator/merge/PropertiesMerger.java
```diff
@@ -93,13 +99,18 @@ public static void save(Properties properties, File file) {
...
+      return;
+    }
+    if (!updateFile.exists()) {
+      Log.LOGGER.finest("Update file does not exist: " + updateFile.getAbsolutePath());
       return;
     }
     Properties updateProperties = load(updateFile);
     Properties workspaceProperties = load(workspaceFile);
-    SortedProperties updatedPropeties = new SortedProperties();
-    updatedPropeties.putAll(updatedPropeties);
```
hohwille: 102 was a stupid bug. Instead of `updatedProperties` it should have been `updateProperties`. To avoid such confusion, I renamed the variable as well.

configurator/src/main/java/com/devonfw/ide/configurator/merge/DirectoryMerger.java
```diff
@@ -94,12 +97,18 @@ public void inverseMerge(File workspaceFile, VariableResolver resolver, boolean
         Log.LOGGER.warning("Workspace is missing directory: " + workspaceFile);
         return;
       }
+      Log.LOGGER.fine("Traversing directory " + updateFile);
       for (File child : updateFile.listFiles()) {
-        inverseMerge(workspaceFile, resolver, addNewProperties, new File(updateFile, child.getName()));
```
hohwille: 98 was a severe bug. For recursion workspaceFile was never changed. However, directory traversal also has to be applied here so the child file or folder has to be passed instead of `workspaceFile`. Hence this code never worked.

configurator/src/main/java/com/devonfw/ide/configurator/logging/Log.java
```diff
@@ -21,11 +25,19 @@
   static {
     try {
       FileHandler logFileHandler = new FileHandler("Configurator.log");
-      logFileHandler.setLevel(Level.ALL);
```
hohwille: java.util.logging is kind of strange. Even though it looks like every level will be logged with this config, nothing below INFO was logged. You also need to tweak the root logger as I do in my fix below.
Also I found the multiline log format very ugly and hard to read so I created my own and use it both for console and for logfile.

configurator/src/main/java/com/devonfw/ide/configurator/Configurator.java
```diff
@@ -100,10 +100,38 @@ private void saveChangesInWorkspace(boolean saveNewProperties) {
   private VariableResolver createResolver(File variablesFile) {
 
     Properties variables = PropertiesMerger.loadIfExists(variablesFile);
-    variables.put(CLIENT_ENV_HOME_VARIABLE, CURRENT_WORKING_DIRECTORY);
```
hohwille: this was the only variable that could be resolved. However, the settings also contain variables like `SETTINGS_PATH` that need to be resolved. I added all variables I thought are required and make sense.
BTW: For most projects I tried, the `variablesFile` (replacement-patterns.properties) does not exist. Maybe it would be better to configure the path to it in the `devon-ide` scripts itself rather than in the settings (`ide-properties.bat` - that also lacks multi-platform support).

hohwille:
I am finally ignoring our 4-eyes workflow as nobody is doing reviews or giving feedback hence we otherwise stall our progress and leave people sitting with bugs...

Issue #31, opened 2019-02-15
redesign, reduce scripts, increase UX


hohwille:
see https://github.com/devonfw/devon-ide/wiki/Redesign

hohwille:
PR #32 has been merged. Hence done. I might do some polishing commits with this issue id but can already close this.

hohwille:
Mhm, scripts in top-level dear are not yet complete and reduced as planned. Hence reopening.

hohwille:
So the only thing left to do woud be:
* create a commandlet(s) to open console (cmd, cygwin, git-bash, iTerm2, powershell, etc.)
* delete current console scripts (console.bat, console-*.bat)
* optionally create OS specific console script on setup
* consider about ConEmu / Commander support (IMHO some optional addon to be explained/scripted in `system/windows` folder).

hohwille:
> So the only thing left to do woud be:
> * create a commandlet(s) to open console (cmd, cygwin, git-bash, iTerm2, powershell, etc.)
Created issue #61 
> * delete current console scripts (console.bat, console-*.bat)
Part of #61 
> * optionally create OS specific console script on setup
Created issue #60 
> * consider about ConEmu / Commander support (IMHO some optional addon to be explained/scripted in `system/windows` folder).
Created issue #62 

Hence I will close this issue that already has a very long history and gets messy.

Issue #27, opened 2019-01-10
Improve handling of environment variables when IDE is configured multiple times


hohwille:
If you setup your IDE (using `source env.sh`, using `IDEenv`, running `console`, etc.) several environment variables are set/modified (e.g. `JAVA_HOME`, `PATH`, etc.). If this happens more than once (maybe even with different parallel IDE project installations) this can get a little messy:
Some environment variables like `PATH` are not entirely overwritten but extended. So e.g. for `PATH` with every call several directories are prepended. So if you have multiple of these IDE environment variable setups you will have a very long and messy path.
It would be awesome to find a solution how we can detect this and replace the previous extensions rather than extending again.
To make it very clear, here is what I mean:
```
➜  devon source env.sh
IDE environment has been initialized.
➜  /projects echo $PATH
/projects/devon/software/maven/bin:/projects/devon/software/java/bin:/projects/devon/software/eclipse::/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/VMware Fusion.app/Contents/Public:/opt/X11/bin
➜  /projects cd mmm
➜  mmm source env.sh
IDE environment has been initialized.
➜  /projects echo $PATH
/projects/mmm/software/maven:/projects/mmm/software/java/bin:/projects/mmm/software/eclipse:/projects/devon/software/maven/bin:/projects/devon/software/java/bin:/projects/devon/software/eclipse::/projects/devon/software/maven/bin:/projects/devon/software/java/bin:/projects/devon/software/eclipse::/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Applications/VMware Fusion.app/Contents/Public:/opt/X11/bin
```

hohwille:
Fixed.
https://github.com/devonfw/devon-ide/wiki/varibales

PR #25, opened 2018-12-19
#21: improve OS support

hohwille:
#23: proper fix for dirname of sourced script

Issue #23, opened 2018-12-17
Proper fix for dirname in env.sh


hohwille:
User @maihacke changed `env.sh` to use `$(dirname "$0")`:
https://github.com/devonfw/devon-ide/commit/b9cca7862995eafbac07000f55eafa5d639019c7#diff-59bbdf4f0007715da1871587ba4ffd3d

I assume this was some kind of bugfix for MacOS. Also this seems more reliable to me as `$BASH_SOURCE` is proprietary and only works in bash. What is the purpose why this was ever changed to this?

User @vapadwal wants to change `env.sh` back to `$(dirname "$BASH_SOURCE")`:
https://github.com/devonfw/devon-ide/commit/985a51ed992edd9d129d963c61e25db6460ba357#diff-59bbdf4f0007715da1871587ba4ffd3d

As I do not like to have git wars, I would kindly ask all participants to clarify the reasons for these changes and find a peaceful solution.

hohwille:
I was able to reproduce the error somehow:
```
Jorgs-MacBook-Pro:devon hohwille$ . env.sh
dirname: illegal option -- b
usage: dirname path
IDE environment has been initialized.
Jorgs-MacBook-Pro:devon hohwille$ echo $0
-bash
```

So in some cases bash has `-bash` as `$0` what is causing the issue.
Hence I think both "fixes" are wrong.
IMHO we need an if that tests if we are in bash and then uses `$BASH_SOURCE` and otherwise uses `$0` to make it work in any shell.

maihacke:
Please try 
. ./env.sh

hohwille:
Create `test.sh` with this content:
```
echo "0=$0"
echo "BASH_SOURCE=$BASH_SOURCE"
echo "_=$_"
````

This will produce:
```
$ . ./test.sh
0=-bash
BASH_SOURCE=./test.sh
_=BASH_SOURCE=./test.sh
```

**Comment by hohwille** [comment]
```
$ zsh
% . ./test.sh
0=./test.sh
BASH_SOURCE=
_=BASH_SOURCE=
```

So in `zsh` the `$0` variant is correct.

hohwille:
`$_` seems to be the last command (even within the current script)

maihacke:
yepp, nonsense :-/

maihacke:
maybe we should go with ${BASH_SOURCE[0]-$0} then
Should work in Bash and ZSH

hohwille:
[...]

PR #32, opened 2019-02-15
#31: introduce devon CLI with plugable commands, increase UX, reduce script confusion

hohwille:
This PR implements #31 improves OS support (#25) and also fixes the following issues:
* #23 
* #27 
* maybe some more I forgot to mention.


hohwille:
This PR is introducing 3rd party libs for `ide-configurator`.
IMHO we should avoid using `maven-shade-plugin` or `maven-assembly-plugin` to generate a fat jar as this has some license implications. Then the JAR contains a mixture of classes from different vendors and with different licenses. Instead we should keep the `ide-configurator` JAR free from 3rd party classes and simply add these as their original JARs to the scripts package release. There we can include a `TERMS_OF_USE` and/or `LICENSE` file explaining the used licenses. Downloading and including the relevant licenses could be automated with the `license-maven-plugin`

hohwille:
To complete this I see the following aspects:
* remove further console variant scripts for windows
* create a command plugin for opening console windows for different OS and types (`devon console` or `devon console cygwin`/`devon cygwin`, ...)
* Add a `setup` script in the toplevel directory that can be double clicked. What would people expect? Should it also open the IDE (IMHO no) or just setup? Should we still create `eclipse-*` and `vscode-*` scripts by default? Or will all users pickup the new CLI approach? IMHO to catch up with our large user-base the toplevel script should create the IDE launch scripts for the main workspace automatically. Then we will have less questions. People can still delete the scripts and use CLI only if they like.
* Adopt IDE scripts to create *.bat files to launch IDE on windows instead of bash scripts (5 minutes of work).
* Integrate devcon?
* Add lots of documentation...

Issue #43, opened 2019-03-25
integrate PDF documentation into build and release


hohwille:
Wiki documentation has been entirely rewritten and is now detailed and improved.
We should be able to generate it as PDF.
This generation should be included into the build process and PDF should be part of the scripts package.

hohwille:
Done.

PR #52, opened 2019-03-27
#43: integrated documentation

hohwille:
Solves #43


hohwille:
How sad. Travis does not seem to support CI friendly maven. Maybe an old version of maven? Need to check .travis.yml.

Issue #41, opened 2019-03-25
Proper way to detect if eclipse workspace is locked


hohwille:
We need to figure out if an eclipse workspace is locked or not
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=197428
* https://stackoverflow.com/questions/33431762/how-to-check-if-eclipse-workspace-is-locked/33432112?noredirect=1#comment96829310_33432112
* https://stackoverflow.com/questions/14266016/why-doesnt-filechannel-trylock-prevent-access-across-multiple-processes

markusschuh:
First challenge: eclipse supports [different lock mechanisms:](https://help.eclipse.org/2019-03/topic/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html?#osgilocking)  
The default -  `-Dosgi.locking=java.nio` seems to be then one, used by devon. (checked on Windows 7). But this may be also "java.io" on macOS / linux.

The good part: If java.nio file locking is used, on windows this uses a file handle, also visible by windows. And Users have the right to read out their filehandles. But that needs a tool to access the kernel - even WMI has no call for the file handles. (what would allow some command line use)

I have tried the check with help of a some java code, using tryLock() of [java.nio.channels.FileLock](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/AsynchronousFileChannel.html#tryLock-long-long-boolean-) but this works.

It would work with external tool "handle.exe" from [sysinternals](https://live.sysinternals.com) - owned by Microsoft. But even then it is nasty to script that with help of windows CMD:

    :: untested code
    set path_lockfile=%WORKSPACE_PATH%\.metadata\.lock
    if exist %path_lockfile% (
      :: findstr needs double backslashes
      set path_lockfile_for_search=%path_lockfile:\=\\%
      for /f "usebackq tokens=1-3,*" %%P in (`handle64 -nobanner -p javaw.exe ^| "%SystemRoot%\system32\bin\findstr" /R "^javaw.exe %path_lockfile_for_search%$"`) do (
        :: not yet realized: handle output like below. 
        :: javaw.exe pid:  <thispid> <domain\username>
        ::  <lockhandle>: File          <filename>
        :: javaw.exe pid:  <anotherjavaw_pid> <domain\username>
        ::javaw.exe pid:  <anotherjavaw_pid> <domain\username>
      )
    )

Meta Code:
Loop over results:
   for lines with token1 == javaw.exe set var javaw_pid
[...]

PR #64, opened 2019-04-01
Review of unix scripts

markusschuh:
Resolve the warnings & implement most important recommendations of [ShellCheck](https://github.com/koalaman/shellcheck)


scripts/src/main/resources/scripts/command/build
```diff
@@ -50,12 +50,12 @@ then
...
 else
   echo "Currently not in a buildable project or build system could not be auto-detected"
-  exit -1
+  exit 255
 fi
 
 if [ -n "${1}" ]
 then
-  BUILD_OPTS="${@}"
+  BUILD_OPTS="${*}"
```
hohwille: I do not know the difference between those two, but I assume you changed this on purpose and consider it as an improvement so if tests work out I am fine with it.

scripts/src/main/resources/scripts/command/build
```diff
@@ -50,12 +50,12 @@ then
...
 fi
 echo "${BUILD_CMD} ${BUILD_OPTS}"
-doDevonCommand $BUILD_CMD ${BUILD_OPTS}
+doDevonCommand $BUILD_CMD "${BUILD_OPTS}"
```
hohwille: I think you introduced a bug here. The `BUILD_OPTS` will allow to provide multiple arguments to the build system. So e.g. for maven I could set `BUILD_OPTS=-DskipTests -Pdeploy`. With your change I assume that both different args would be passed as a single parameter and maven will fail. We could test it to get sure...

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -65,7 +65,7 @@ then
 else
   #mirror="https://mirror.math.princeton.edu"
   mirror="https://ftp.osuosl.org"
-  software_version=`doGetVariable ECLIPSE_VERSION 2018-12`
+  software_version=$(doGetVariable ECLIPSE_VERSION 2018-12)
```
hohwille: shouldnt this be changed to `${ECLIPSE_VERSION:-2018-12}`?

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -84,7 +84,7 @@ else
   if doIsMacOs
   then
     echo "Doing workarounds for MacOS quirks..."
-    echo -e "#!/bin/bash\n${DEVON_IDE_HOME}/software/eclipse/Eclipse.app/Contents/MacOS/eclipse \$@" > "${ECLIPSE_HOME}/eclipse"
+    echo -e "#!/bin/bash\n${DEVON_IDE_HOME}/software/eclipse/Eclipse.app/Contents/MacOS/eclipse "'"$@"' > "${ECLIPSE_HOME}/eclipse"
```
hohwille: I also think this introduces a bug. We are writing a script here and inside the generated script the parameters to eclipse should be `$@`. Anyhow I justed tested this and both versions are not working as exclamation mark needs to be escaped as it seems... But maybe also the latter differes from interactive shell and from script as this used to work...

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -147,7 +147,7 @@ if [ "${run}" != "false" ] || [ -n "${CONFIGURATOR_MODE}" ]
 then
   if [ -e "${WORKSPACE_PATH}/.metadata/.lock" ]
   then
-    echo doFail "Eclipse workspace ${WORKSPACE} is locked.\nIt seems as if Eclipse is already running for this workspace.\nOtherwise consider deleting this file\n${WORKSPACE_PATH}/.metadata/.lock"
+    doFail "Eclipse workspace ${WORKSPACE} is locked.\nIt seems as if Eclipse is already running for this workspace.\nOtherwise consider deleting this file\n${WORKSPACE_PATH}/.metadata/.lock"
```
hohwille: this echo is on purpose because until #41 is properly fixed. So please do not remove it as it will cause a bug.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -158,13 +158,13 @@ fi
...
+ echo -e "#!/bin/bash\ncd \$(dirname \${0})\ncd workspaces/${WORKSPACE}\nsource ${DEVON_IDE_HOME}/scripts/devon eclipse" > eclipse-"$WORKSPACE"
+ chmod a+x eclipse-"$WORKSPACE"
  echo "Created script eclipse-$WORKSPACE"
 fi
 
 if [ "${run}" != "false" ]
 then
   echo "launching Eclipse..."
-  eclipse -clean -data "${WORKSPACE_PATH}" -keyring "~/.eclipse/.keyring" -vm "${JAVA_HOME}/bin/java" -showlocation ${WORKSPACE} -vmargs ${ECLIPSE_VMARGS} &
+  eclipse -clean -data "${WORKSPACE_PATH}" -keyring ~/.eclipse/.keyring -vm "${JAVA_HOME}/bin/java" -showlocation "${WORKSPACE}" -vmargs "${ECLIPSE_VMARGS}" &
```
hohwille: good catch. You fixed a bug here. `~` is not resolved if quoted...

scripts/src/main/resources/scripts/command/gradle
```diff
@@ -25,7 +25,7 @@ then
     echo "Gradle is already installed at ${GRADLE_HOME}"
   fi
 else
-  software_version=`doGetVariable GRADLE_VERSION 5.2.1`
+  software_version=$(doGetVariable GRADLE_VERSION 5.2.1)
```
hohwille: Same as above...

scripts/src/main/resources/scripts/command/gradle
```diff
@@ -34,5 +34,5 @@ if [ "${setup}" = "true" ]
 then
   doRunCommand "gradle -v" "verify installation of gradle"
 else
-  gradle ${@}
-fi
\ No newline at end of file
+  gradle "${@}"
```
hohwille: Again something similar. Maybe I am also mistaken but IMHO quoting provides everything as a single argument, or not? Then this would be a bug.

scripts/src/main/resources/scripts/command/ide
```diff
@@ -36,7 +36,7 @@ function doSetup() {
   do
     echo
     echo "*** Setting up ${tool} ***"
-    doDevonCommand ${tool} setup
+    doDevonCommand "${tool}" setup
```
hohwille: IMHO not needed here but will also not hurt.

scripts/src/main/resources/scripts/command/java
```diff
@@ -15,7 +15,7 @@ function doSetupJava() {
       echo "Java is already installed at ${JAVA_HOME}"
     fi
   else
-    local software_version=`doGetVariable JAVA_VERSION 11.0.2`
+    local software_version=$(doGetVariable JAVA_VERSION 11.0.2)
```
hohwille: again as above.

scripts/src/main/resources/scripts/command/mvn
```diff
@@ -25,7 +25,7 @@ then
     echo "Maven is already installed at ${MAVEN_HOME}"
   fi
 else
-  software_version=`doGetVariable MAVEN_VERSION 3.6.0`
+  software_version=$(doGetVariable MAVEN_VERSION 3.6.0)
```
hohwille: ${MAVEN_VERSION:-3.6.0}

scripts/src/main/resources/scripts/command/ng
```diff
@@ -19,16 +19,17 @@ then
...
   fi
 else
   doRunCommand "npm install -g @angular/cli@latest --unsafe" "install angular-cli"
 fi
 
 if [ "${setup}" != "true" ]
-  ng ${@}
-fi
\ No newline at end of file
+then
```
hohwille: great fix. Thanks.

scripts/src/main/resources/scripts/functions
```diff
@@ -423,10 +423,10 @@ function doInstall() {
   fi
   echo "Starting installation from ${2} to ${1}"
   doDownload "${2}"
-  doExtract ~/Downloads/${2/*\//} "${4}"
+  doExtract ~/Downloads/"${2/*\//}" "${4}"
   if [ -n "${3}" ]
   then
-    doMoveGlobSafe ${DEVON_IDE_HOME}/updates/extracted/${dir} ${DEVON_IDE_HOME}/updates/extracted/${3}
+    doMoveGlobSafe "${DEVON_IDE_HOME}/updates/extracted/${dir}" "${DEVON_IDE_HOME}/updates/extracted/${3}"
```
hohwille: This is wrong. ${3} may be a glob pattern.

hohwille:
I also was not aware of ShellCheck before. Maybe we can even integrate it into travis build so we get warnings/errors if we do not follow best practices. However, only makes sense if it is not too strict.

scripts/src/main/resources/scripts/command/build
markschuh: I should have added the links to related shellsheck error codes into the bode of the commit messages, sorry. This one was [SC2124](https://github.com/koalaman/shellcheck/wiki/SC2124),

The use of $@ here is an implicit cast - the result by using $* does not change here. (It would be different, if IFS were changed - but this is not the case). See [Special parameters and shell variables](https://wiki.bash-hackers.org/syntax/shellvars#special_parameters_and_shell_variables)

Nevertheless, when I reread the change within whole file context now, my conclusion was wrong: It is the BUILD_OPTS var, which need to be an array, no string. At least that were the right use, when space containing options should be handled correctly. A very rare but not impossible case. Imagine e.g. a logfile path parameter in cygwin bash, where some path element contains a space.

We should skip the change here, since that keeps the shellcheck warning alive, something is here not yet perfect. 

By the way: There is a vscode-shellcheck extension

scripts/src/main/resources/scripts/command/build
markschuh: You are completely right. This change is wrong. I'm unsure, why I have did this change, since there is no shellcheck warning for this.

scripts/src/main/resources/scripts/command/eclipse
markschuh: True (as mentioned within #64 - but since this was not a shellcheck finding, I skipped this change in this pull request.

scripts/src/main/resources/scripts/command/eclipse
markschuh: The change is right - if you ever intend to support spaces in options. $@ is an array variable, if it is used without double quotes spaces within different elements will split the option into more than one. If used WITH double quotes the shell will automatically protect those elements by double quotes:   1foo 2bar "3foo bar"

With respect to "!" - indeed, during interactive use this needs to be escaped when in double quotes. It would be better, to do this also within scripts. Anyway I personally would use a HERE document here instead of embedding \n within the string. 
In theory there could be a bash on some strange OS, that only accepts different line endings than a newline. (No, not bash on windows)

scripts/src/main/resources/scripts/functions
markschuh: Ah - so I will move the end quote before the ${3}

markschuh:
Best is to use [ShellCheck](https://www.shellcheck.net/) even before code is commited, e.g. with help of the [shellcheck vscode](https://github.com/timonwong/vscode-shellcheck) extension.

Different [Install](https://github.com/koalaman/shellcheck#user-content-installing) exists, on Windows it could be run native (installed via scoop), inside WSL (win10) or as a docker container.

ShellCheck can be configured to [ignore](https://github.com/koalaman/shellcheck/wiki/Ignore) specific checks, Still missing is an option to ignore via priority. (e.g. don't show recommendations)

[TravisCI](https://github.com/koalaman/shellcheck/wiki/TravisCI) integration exists and shellcheck is meanwhile preinstalled there.

The only "challenge" for devon-ide is, that the shell scripts can't be selected by file ending, but need to be listed all by name. I guess, it would be more helpful for CI/CD and browser syntax highlighting, if the .sh ending were used. But this has new challenges during deployment, when the file endings will be removed there.

scripts/src/main/resources/scripts/functions
hohwille: yep

hohwille:
Awesome, but according to shellcheck I seem to be stupid:
https://github.com/timonwong/vscode-shellcheck/issues/50

scripts/src/main/resources/scripts/command/gradle
markschuh: Quoting of strings or quoting in string context turns the result into a single string. But $@ is a array variable and the standard command line is not a string context. So the result of "$@" is - for arguments ["foo","bar","foo bar"] - again:  "foo" "bar" "foo bar",

see https://wiki-dev.bash-hackers.org/scripting/posparams#mass_usage

scripts/src/main/resources/scripts/command/eclipse
markschuh: I know, that the reason for the log message  (another ShellCheck finding) "Tilde does not expand in quotes" as listed on top of PR.

markschuh:
I have added your CRs ( except the postponed replacement of doGetVariable ). I'm unsure, what happens with this PR if I upgrade to lastet commit in my forked repo, so I do not yet push the merge from my local clone to remote. The conflicts still allow auto resolving - so I assume, the merge of pull request can be handled,

N.B: the vscode-shellsheck is a linter. And as some other linters I have seen - and with the default setting "onType" - it should underline a **new** entry line like `echo $@`. If not, I guess your shellcheck can't be found by the linter. Documentation may not be perfect - but I have seen much, much worser extensions on the vscode marketplace.

hohwille:
Even though this PR itself was not merged and has too many conflicts now, I worked your bug fixes, suggestions and improvements into the code. I will close this PR now. Thanks for your work, experience in bash, bugfixes and hints.

PR #146, opened 2019-07-24
Updated vscode command/script

jambulud:
*devon vscode create-script* command was creating eclipse scripts instead of vscode.

### DIFF
--- a/scripts/src/main/resources/scripts/command/vscode
+++ b/scripts/src/main/resources/scripts/command/vscode
@@ -97,10 +97,10 @@ do
   elif [ "${1}" = "setup" ]
   then
     doSetup setup
-    doCreateIdeScript eclipse
+    doCreateIdeScript vscode
   elif [ "${1}" = "create-script" ]
   then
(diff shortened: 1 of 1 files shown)


scripts/src/main/resources/scripts/command/vscode
```diff
@@ -97,10 +97,10 @@ do
   elif [ "${1}" = "setup" ]
   then
     doSetup setup
-    doCreateIdeScript eclipse
+    doCreateIdeScript vscode
   elif [ "${1}" = "create-script" ]
   then
-    doCreateIdeScript eclipse
+    doCreateIdeScript vscode
```
hohwille: ups, stupid copy&paste bug I made. Thanks for finding and fixing 👍

hohwille:
I am unsure what is wrong with CLA hook. After all changing two words should not end up in a copyright sueing case so I might just ignore this ;)

hohwille:
CLA fixed.

Issue #169, opened 2019-08-02
generic tar extract support (linux/mac setup fails with *.xz files)


sobkowiak:
I have downloaded the latest version of ide scripts from https://repo1.maven.org/maven2/com/devonfw/tools/ide/devon-ide-scripts/3.0.0-beta11/. The command `setup` has problem with installation of nodejs. The script cannot handle files with extension `*.xz`

````
*** Setting up node ***
Starting installation of node.js to /home/ksobkowiak/work/tmp/devon-ide-scripts-3.0.0-beta11/software/node from https://nodejs.org/dist/v10.16.0/node-v10.16.0-linux-x64.tar.xz
Artifact already exists at /home/ksobkowiak/Downloads/node.js-v10.16.0-linux.tar.xz
To force update please delete the file and run again.

******** ATTENTION ********
ERROR: Unknown archive format: .xz. Can not extract /home/ksobkowiak/Downloads/node.js-v10.16.0-linux.tar.xz
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
ERROR: command 'node setup' failed with exit code 255
````

Category: Bug

Severity: High

Steps to reproduce (In case of bug):

- Unpack the latest version of `devon-ide-scripts-3.0.0-beta11.tar.gz` and run the `setup` script

Environment - Fedora 30, devon-ide 3.0.0-beta11

markusschuh:
I'd suggest to add compatibility to use xz format for the archives - but at the same time I recommend to use gz as far as possible - and add this recommendation to the documentation. For example, even nodejs for linux x64 is still available in tar.gz archive format. I just needs a switch to  "All download options" to locate the related URL.

The issue is just, that "tar" isn't identical on all platforms - and it is a bit difficult to detect / ensure compatibility of tar with xz. To my knowledge
- macOS uses "bsdtar" and has xz compatibility built in since about 10.9.0 or so
- most other devon-ide supported enviroments use GNU tar, which delegates any compression/decompression task to external utilities, which is "xz" in this case. "xz-utils" (or similar packages) are today typically part of default installations - but this is no must. Earlier or later somebody runs devon-ide on a system without xz installed.

Therefore use of tar.gz is still the most compatible one - and many software is still distributed this way in addition to tar.xz as another option. XZ is just the default, since this "extended compress" is a bit more efficient than gz in terms of both comressed size as well as compress/decompress times.

hohwille:
From @markusschuh in #170 
> A simple
[...]

PR #170, opened 2019-08-02
#169: linux/mac setup fails with *.xz files

sobkowiak:
The `functions` file has been extended by handling the `.xz` files with `tar xf file`

### DIFF
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -355,7 +355,7 @@ function doExtract() {
     doRunCommand "tar xf '${1}' -C '${target_dir}'"
   elif [ "${ext}" = ".xz" ]
   then
-    doRunCommand "tar xf '${1}' -C '${target_dir}'"
+    doRunCommand "tar xJf '${1}' -C '${target_dir}'"
   elif [ "${ext}" = ".tgz" ]
   then
     doRunCommand "tar xfz '${1}' -C '${target_dir}'"
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -353,6 +353,9 @@ function doExtract() {
   elif [ "${ext}" = ".tar" ]
   then
     doRunCommand "tar xf '${1}' -C '${target_dir}'"
+  elif [ "${ext}" = ".xz" ]
+  then
+    doRunCommand "tar xf '${1}' -C '${target_dir}'"
   elif [ "${ext}" = ".tgz" ]
   then
     doRunCommand "tar xfz '${1}' -C '${target_dir}'"


scripts/src/main/resources/scripts/functions
```diff
@@ -351,6 +351,9 @@ function doExtract() {
   then
     doRunCommand "unzip -q '${1}' -d '${target_dir}'"
   elif [ "${ext}" = ".tar" ]
+  then
+    doRunCommand "tar xf '${1}' -C '${target_dir}'"
+  elif [ "${ext}" = ".xz" ]
   then
     doRunCommand "tar xf '${1}' -C '${target_dir}'"
```
hohwille: shouldnt it be "tar xJf ..." for *.xz files?
BTW: Which software tool is using this odd format?

hohwille: Will that be supported in git-bash (mingw) after all?

hohwille:
@sobkowiak Thanks for your improvement. I am happy to receive PRs with features or cases I am not aware of. 👍 
So I am happy to include this change. However, could you comment on my little review question?

scripts/src/main/resources/scripts/functions
sobkowiak: tar detects the file format, but you are right, maybe adding the option J would be better

sobkowiak: I"ll check the bash-git tomorrow

sobkowiak: > Will that be supported in git-bash (mingw) after all?

As you can see in the linked issue (#169) nodejs is provided as .[xz](https://en.wikipedia.org/wiki/Xz) file (`node.js-v10.16.0-linux.tar.xz`) I can see more and more often this file format.

sobkowiak: > 
> 
> I"ll check the bash-git tomorrow

Ye, it works under git-bash

markusschuh:
More with respect to the original code of doExtract in functions - not specifically this change:

For both bsdtar (macOS) as well as gnu-tar it is recommended to let them autodectect archive formats of existing archives and to not specify the decompress format as additional option.
A simple 

 tar xf $archive

will work for all archives, first "tar"ed, then - optionally - compressed with either gzip, bzip2 or xz.

The current code within "functions" tries to redo the internal tar logic within the outside scripting. This looks ineffective. Also it unnecessarily introduces small glitches - e.g the *.tar.bz2 case is not yet handled properly. I would just check for an acceptable extension

`ext=$( echo $filename | sed -E -e 's/^.*(\.(zip|dmg|tar|tar\.gz|tgz|tar.bz2|tbz2|tar.xz))$/\1/' )`

and then differentiate between the cases zip, dmg, empty string or else - where the else case is always this

`doRunCommand "tar xf '${1}' -C '${target_dir}'"`

hohwille:
@markusschuh thanks for your feedback. I agree that we should solve this more general and especially support `*.tar.*` this way. I assume that the autodetection of tar is (even more) reliable and works on all platforms. Then I would not merge this PR and recreate the change from scratch.

hohwille:
Replaced with PR #176

PR #182, opened 2019-08-15
Everything added

SiCawn:
=== Fast Search in Windows
https://voidtools.com/[Everything] is a desktop search utility for Windows that can rapidly find files and folders by name.


maybeec:
Mh I would say it makes sense to put it as an optional package, but for me this is not mandatory to have. I have everything installed on my machine already. So in my opinion, we should keep the IDE on a minimal and allow maybe by devon install everything to install further tools if needed. This anyhow might be worth to have a configuration for such default installs to make it customizable for every project.

hohwille:
@maybeec this is just adding a hint to the documentation section for the advanced tooling. I think this tool is cool and worth mentioning.

Issue #178, opened 2019-08-14
setup needs administrator rights


maybeec:
Just a question. I installed a default setup on my system. I got twice a "configuration as an admin" dialog to allow writing into the registry for example.
Is this working for example on a machine without admin rights or does the setup stop in this moment, which would be a pitty.

hohwille:
1. this has to be done only once.
2. if you want to know if the config fails without admin rights simply click on cancel and see how the scripts behave.

No time for testing right now, but I will have a look before we finalize the release. As I do not have windows as OS such tests are a little low prio from my end.

hohwille:
Maybe we should allow a variable / property to disable stesp that need admin rights. Then such users can put this variable into their `~/devon.properties`.

markusschuh:
This is caused by the following two lines at the end of setup.bat:
```
call regedit.exe /s system/windows/cmd/devon-cmd.reg
call regedit.exe /s system/windows/power-shell/devon-power-shell.reg
```

This does not need admin right since only the current user's registry path is changed. I'd the message of the user account control would look different for admin rights.

Anyway it seems to work without the pop up, when the "newer" (less older) reg.exe were used:
```
reg import system/windows/cmd/devon-cmd.reg
reg import system/windows/power-shell/devon-power-shell.reg
```

hohwille:
Should be fixed with PR #183 
@maybeec and @markusschuh you rock!
I will also add you as developers in the POM. Thanks for your contributions and making `devon-ide` better. I think we need some final testing and bugfixing and can then go for our first official release.

hohwille:
FYI: To be consistent, we should make the change for all places using `regedit.exe`:

This one was missed:
https://github.com/devonfw/devon-ide/blob/8aeb3b4713d60276c0978c23acb68274a1ab07b6/scripts/src/main/resources/scripts/command/ide#L178

hohwille:
Now complete.

PR #183, opened 2019-08-16
use new API for regedit

maybeec:
#178 fix proposed by @markusschuh


hohwille:
👍 thanks.
I assume this works as expected. We need to test this properly once I have created the next release.

Issue #198, opened 2019-08-20
PowerShell support broken


hohwille:
I am not sure if it is related to #178 but now PowerShell integration (`Open Devon PowerShell here`) stopped working.
When I click on it, I get this error message in a Popup:
```
This file does not have a program associated with it for performing this action. Please install a program or, if already installed, create an association in the Default Programs control panel.
```

hohwille:
https://github.com/devonfw/devon-ide/blob/master/scripts/src/main/resources/system/windows/power-shell/devon-power-shell.reg

hohwille:
I fixed the path to `devon.bat` in the registry file. I already tested this on windows but it still fails with the same error.

hohwille:
OK. Seems as if the variable resolution does not work anymore. We should go for `profile.ps1` file like we do for other shells.
However, devon.bat does not work properly in powershel while it works in CMD.
It is entirely wired. I hacked some debug code into it and did tweak this block:
```
...
:setup_env
set "WORKSPACE=main"
echo seting up devon ide
pushd %CD%
echo buh
:iterate_backwards
echo hi
if exist scripts\environment-project.bat (
  echo a
) else (
  echo b
)
echo here
if exist scripts\environment-project.bat (
  echo foo
  set DEVON_IDE_HOME=%CD%
  echo DEVON_IDE_HOME=%CD%
  call scripts\environment-project.bat
  echo devon-ide environment variables have been set for %CD% (workspace %WORKSPACE%)
  popd
  goto :eof
) else (
  echo mmmmmm
)
echo bla
...
```
When I call `devon.bat` in powershell, I get this output:
```
seting up devon ide
buh
hi
b
here
```
Please note that if I update the working echo statements, I get the updated output so it is for sure I am editing the right file that is also called from powershell.
But how can the same if else block evaluate fine with only echos but terminate the entire processing in the second case? I am totally puzzled. Windows sucks so hard...

hohwille:
I outcommented every single statement in the problematic if.
The error comes from the `goto` statement.
Is `goto` not working inside if-statements when I am inside power-shell?
OMG, Windows is such a pile of hacks.

hohwille:
Is somebody keen on implementing and maintaining a `devon.ps1` script as well?
Or should we simply drop power-shell support?

hohwille:
I found another bug and I assume this could never ever have worked before:
[...]

PR #208, opened 2019-08-21
#198: fixed powershell integration

hohwille:
fix for #198

### DIFF
--- a/scripts/src/main/resources/system/windows/power-shell/devon-power-shell.reg
+++ b/scripts/src/main/resources/system/windows/power-shell/devon-power-shell.reg
@@ -3,8 +3,8 @@ Windows Registry Editor Version 5.00
 [HKEY_CURRENT_USER\Software\Classes\Directory\shell\devon_ps]
 @="Open Devon PowerShell here"
 [HKEY_CURRENT_USER\Software\Classes\Directory\shell\devon_ps\command]
-@="powershell.exe -NoExit -Command Set-Location -LiteralPath '%L';devon.bat"
+@="cmd.exe /s /c devon.bat & powershell.exe -NoExit -Command Set-Location -LiteralPath '%L'"
 [HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\devon_ps]
 @="Open Devon PowerShell here"
 [HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\devon_ps\command]
-@="powershell.exe -NoExit -Command Set-Location -LiteralPath '%V';devon.bat"
+@="cmd.exe /s /c devon.bat & powershell.exe -NoExit -Command Set-Location -LiteralPath '%V'"
--- a/scripts/src/main/resources/system/windows/power-shell/devon-power-shell.reg
+++ b/scripts/src/main/resources/system/windows/power-shell/devon-power-shell.reg
@@ -3,8 +3,8 @@ Windows Registry Editor Version 5.00
 [HKEY_CURRENT_USER\Software\Classes\Directory\shell\devon_ps]
 @="Open Devon PowerShell here"
 [HKEY_CURRENT_USER\Software\Classes\Directory\shell\devon_ps\command]
-@="powershell.exe -NoExit -Command Set-Location -LiteralPath '%L';%USERPROFILE%\\scripts\\devon.bat"
+@="powershell.exe -NoExit -Command Set-Location -LiteralPath '%L';devon.bat"
 [HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\devon_ps]
 @="Open Devon PowerShell here"
 [HKEY_CURRENT_USER\Software\Classes\Directory\Background\shell\devon_ps\command]
-@="powershell.exe -NoExit -Command Set-Location -LiteralPath '%V';%USERPROFILE%\\scripts\\devon.bat"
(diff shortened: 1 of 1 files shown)


hohwille:
I found another bug and I assume this could never ever have worked before:
When powershell runs a command (e.g. `devon.bat`) it does not behave like CMD but like a real shell. That is, it creates a sub-shell where all the side-effects of that command take place but they do not interfere the parent shall. Therefore `devon.bat` can properly set the path but afterwards it exists and returns to the powershell that does not see any of these changes to environment variables.
That is why we are not executing the `devon` command in bash but `source` it. As CMD is based on the side effects we had this different approach for windows CMD so far.
AFAIK there is no `source` for powershell to run a bat file allowing modifications of env variables. This would mean:
* We had to run `devon.bat` prior to running powershell so it is inheriting the changes.
* We can not update the environment variables in powershell with just running `devon` command what makes powershell support still kind of broken.
* Alternatively we could make devon command open a new powershell session after setting environment variables. This could however, lead to ugly cascaded nesting of shells.

hohwille:
I was wrong about powershell. Sourcing is supported:
https://superuser.com/questions/71446/equivalent-of-bashs-source-command-in-powershell

hohwille:
Sorry to say so, but powershell is broken. Maybe it is because windows 10 comes with powershell 1.0 while version 6 is already out. But this entirely sucks and I am giving up. It seems that some windows fan who is keen on powershell support would need to take over.
```
PS C:\projects\cool-project\system\windows\power-shell> . $env:USERPROFILE/scripts/devon.bat
devon-ide environment variables have been set for C:\projects\cool-project in workspace main
PS C:\projects\cool-project\system\windows\power-shell> mvn -v
mvn : The term 'mvn' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ mvn -v
+ ~~~
    + CategoryInfo          : ObjectNotFound: (mvn:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

PS C:\projects\cool-project\system\windows\power-shell> echo $env:PATH
C:\Program Files (x86)\Common Files\Oracle\Java\javapath;C:\ProgramData\Oracle\Java\javapath;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\Program Files\Git\cmd;C:\Users\hohwille\AppData\Local\Microsoft\WindowsApps;C:\Users\hohwille\scripts;
PS C:\projects\cool-project\system\windows\power-shell> devon mvn -v
Running: /c/projects/cool-project/software/maven/bin/mvn -v
Apache Maven 3.6.1 (d66c9c0b3152b2e69ee9bac180bb8fcc8e6af555; 2019-04-04T21:00:29+02:00)
Maven home: C:\projects\cool-project\software\maven
Java version: 11.0.4, vendor: AdoptOpenJDK, runtime: C:\projects\cool-project\software\java
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
devon-ide environment variables have been set for C:\projects\cool-project in workspace main
PS C:\projects\cool-project\system\windows\power-shell>
```

hohwille:
```
Set-Alias : A positional parameter cannot be found that accepts argument 'C:\Users\hohwille\scripts\devon.bat'.
At \\vmware-host\Shared Folders\Documents\WindowsPowerShell\profile.ps1:1 char:1
+ Set-Alias -Name devon -Value . $env:USERPROFILE\scripts\devon.bat
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Set-Alias], ParameterBindingException
    + FullyQualifiedErrorId : PositionalParameterNotFound,Microsoft.PowerShell.Commands.SetAliasCommand
```
So all tricky and nothing working as expected with powershell.
Even 
```
echo Welcome to PowerShell
```
Is giving this output:
```
Welcome
to
PowerShell
```
There are really nice concepts in PowerShell but I am facing too many drawbacks here.

I think it is a good move that Microsoft deeply integrates bash, ssh and the entire linux subsystem as well as docker into windows. I will therefore focus on bash and leave powershell support open to some drivers who really want to have this feature. I will do a minimum fix now and are done.

BTW: `profile.ps1` is not resolved from `$env:USERPROFILE\Documents\WindowsPowerShell\profile.ps1`. On my machine `Documents` are mounted from VMWareFusion to a totally different location:
```
$profile | Format-List * -force

AllUsersAllHosts       : C:\Windows\System32\WindowsPowerShell\v1.0\profile.ps1
AllUsersCurrentHost    : C:\Windows\System32\WindowsPowerShell\v1.0\Microsoft.PowerShell_profile.ps1
CurrentUserAllHosts    : \\vmware-host\Shared Folders\Documents\WindowsPowerShell\profile.ps1
CurrentUserCurrentHost : \\vmware-host\Shared Folders\Documents\WindowsPowerShell\Microsoft.PowerShell_profile.ps1
Length                 : 89
```

So if someone wants to take over, he would have to figure out via which variables one can figure out where to put or find `profile.ps1`.

hohwille:
https://stackoverflow.com/questions/49027851/setting-environment-variables-with-batch-file-lauched-by-powershell-script

PR #250, opened 2019-09-16
Renaming URLs to correct repo name

jdiazgon:
This repository has been renamed from `devon-ide` to `ide`. This has broken our CobiGen Oomph installer as the URLs are not correct.

This PR fixes this issue.

### DIFF
--- a/404.html
+++ b/404.html
@@ -9,7 +9,7 @@
 
 <body class="error no-js">
   <script type="text/javascript">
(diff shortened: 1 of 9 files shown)


hohwille:
BTW: I am not entirely sure if this OOMPH content is perfectly right in this place as I am not really able to maintain it since I am not using it and have way too little knowledge. However, lets hope it is now stable for a while...

Issue #175, opened 2019-08-13
Ability to import project into Eclipse automatically


hohwille:
For Eclipse (and with additional issues also for other IDEs) we should introduce the ability to import an existing (maven) project into the current workspace:
```
devon eclipse import [«path»]
```
Should import the specified project from the given `«path»` into the current Eclipse workspace. If `«path»` is omitted, the current working directory is used.
My proposal would be to always use the current workspace:
```
cd workspaces/main
devon eclipse import ../test-workspace/myproject
```
Would import `myproject` into the Eclipse of the `main` workspace and not into `test-workspace`, what IMHO is consistent and allows special cases (e.g. importing `workspaces/main/CobiGen_Templates` into all workspaces using `--all` option).

If the project is a maven project, the import should behave like `Import existing maven projects into workspace`. Other cases are just nice-to-have (gradle, general plain project).
IMHO the most tricky thing is to figure out, how to trigger the import in Eclipse from CLI.

As extension it should also be possible to provide a git URL for `«path»`. This would clone the git URL to the current working directory and import the resulting project.

If we managed to do that, we should also extend `java` commandlet such that it could clone and import my-thai-star into the current workspace (unless already done).

hohwille:
https://www.eclipse.org/forums/index.php/t/109302/

hohwille:
As an alternative we could do that manually once and observe all that Eclipse is doing:
* create `.project`
* create `.classpath`
* create changes in workspace `.metadata` folder

This is however reinvent the wheel and would be hard to maintain. If we are able to trigger the eclipse feature programatically it would be awesome.

hohwille:
I tried `eclipse -application org.eclipse.cdt.managedbuilder.core.headlessbuild -import ...` but Eclipse is a big mess as always:
```
Could not resolve module: org.eclipse.mylyn.bugzilla.core [277]
  Unresolved requirement: Require-Bundle: org.apache.xmlrpc
    -> Bundle-SymbolicName: org.apache.xmlrpc; bundle-version="3.0.0.v20100427-1100"
```
I "resolved" this by uninstalling bugzilla support for the moment and ran into the next issue:
```
Could not resolve module: org.eclipse.epp.logging.aeri.ide [145]
  Unresolved requirement: Import-Package: org.apache.lucene.document; version="[7.1.0,8.0.0)"
```
This lead me to this bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=540306
[...]

PR #252, opened 2019-09-16
added OS-specific path handling and no-backgrounding flag for 'doStartEclipse' function

phschimm:
Still contains `echo` statements and comments for code review / debugging.
Needs testing under macOS / Linux.


scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -213,7 +226,12 @@ function doImportProject() {
 
   local setup_templ="${setup_templ_dir}/${setup_templ_name}"
   local setup_ws="${setup_ws_dir}/${setup_ws_name}"
-  
+
+  if doIsWindows # cut drive letter from Windows path only (not neccesarry under macOS/Linux)
+  then
+    proj_dir=${proj_dir:2} # for example 'C:/Users/' becomes '/Users/'
```
hohwille: never do things like this. Use https://cygwin.com/cygwin-ug-net/cygpath.html instead (also available in git-bash/MINGW/MSYS).

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -152,16 +152,29 @@ function doAddPlugin() {
   fi
 }
 
+# $1 no-backgrounding flag -nb
 function doStartEclipse() {
-  doConfigureEclipse -u
-  echo "launching Eclipse in workspace ${WORKSPACE} at ${WORKSPACE_PATH}"
-  if doIsWindows
+  if [ "${1}" = "-nb" ]
```
hohwille: what option is `-nb` for? You did not add it to the help usage.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -213,7 +226,12 @@ function doImportProject() {
...
-  
+
+  if doIsWindows # cut drive letter from Windows path only (not neccesarry under macOS/Linux)
+  then
+    proj_dir=${proj_dir:2} # for example 'C:/Users/' becomes '/Users/'
+  fi
+
   echo "SETUP_TEMPL: ${setup_templ}" #DEBUG
   echo "PROJ_NAME:   ${proj_name}" #DEBUG
   echo "PROJ_DIR:    ${proj_dir}" #DEBUG
```
hohwille: where do these debug outputs come from? They are not present on master branch but for some reason not show as added within this PR. Github bug?
Should actually not make it into the release...

scripts/src/main/resources/scripts/command/eclipse
hohwille: Please also stay conform with POSIX conventions:
https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

phschimm: It launches the Eclipse instance without `start` (Windows) or an appended `&` (macOS/Linux).
So the terminal waits until the Eclipse instance is terminated (required for automated file cleanup).

Do we add deatails to internal functions to the documentation usually?
The function `doConfigureEclipse` also allows for an optional `-u` flag, which isn't documented anywhere (or maybe I've overlooked it) ...

scripts/src/main/resources/scripts/command/eclipse
phschimm: I added them in an earlier commit ... 
Might still be useful for testing on macOS/Linux I thought (branch is still WIP).

scripts/src/main/resources/scripts/command/eclipse
phschimm: Ok, @markusschuh already gave me a hint in that direction but I'm not sure how to get rid of the leading drive letter that way. His suggested code was:

if command -v cygpath >/dev/null
then
  proj_dir=$( cygpath --mixed "${proj_dir}" )
fi

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -182,75 +195,79 @@ function doImportProject() {
...
-  
-  echo "SETUP_TEMPL: ${setup_templ}" #DEBUG
-  echo "PROJ_NAME:   ${proj_name}" #DEBUG
-  echo "PROJ_DIR:    ${proj_dir}" #DEBUG
+
+  if command -v cygpath >/dev/null # if cygwin/git bash are used (Windows only)
+  then
+    proj_dir=$(cygpath -m "${proj_dir}") # convert Unix-style path to Windows-style path
+    proj_dir=$(echo "${proj_dir}" | cut -d':' -f2-) # if existent, cut drive letter
+  fi
```
phschimm: I'm using `cygpath` now to convert the path format and cut the drive letter based on the existence of a `:` in the path (which is only ever the case under Windows).

One (minor) problematic result of this approach is, that if the user enters something like `/myDir/subDir` under Windows, `cygpath` will turn it into `C:/Program Files/Git/myDir/subDir` (because no drive was defined, `cygpath` defaults to the Git/Cygwin directory).

If the user correctly enters `/c/myDir/subDir`, everything works as expected:
* `cygpath` turns the path into `C:/myDir/subDir` (mixed format)
* using `cut` and the delimiter `:` the leading drive letter is removed

I doubt that many people will make use of the option to enter an absolute path, since `cd`'ing to the desired directory is far more convenient (which doesn't excuse the problematic behaviour of course). However, as stated in the initial [issue description](https://github.com/devonfw/ide/issues/175#issue-479992948), this functionality could be useful in certain cases, so I wanted to support it.

Issue #265, opened 2019-10-01
Eclipse commandlet doesn't install Java


jdiazgon:
### Expected behavior

When installing Eclipse by running command `devon eclipse`, it should download the needed JDK so that Eclipse can be installed properly out of the box.

### Actual behavior
 
After running `devon eclipse`, Eclipse gets installed successfully. However, when installing the plug-ins I get the following errors:

![image](https://user-images.githubusercontent.com/31688036/65975963-bb83f180-e46f-11e9-8927-ea161eebde85.png)

IMHO, `devon eclipse` should call `devon java` so that this error does not occur. Eclipse without Java does not make sense, so anyway we should force the installation of a JDK.

### Steps to reproduce (bug)

1. Don't run `devon java`
2. Run `devon eclipse`

### Comments
Removing Eclipse, running `devon java` and `devon eclipse` afterward, fixed the issue

### Affected version:

* OS: Windows
* Browser: Chrome

jdiazgon:
>  why the java setup is not also called, when the users just executes devon eclipse?

I completely agree with you. From a user point of view, `devon eclipse` should install correctly Eclipse. This means that it should download all needed dependencies (like Java)

If this logic was implemented because of performance issues, then I propose the following:

```BASH
> devon eclipse

Eclipse has not yet been installed on your computer.
Please run "devon eclipse setup" to install it.
```

hohwille:
I do not fully understand what is going on here:
1. devon-ide has a setup. Initially users will call `setup[.bat]` at any later time they can also do updates (`devon ide update`). Also one can install individual tools even if they are not planned to be used by the project according to their settings (e.g. `devon eclipse setup`).
2. next there is the usage of the ide after it has been setup (e.g. `devon eclipse`). This currently does not recursively validate the installation but expects that a setup has been performed and the installation is sane.

Can you
* confirm that everything works fine if the setup has been perfromed?
* clarify if your expecation really is that every devon command should recursively verify the installation of all required tools and reinstall as needed? If so this is a general change of the design and should not be addressed by a PR that only considers eclipse. Please not that this currently also conflicts with requirements from @maybeec who wanted to have the most little overhead and logging output so devon commands are fast and do not spam the log output.

hohwille:
https://github.com/devonfw/ide/blob/master/documentation/setup.asciidoc
What is the `setup` good for if we expect that the users do not follow the installation process and just ignore it?

jdiazgon:
[...]

PR #266, opened 2019-10-01
Eclipse with JDK out of the box

jdiazgon:
This PR fixes issue #265. It forces the installation of a JDK prior to installing Eclipse.

I have tested it on my local machine and now everything is working fine.

### DIFF
--- a/scripts/src/main/resources/scripts/command/eclipse
+++ b/scripts/src/main/resources/scripts/command/eclipse
@@ -21,10 +21,7 @@ function doInstallEclipsePlugin() {
 }
 
 function doSetup() {
-  if [ -n "${1}" ]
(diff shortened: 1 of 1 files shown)


jdiazgon:
Should the same be done for Maven? The same logic is found [here](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/mvn#L7)

hohwille:
@jdiazgon Thanks for your effort to make devon-ide better. However, we need to be clear about requirements and design and not do changes locally without following a consistent strategy. Therefore please read my comments in #265

jdiazgon:
Closing as in #265 we have discussed better alternatives

Issue #59, opened 2019-03-28
Integrate CobiGen CLI as commandlet


hohwille:
It will be a big thing but we definitely need to integrate CobiGen into `devon-ide` and our new CLI. I leave this issue to the experts from the CobiGen team...

hohwille:
Besides the CLI also we need to integrate cobigen as a plugin for Eclipse. Therefore all we need is a public update site URL and the feature id:
https://github.com/devonfw/devon-ide/blob/8a8c4a661d81e325260fd6c44423bfdb07f64b74/scripts/src/main/resources/scripts/command/eclipse#L106

hohwille:
@maybeec / @jdiazgon any progress, any milestone, any feedback?
Seems I have to release `3.0.0` without CobiGen integration.

hohwille:
This issue is not complete. Usually PRs should come with code AND documentation.
However, #273 only shipped with cobigen commandlet.
The documentation for this is still missing.
It should be fine to like to the CobiGen CLI documentation but we at least have to create a small AsciiDoc for the `cobigen` commandlet and link it from the list of commandlets.

hohwille:
To not block the progress I already merged #273 but also wanted to note that we define a variable for each tool to configure the exact version. This is helpful to allow (large) projects to exactly pre-define the version of the tools used in their `devonfw-ide` installation. It might be that CobiGen drops support for Java8 and a project still needs to stay on Java8 or the latest CobiGen may be shipped with a bug. Having no way to control this does not meet the goals of `devonfw-ide` to allow stability and freedom for the projects.
We need an additional PR that will honor an optional `COBIGEN_VERSION` variable (just analogue to all other commandlets).

hohwille:
Further, the current approach with `doDownload` is a hack.
It will not support software updating. You would need to rewrite the entire logic to be able to detect the currently installed version and compare if a newer version is available.
IMHO CobiGen has to provide the CLI in a completely different way as *.tar.gz file with executable Flags and such that it contains what needs to be in the software folder to be fully functional.

```
cg -h
Error: Unable to access jarfile /Users/hohwille/projects/devon/software/cobigen-cli/cobigen.jar
```

hohwille:
Considered done. Created #340 for the remaining issue discussed above.

PR #273, opened 2019-10-11
Cobigen CLI commandlet

jdiazgon:
Addressing #59. This PR adds a new commandlet capable of installing the CobiGen CLI inside of the ide.

Ready to be merged.


scripts/src/main/resources/scripts/command/cobigen
```diff
@@ -0,0 +1,58 @@
...
+# $@: optional maven arguments
+function doSetup() {
+  # We need a JDK
+  doDevonCommand java -q setup
+  # We need Maven for installing plug-ins
+  doDevonCommand mvn -q setup
+  if [ -n "${1}" ] || [ ! -d "${COBIGEN_CLI_HOME}" ]
+  then
+    local software_version="LATEST"
+    local download_url="https://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=com.devonfw.cobigen&a=templates-devon4j&v=${software_version}"
```
jdiazgon: This is the download URL I need to change. Right now I'm just downloading CobiGen templates and not the CLI.

scripts/src/main/resources/scripts/command/cobigen
```diff
@@ -0,0 +1,58 @@
...
+    then
+      doRunCommand "unzip -q '${COBIGEN_CLI_HOME}/${software_name}.jar' 'cobigen.bat' -d '${COBIGEN_CLI_HOME}'"
+      doRunCommand "unzip -q '${COBIGEN_CLI_HOME}/${software_name}.jar' 'cg.bat' -d '${COBIGEN_CLI_HOME}'"
+      doRunCommand "unzip -q '${COBIGEN_CLI_HOME}/${software_name}.jar' 'class-loader-agent.jar' -d '${COBIGEN_CLI_HOME}'"
+    else
+      doRunCommand "unzip -q '${COBIGEN_CLI_HOME}/${software_name}.jar' 'cobigen.sh' -d '${COBIGEN_CLI_HOME}'"
+      doRunCommand "unzip -q '${COBIGEN_CLI_HOME}/${software_name}.jar' 'cg.sh' -d '${COBIGEN_CLI_HOME}'"
+      doRunCommand "unzip -q '${COBIGEN_CLI_HOME}/${software_name}.jar' 'class-loader-agent.jar' -d '${COBIGEN_CLI_HOME}'"
+    fi
+
```
markusschuh: The "doIsWindows" is a bit misleading name: It (currently) results in "yes", when the local bash environment is either msys (typically installed by git-for-windows) or cygwin. But both those firewalls need the *.sh files, when cobigen shall be started from **inside** the environment. In addition to that, the *.bat may also help - but only, when a developer also uses windows-CMD.

It might be easiest to extract just all files in any case.

hohwille: Yes, when we install maven or other multi-platform tools, we just extract everything.
Cant you just use `doInstall` function for cobigen as well?

scripts/src/main/resources/scripts/devon.properties
```diff
@@ -15,6 +15,8 @@ export JAVA_HOME=${DEVON_IDE_HOME}/software/java
 #export M2_REPO=~/.m2/repository
 export M2_REPO=${DEVON_IDE_HOME}/conf/.m2/repository
 export MAVEN_HOME=${DEVON_IDE_HOME}/software/maven
+# CobiGen CLI needs this
+export M2_HOME=${DEVON_IDE_HOME}/software/maven
```
hohwille: conflicts with #5 - can't you just set this variable within the `cobigen` commandlet? You may even export it there if really needed, but exporting it here will also expose it for any shell when just `devon` is called.
Further, `M2_REPO` was IMHO historically given for some reason. However, the shortcut M2 is also historically. Maven itself states on its website that M2 should be avoided and it should be `MAVEN_HOME` rahter than `M2_HOME`. Tying the version to `.m2` folder and some variable names was a kind of design flaw in maven2 and we are at maven3 already for years.

scripts/src/main/resources/scripts/functions
```diff
@@ -357,7 +357,7 @@ function doExtract() {
   if [ "${filename_base/*\./.}" = ".tar" ] || [ "${ext}" = ".tar" ] || [ "${ext}" = ".tgz" ] || [ "${ext}" = ".tbz2" ]
   then
     doRunCommand "tar xf '${1}' -C '${target_dir}'"
-  elif [ "${ext}" = ".zip" ]
+  elif [ "${ext}" = ".zip" ] || [ "${ext}" = ".jar" ]
```
hohwille: I have seen strange effects where JAR files needed to be extracted with the `jar` tool instead of `unzip` that was generating checksum errors in some edge-cases.

jdiazgon:
Thanks for the feedback! I will update it as soon as possible.

> why do you distribute cobigen CLI as JAR? IMHO you should distribute it as *.tgz and set permissions accordingly

Seems a good idea! I think I will implement it for the next release as we are already on a hurry for the current version and I don't have time for new changes.

scripts/src/main/resources/scripts/command/cobigen
jdiazgon: Unfortunately I cannot use `doInstall` as my `.jar` extension is not on the list of `doExtract`.

scripts/src/main/resources/scripts/devon.properties
jdiazgon: > can't you just set this variable within the cobigen commandlet? 
I have tried to do that by adding the `export` statement to the commandlet.

However, this is not working for me because that `export` only takes effect on the subshell (please see [this explanation](https://askubuntu.com/questions/53177/bash-script-to-set-environment-variables-not-working)).

I need the `M2_HOME` variable because inside CobiGen CLI we invoke Maven in order to resolve all the CobiGen plug-ins. If it is not set, it throws this error:

```Java
java.lang.IllegalStateException: Maven application directory was not specified, and ${maven.home} is not provided in the system properties. Please specify at least on of these.
```

Maybe you can explain me how to correctly `export` the variable.

maybeec:
I am not sure whether I got it right, but cobigen-cli is a runnable jar, so you cannot just release it as it tar.gz or something like this. @jdiazgon as far as I remember, we simply extract a pom.xml file to dynamically download other dependencies, nothing more right?

jdiazgon:
> we simply extract a pom.xml file to dynamically download other dependencies

That's right. However, for JDK 11 I had to add a [Java agent](https://www.baeldung.com/java-instrumentation) as the class loading has completely changed. Right now our runnable jar contains that pom.xml and the jar of the Java agent.

maybeec:
@jdiazgon the java11 compatibility is fixed, right? so this could be merged?

jdiazgon:
> the java11 compatibility is fixed, right? so this could be merged?

It is fixed but we should not merged yet as the download URL is not the correct one. Please see [this](https://github.com/devonfw/ide/pull/273#discussion_r333915734).

Also the Travis build is failing with the following error:

```bash
In scripts/command/cobigen line 35:
  doEcho "Running: ${cobigen_cmd} ${@}"
                                  ^--^ SC2145: Argument mixes string and array. Use * or separate argument.
In scripts/command/cobigen line 41:
    "${cobigen_cmd}" "${@} --verbose"
                      ^--^ SC2145: Argument mixes string and array. Use * or separate argument.
```

It seems a Linux related error. I will fix it ASAP.

markusschuh:
With respect to

```
In scripts/command/cobigen line 35:
  doEcho "Running: ${cobigen_cmd} ${@}"
                                  ^--^ SC2145: Argument mixes string and array. Use * or separate argument.
In scripts/command/cobigen line 41:
    "${cobigen_cmd}" "${@} --verbose"
                      ^--^ SC2145: Argument mixes string and array. Use * or separate argument.
```

The findings of shellcheck (which is run as part of travis) are correct ones. An [explanatory URL](https://github.com/koalaman/shellcheck/wiki/SC2145) is available within the travis build details.

I'd say, the correct fixes were to use the following code instead:

```
  doEcho "Running: ${cobigen_cmd} ${*}"
. . .
    "${cobigen_cmd}" --verbose "${@}"
```

This is not only true for linux but for bash scripts on any platform.

scripts/src/main/resources/scripts/command/cobigen
```diff
@@ -32,13 +32,13 @@ function doSetup() {
 # $@: optional CobiGen arguments
 function doRunCLI() {
   local cobigen_cmd="cobigen"
-  doEcho "Running: ${cobigen_cmd} ${@}"
+  doEcho "Running: ${cobigen_cmd} ${*}"
   if doIsQuiet
   then
-    "${cobigen_cmd}" "${@}"
+    "${cobigen_cmd}" "${*}"
```
markusschuh: The use of "${@}" vs "${*}" is a bit hard to understand in bash.  In this case,  you should **keep** the  "${@}", since that keeps the "quoting per each single argument", so  if you call this commandlet as
   commandlet   a  "start end"
these - hypothecial - arguments are correctly forwarded to the internally called command 
arg1=a
arg2="start end"

scripts/src/main/resources/scripts/command/cobigen
```diff
@@ -32,13 +32,13 @@ function doSetup() {
 # $@: optional CobiGen arguments
 function doRunCLI() {
   local cobigen_cmd="cobigen"
-  doEcho "Running: ${cobigen_cmd} ${@}"
+  doEcho "Running: ${cobigen_cmd} ${*}"
```
markusschuh: Looks good. **Within** echoed quoted strings, only the $* - or ${*}  -  can be used. See below.

scripts/src/main/resources/scripts/command/cobigen
```diff
@@ -32,13 +32,13 @@ function doSetup() {
...
   if doIsQuiet
   then
-    "${cobigen_cmd}" "${@}"
+    "${cobigen_cmd}" "${*}"
     exit ${?}
   else
-    "${cobigen_cmd}" "${@} --verbose"
+    "${cobigen_cmd}" "${*} --verbose"
```
markusschuh: Similar here as above. But you also need to remove the "--verbose" from within the quote. The line should be
"${cobigen_cmd}" "${@}" --verbose

scripts/src/main/resources/scripts/devon.properties
```diff
@@ -15,6 +15,8 @@ export JAVA_HOME=${DEVON_IDE_HOME}/software/java
...
 export MAVEN_OPTS=-Xmx512m -Duser.home=${DEVON_IDE_HOME}/conf
```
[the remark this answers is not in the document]
jdiazgon: I have found a way to programmatically set `MAVEN_HOME` instead of `M2_HOME` inside the CobiGen CLI. I have tested it and it should not be a problem now.

jdiazgon:
Besser spät als nie. This PR is ready to be merged as the CobiGen CLI has been released to public.

maybeec:
@hohwille  if this PR is fine for you as well now, please merge it.

jdiazgon:
The link which should download the latest version of the CLI is returning 404: 

https://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=com.devonfw.cobigen&a=cli&v=LATEST

We guess that Maven Central needs to refresh its metadata and it will take some hours to be available. However, I released yesterday and it should have been enough time. IMHO, if in 3 hours it is not yet working, we should go for a URL that works like:

https://repository.sonatype.org/service/local/artifact/maven/redirect?r=central-proxy&g=com.devonfw.cobigen&a=cli&v=1.0.1

jdiazgon:
I had to change the "LATEST" URL as it was not working yet. PR ready to be merged.

maybeec:
@jdiazgon, the LATEST link is working now.

scripts/src/main/resources/scripts/command/cobigen
hohwille: I can support .jar in `doExtract` in a second. Why do you have "useless" files in your CLI release?
Can you change the way you bundle the release? (we do not have to block this PR now, but on the long run I would prefer cleaning this code as it appears odd to me).
Further the content of this JAR is designed/defined in CobiGen but here you are spreading redundant knowledge about it. What if some day CobiGen decides to rename or replace `class-loader-agent.jar`? Then this will break and nobody will notice as the CobiGen team is not anymore aware about this dependency in another repo...

hohwille:
Not entirely complete. See:
https://github.com/devonfw/ide/issues/59#issuecomment-574076221

PR #280, opened 2019-10-16
Added commands to install&update devon4node and cicdgen

dario-rodriguez:
This PR includes two new commands:

- devon4node
- cicdgen

The purpose of those commands is to install&update&execute those devonfw tools.


dario-rodriguez:
Anyway, for me doesn't makes sense create a command inside the ide in order to execute devon4node commands. For me is ok if you install it globally when you install npm.

sjimenez77:
While we have access to `devon4node` and `cicdgen` commands globally once devonfw ide is installed for me it si completely fine too. Both are CLI tools and for the moment it is not necessary to anything else.

Does it make sense?

hohwille:
I think it is complex to clarify this while chatting in this issue. I was asking for workshops and discussions about shaping the devon-ide CLI and its philosophy. So far nobody ever gave feedback to any of my questions regarding the CLI design so I made it up on my own.
I will try to setup a virtual call so we can have a discussion and everybody understands the whole story (as I think this is not yet clear). As everybody is very busy and we are also sharp on time with the release we need some kind of luck to get everything together...

dario-rodriguez:
Hello, as you can see in the npm page, this a beta version so the code is not merge yet. You can find it in the PR is already created: https://github.com/devonfw/devon4node/pull/17
Related with the name, we already noted that name is a bit confusing, so we decided to change the name to @devon4node/cli. The problem is that some ppl are testing the current one so we won't release a new version until we have the feedback.
Maybe the name confuses you so let me explain what we want: Add to devon-ide only the denvon4node CLI, like cobigen and ng.
The motivation is, as you mention before, to be align with other stacks. As the command node already exists maybe I choose a bad name for the new one. 
I will update the node command in order to follow your comments and then we can discuss if this approach match with the ide.

maybeec:
Hi @dario-rodriguez, what is quite interesting is, that you are able to generate also devon4j archetype with your cli. Is that right? What is the difference to the command devon4j provided into the devon ide, which generates a new devon4j project?

In addition, I would suggest to include the commands of the devon4ng cli into the respective commands already available. For example as @hohwille  said, there is already the command `devon java create ...` which creates a new devon4j project based on the devon4j maintained archetype. If devon4ng cli does something different it might be interesting for consistency to include the devon4j generation of devon4ng cli into the `devon java ...` command.

dario-rodriguez:
Hello @maybeec ,
The two commands that I added to this PR are:
- devon4node (needs rework): this command create a new devon4node projects.
- cicdgen: this command is for generate all files related to CICD (pipelines, Dockerfiles...) for all devonfw stacks. It's in the context of devonfw shop floor and prepare the project to work in the Production Line and the deployment with docker/OpenShift.

Another solution could be, instead of creating a new command for cicdgen, add it in the current commands (java, ng, node) like: `devon java cicd`, `devon ng cicd` or something similar. Anyway, it must be separated from the current `create` command, as not all project that uses devon4j will use the PL or docker. That's why we didn't add those files to the current archtype.

What do you think @hohwille ? In your opinion which is the best approach?

maybeec:
This really sounds good. I like the idea of the sub "cicd" command.

documentation/cicdgen.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+
+.Usage of `devon cicdgen`
+[options="header"]
+|=======================
+|*Argument(s)*   |*Meaning*
+|`setup`                 |setup cicdgen (install and verify)
+|`update`                |update cicdgen (reinstall with @latest version and verify)
+|`java «args»`           |generate cicd files for the currect devon4java project
+|`ng «args»`             |generate cicd files for the currect devon4ng project
+|`node «args»`           |generate cicd files for the currect devon4node project
```
hohwille: I think @maybeec expected this to be integrated in the according commands.
IMHO to do that easily, we can build a bridge in those commandlets.
So 
* `devon java cicd «args»` will be an alias/synonym that delegates to `devon cicdgen java «args»`.
* `devon ng cicd «args»` will be an alias/synonym that delegates to `devon cicdgen ng «args»`.
* `devon node cicd «args»` will be an alias/synonym that delegates to `devon cicdgen node «args»`.

If you have the chance to do that today or on Monday, it would be awesome. Delegation is simply done by `doDevonCommand` function. You will find several examples in existing commandles (esp. `build` is full of those).

scripts/src/main/resources/scripts/command/node
```diff
@@ -48,24 +69,37 @@ then
...
-then
+;;
+"setup" | "s")
   doSetup setup
-else
+;;
+"create" | "c")
+  shift
+  doSetupDevon4node setup
+  devon4node new "${@}"
```
hohwille: 1. we have 3 such blocks. Therefore it would be nice to have a function to make it only one line (e.g. `doDevon4node`).
2. The `setup` arg IMHO does not seem to make sense here as all calls to `doSetupDevon4Node` are followed by this argument. You can remove it as well as the according if checking for it in `doSetupDevon4Node`.

scripts/src/main/resources/scripts/command/node
```diff
@@ -48,24 +69,37 @@ then
 fi
 
 #CLI
-if [ "${1}" = "-h" ] || [ "${1}" = "help" ]
-then
-  echo "Setup or run yarn."
+case ${1} in 
+"help" | "-h")
```
hohwille: nice. I should think about using this syntax in all commandlets as it is more compact. Thanks! 👍

scripts/src/main/resources/scripts/command/node
```diff
@@ -31,6 +31,27 @@ function doSetup() {
...
+  else
+    doDevonCommand npm -q setup
+    local npm_command="${DEVON_IDE_HOME}/software/node/npm"
+    if [ ! -x "${npm_command}" ]
+    then
+      npm_command="${DEVON_IDE_HOME}/software/node/bin/npm"
+    fi
+    doDevonCommand npm -q setup
+    doDevonCommand yarn -q setup
+    doRunCommand "'${npm_command}' install -g @devon4node/cli@latest" "install devon4node"
```
hohwille: This is overcomplicated:
1. `doDevonCommand npm -q setup` is called twice
2. if you just do `doDevonCommand npm -q install -g @devon4node/cli@latest` you can replace everything else (except `doDevonCommand yarn -q setup` if you also need `yarn` as prerequisite for `devon4node`).

documentation/cicdgen.asciidoc
dario-rodriguez: This functionality will be added in the next update of this PR

dario-rodriguez:
Hello @hohwille ,

This PR is updated in order to satisfy your comments.

PR #284, opened 2019-10-22
Travis CI modified to add documentation workflow

sjimenez77:


.travis.yml
```diff
@@ -1,4 +1,62 @@
 language: java
 jdk:
   - openjdk8
+  - openjdk11
```
hohwille: We might think about dropping openjdk8 quite soon then
* build time with mulitple JDKs rises
* we are wasting more resources
* devon-ide mainly consists of bash stuff that does not depend on the JDK at all
* devon-ide already ships with JDK11 by default so support for openJDK is not that important anymore
* JDK8 is running out of free support anyhow

sjimenez77: I think that we can remove `openjdk8` as you suggest.

Issue #110, opened 2019-05-10
Tools update support


hohwille:
With `devon ide update` it should be possible to update the entire devon-ide.
Thus if the project lead defines a tool version in `devon.properties` inside the `settings` Git repo and updates it (e.g. `VSCODE_VERSION=1.99.1`) then the above command should be able to also update according tools (e.g. `vscode`) to that version (and reinstall all plugins).

This is however challenging. Many of the tools we support have a bad or inconsistent version and release management:
* Some tools do not offer a way to query the current version (fine is `mvn -v` or `java -version` however the output might be tricky to parse, but vscode or eclipse do not support a commandline option for this).
* Some tools have download URLs that do not easily correlate to the release number. Hence magic knowledge is required to figure out the download URL from the version and OS (see e.g. issue #109)

For the first point we could easily write a `devon-tool-version.properties` file into the `software/«tool»` folder as a workaround (at least for tools that do not support a proper way to determine the installed version).
The second point is even more tricky. IMHO we could only resolve this with either of the following options:
* the user has to provide more information (`VSCODE_VERSION=1.99.1` would not be enough but also `VSCODE_HASH=51b0b28134d51361cf996d2f0a1c698247aeabd8` and `VSCODE_REV=1554971043`) - however, these values are tricky to figure out and users will be lost in space
* we provide a kind of mapping to get the details from the version and maintain that for every version. However, we would need to continuously provide updates of this and VS Code changes quickly. So I assume we are unable to keep the pace and would end up as blocker
* we maintain the mapping separate from devon-ide releases and provide it as separate online configuration download or service. Maintenance could be supported by the entire community so the first user needing the new version would do the work (done in 5 Minutes if you know what to do, we could provide a guide)
* we fully automate this and follow redirects, etc.
* we get in touch with the makers of the tools and ask for improving their downloads
* we redistribute the software in a systematic way

hohwille:
For VSCode there are links that redirect to the latest hash. E.g.:
* https://update.code.visualstudio.com/1.33.1/win32-x64-user/stable
* https://update.code.visualstudio.com/1.33.1/darwin/stable
* https://update.code.visualstudio.com/1.33.1/linux-x64/stable

hohwille:
more or less done with PR #114 
However, currently you need to rerun`devon ide setup`.
So only `devon ide update` needs to be slightly modified to complete this.
Please also note that some tools like e.g. jenkins do not support this new approach and also need to be updated.

[...]

PR #292, opened 2019-10-28
WIP: Added command to setup devon4net and the related documentation

jmsanmu:
This pull request contains the new command dot net to be able to setup the dotnet environment and also the create command to generate the devon4net API project.

If the dotnet SDK is not present on the machine, the user will be redirected to the URL https://dotnet.microsoft.com/download to get the SDK. This behavior is due to Microsoft ecosystem. The use of special paths is mandatory to to detect previous installed SDK and run properly the Visual Studio environment.


scripts/src/main/resources/scripts/command/dotnet
```diff
@@ -0,0 +1,63 @@
...
+  then
+    printf "${RED}dotnet NOT running${NC} \n"
+    printf "${RED}Please install dotnet SDK.${NC} \n"
+    case "$OSTYPE" in
+      solaris*) echo "SOLARIS : not supported platform" ;;
+      darwin*)  open $downloadDotnetUrl ;; 
+      linux*)   exdg-open $downloadDotnetUrl ;; 
+      bsd*)     echo "BSD : not supported platform" ;;
+      msys*)    start $downloadDotnetUrl ;; 
+      *)        echo "unknown: $OSTYPE.  Not supported platform" ;;
```
hohwille: The idea of devonfw-ide is to fully automate the download and installation of the tools and software.
Is that not possible here?
Should we reuse `doIsWindows`, `doIsMacOs` ?
E.g. `doIsWindows` also considers `cygwin`...
You are of course fine to stay with this `case` statement but we need to be aware that we have to keep that in sync with the other stuff and you should then also add `cygwin` to `msys`.

hohwille: I just tried the download for macos and you can simply implement this installation the same way as done for all the other tools (see e.g. vscode, java, etc.).

jmsanmu: @hohwille many thanks for your comments. I'm going to take a look to your suggestions and update the pull request with the required changes.

scripts/src/main/resources/scripts/command/dotnet
```diff
@@ -23,7 +23,7 @@ function doSetup() {
     if doIsMacOs || doIsWindows
     then
       doDownload "${download_url}"
-      doRunCommand "~/Downloads/${fileSetup}"
+      doRunCommand "$HOME/Downloads/${fileSetup}"
```
hohwille: has to be `${DEVON_HOME_DIR}` instead.
See here:
https://github.com/devonfw/ide/blob/master/documentation/variables.asciidoc

scripts/src/main/resources/scripts/command/dotnet
```diff
@@ -0,0 +1,75 @@
...
+  local download_url
+
+  dotnetCommand=$(dotnet --version)
+  
+  if [ "$dotnetCommand" != "$software_version" ]
+  then    
+    if doIsMacOs
+    then
+      fileSetup="dotnet-sdk-3.0.100-osx-x64.pkg"    
+      download_url="https://download.visualstudio.microsoft.com/download/pr/5c281f95-91c4-499d-baa2-31fec919047a/38c6964d72438ac30032bce516b655d9/dotnet-sdk-3.0.100-osx-x64.pkg"
```
hohwille: whoa. Great cryptic download URL that is provided here. For all other tools we support to configure the desired version via some variable. See #110 for details. For VSCode I had the same cryptic URLs at the beginning and found out that it is also possible to find download URLs with semantic version name instead that redirects to the cryptic download URL.
Is it possible to have multiple .NET installations on the same machine and prevent a "global system" installation?
We already do that for Java, Maven, Gradle, Eclipse, VSCode, IntelliJ, Node.js, etc.
If we do not follow this for .NET we will have a completely different UX for .NET and will cause lots of confusion (see the features that we promise for devonfw-ide and that you do not keep here).

hohwille:
@jmsanmu and @josemasacapgemini any updates? Should we do a meeting together to figure out a proper way to solve this?

hohwille:
Is there any update or input from .NET stack on this? I can offer some support and we can do a meeting together...

sjimenez77:
@jmsanmu could you please check this PR again? Thanks!

hohwille:
Seems as if a direct automatic free download of .NET core is now possible:
https://dotnet.microsoft.com/download

hohwille:
IMHO we could now implement this story (starting from scratch).  See here:
https://dotnet.microsoft.com/download/dotnet/scripts

hohwille:
So as this issue has been stale, I had a look again.
There are download overviews on the web with links to individual OS specific download pages per version:
* https://dotnet.microsoft.com/en-us/download/dotnet/6.0
* https://dotnet.microsoft.com/en-us/download/dotnet/7.0

Again those links are still quite nicely and could be easily generated by `devonfw-ide` for automatic download:
* Windows: https://dotnet.microsoft.com/en-us/download/dotnet/thank-you/sdk-6.0.300-windows-x64-installer
* MacOS: https://dotnet.microsoft.com/en-us/download/dotnet/thank-you/sdk-6.0.300-macos-x64-installer
* Linux: https://dotnet.microsoft.com/en-us/download/dotnet/thank-you/sdk-6.0.300-linux-x64-binaries

However, all these pages then open a download URL via JavaScript that contains some cryptic IDs/hashes:
* Windows: https://download.visualstudio.microsoft.com/download/pr/deb4711b-7bbc-4afa-8884-9f2b964797f2/fb603c451b2a6e0a2cb5372d33ed68b9/dotnet-sdk-6.0.300-win-x64.exe
* MacOS: https://download.visualstudio.microsoft.com/download/pr/a048efaf-c4ee-4c22-a987-2bda2a58253c/3ddde00d1b2cf420a0d870df26bc56f8/dotnet-sdk-6.0.300-osx-x64.pkg
* Linux: https://download.visualstudio.microsoft.com/download/pr/dc930bff-ef3d-4f6f-8799-6eb60390f5b4/1efee2a8ea0180c94aff8f15eb3af981/dotnet-sdk-6.0.300-linux-x64.tar.gz

As you can see the cryptic hash or UUIDs seem to be specific per version + OS + architecture.
Therefore we would need to implement some magic that can resolve the actual download URL from the web.
Either we can find a direct "API" that is giving this to us if available or in the worst case, we would have to download the HTML and grep the actual download link from the HTML.

This approach is giving us some challenges/problems:
* We try to stay flexible via https://github.com/devonfw/ide-mirrors to have the possibility to react to changed download URLs without the need that all our users update to a specific version of devonfw-ide with according changes incorporated.
* We have a common function `doInstall` used by all existing commandlets that is able to determine the download URL from that `ide-mirrors` repo config and do the download and installation in a generic way. All this will not work for `DotNet` so far.

hohwille:
For the record: Inside the downloaded HTML the actual download link can be found in a line like this:
```
window.open("https://download.visualstudio.microsoft.com/download/pr/a048efaf-c4ee-4c22-a987-2bda2a58253c/3ddde00d1b2cf420a0d870df26bc56f8/dotnet-sdk-6.0.300-osx-x64.pkg", "_self");
```
I am not too confident that this remains stable and should be hardcoded into our code in `devonfw-ide` as microsoft may change the HTML structure at any time.

hohwille:
Replaced by PR https://github.com/devonfw/ide/pull/782

PR #296, opened 2019-10-31
Solved problems with cicdgen & devon4node installation

dario-rodriguez:
Fixed a bad conditional in the cicdgen installation.

Also, in order to be sure that the devon4node installation is using the npm inside the IDE, I added the lines to install & use npm that other commands are already using.

### DIFF
--- a/scripts/src/main/resources/scripts/command/node
+++ b/scripts/src/main/resources/scripts/command/node
@@ -33,15 +33,9 @@ function doSetup() {
 
 function doSetupDevon4node() {
   if ! command -v devon4node &> /dev/null
-  then  
-    doDevonCommand npm -q setup
-    local npm_command="${DEVON_IDE_HOME}/software/node/npm"
-    if [ ! -x "${npm_command}" ]
-    then
-      npm_command="${DEVON_IDE_HOME}/software/node/bin/npm"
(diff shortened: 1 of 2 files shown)


scripts/src/main/resources/scripts/command/node
```diff
@@ -34,8 +34,14 @@ function doSetup() {
...
   then  
+    doDevonCommand npm -q setup
+    local npm_command="${DEVON_IDE_HOME}/software/node/npm"
+    if [ ! -x "${npm_command}" ]
+    then
+      npm_command="${DEVON_IDE_HOME}/software/node/bin/npm"
+    fi
     doDevonCommand yarn -q setup
-    doRunCommand "npm install -g @devon4node/cli@latest" "install devon4node"
+    doRunCommand "'${npm_command}' install -g @devon4node/cli@latest" "install devon4node"
```
hohwille:
```
doDevonCommand npm install -g @devon4node/cli@latest
```
Would be even more consistent (see also line 37).

dario-rodriguez: you're right, and as the command that you suggest will install npm if it's not installed, I can remote the installation here.

Issue #316, opened 2019-12-11
devon ide settings url in batch mode


jambulud:
### Expected behavior

As a developer, I want to setup devonfw ide in quiet mode without specifying a settings url so that I can manage the setup with just one command.

### Actual behavior

When running `devon ide setup -b -q` it takes the `b` token as the settings url. 

![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/43032659/70617946-ff366c80-1c11-11ea-8729-c60b45af456d.png)

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1.  run `devon ide setup -b -q`

### Related/Dependent Issues

#315

### Comments/Hints:

if the command "devon" is not found run "./scripts/devon" (in general, the path of devon.bat).

`./scripts/devon ide setup ""  -b -q` works

### Affected version:

* OS: Windows 10
* Console: Powershell

hohwille:
@jambulud thanks for filing this bug report.
First of all I analyzed the problem:
* the `setup.bat` script for windows does not pass arguments at all what is a first bug that is trivial to fix.
* next `setup[.bat]` is invoking `devon ide setup` with the given arguments appended.
* therefore `setup -b -q myUrl` will result in `devon ide setup -b -q myUrl`
* the handling of the CLI options is currently implemented in `functions` that is sourced by all commandlets including `ide` commandlet.
* this currently only works if the options follow immediately after the commandlet.
* however to make it work this way the command would need to be `devon ide -b -q setup myUrl` but as stated above it is `devon ide setup -b -q myUrl`
* finally the `ide` commandlet does not honor `${batch}` and `${quiet}` for the `setup` what is yet another bug.

The first and the last issue are trivial to fix. However, the CLI issue is tricky and to be honest my CLI parsing code in `devonfw-ide` is really poor. Implementing a proper CLI parser framework in bash it IMHO out of question. An options would be to use `getopts` what is a build-in shell command.
The only drawback is that it does not support long options:
https://stackoverflow.com/questions/402377/using-getopts-to-process-long-and-short-command-line-options

This is kind of a pity but I think it would be acceptable to get rid of the long options then (what means that we remove features from our CLI API - however, as we only have more or less inofficial releases yet, it would be acceptable IMHO).

The only alternative would be to add some specific function for handling the standard options and call it from every commandlet while parsing arguments. What does not work it to work on $@ and use shift within a function as each function has its own arguments context.
[...]

Issue #261, opened 2019-09-30
Introduce DEVONFW_IDE_TRACE variable to enable tracing script output


markusschuh:
Edited: Updated the enhancement proposal to use DEVONFW_IDE_TRACE instead. 

As a developer of devonfw-ide I want to have some means for debugging script execution, so that issues can be narrowed down a bit easier.

batch / shell scripting has it limits when it is about debugging. Setup of an IDE that could be used for single step execution isn't trivial, especially not, if both Windows batches and bash scripts need to be debugged. So I propose the following enhancement - it just enables some poor man debugging option - but even this could help:

All `*.bat` scripts of devonfw-ide could get the following line after the initial `@echo off`:
```
if NOT "%DEVON_IDE_DEBUG%"=="" echo on
```

and all `*.sh` scripts the line
```
if [ ! -z "$DEVON_IDE_DEBUG" ]; then set -vx; fi
````
If that is added inside `scripts\functions` the command scripts, that source the `functions`, won't need a change.

With that configuration in place a developer could set `DEVON_IDE_DEBUG` to `y` (or any other value except the empty string) and afterwards the scripts echo each line before execution - the bash scripts additionally a second time, when everything in script line is expanded.

**Comment by hohwille** [comment]
nice idea. Planned for 3.3.0.

**Comment by hohwille** [comment]
I decided to add a debug option (`-d` or `--debug` can now be provided).
See also #315

**Comment by hohwille** [comment]
Fixed with PR #317

**Comment by markusschuh** [comment]
I reopen my enhancement proposal.

The debug option introduced in #317 just outputs a few more messages - only when function `doRunCommand` is used, what does not even yet seems to be used in all cases, where external commands are executed by the devonfw-ide scripts. 
But this does not really enable a quick narrow down, what is happening in the background.

Let's use #363 where my proposal, when realized, could help to narrow down some potential issue. 
Instead the currently already realized `-d`, that is:

```
devon ide -d update settings
```

does not help at help, to understand, what is going on in the background. There is the `bashdb` - bach debugger - but running this in the typical environments, where `devonfw-ide` is used, isn't trivial.

I admit, that  my above proposal might be too vague and imperfect. But why not then discuss a bit on this? For example it could be, that my intention is more to have a TRACE - not only DEBUG output of `devonfw-ide` execution sequence within the scripts.

So a better proposal might be to use in all `*.bat` scripts 

```
if NOT "%DEVONFW_IDE_TRACE%"=="" echo on
```

[...]

Issue #177, opened 2019-08-14
duplicate warnings on initial setup


maybeec:
Category: Bug or enhancement
Bug

Severity: Low , Medium, High, Critical
Medium

Expected behaviour (In case of bug)
If there is no settings path and no software package available, I will be asked after another to serve the download paths or simply continue to take some default from the internet.

Actual/current behaviour (In case of bug)
Currently I have multiple `*** ATTENTION ***` warnings continuing progressing something, that i think I made something wrog. Finally, the settings URL is asked by the system. Why is there so much attention needed if it is the default behavior. It seems like warnings.

In general from a user perspective, it is quite hard to differentiate the log from the system and the actual text necessary to be read by the user as of a question to be answered. It would be nice if logs could be somehow differentiated for example by a leading > or anything like this. 

Steps to reproduce (In case of bug)
Download beta.11 and run the setup

Description of issue / expected enhacement /Comments

Your environment - windows/Linux, Devonfw version , component
windows7

maybeec:
Just as another example, there comes a text `To be fully functional please rerun 'devon' command to update your PATH properly.` i the middle of the dowloading and setting up mvn. I would bet, that nobody is reading it as there is so much text.

couldn't it be even more sutable to make 90% of the logs just visible on a --verbose flag? I am just interested in the progress and on the step being progressed at the moment. Nothing more ;)

hohwille:
Use `-q` option for quiet output.
Everything else you are complaining here is prio minus infinity for me at the moment.
I also disagree that a an extra error log output is a bug. A bug means that something is not working.
Thanks for the feedback. I am just frank with my straight answer to clarify the expectation how fast these issues will be fixed. This project has already eaten tripple of my resources planned for it and so far I still have no other developers working on this.
Feel free to provide pull requests.

maybeec:
Ok, totally agree about the priority of this.
Anyhow, for the one who is implementing this later. It would be better to always put less into the log. Just what is necessary to be read by the user by default. If he wants to get more, you can simply provide a --debug or --verbose flag. The other way around with the quite flag does not make sense to me and would disturb 90% of the users.

hohwille:
I think there is still a trade-off to make.
Users will like the status output from `curl` when downloading tools as otherways they might think the installer crashed or hangs. However, already such output is already quite verbose.
Further bash is often creating very poor error messages with limited contect info.
You also need to be aware that I have to be able to give support for users running into issues.
[...]

PR #317, opened 2019-12-12
#316: fixed #315: quiet #261: debug mode #177: improve output

### LINKED ISSUE #315: Setup quiet mode
### Expected behavior

As a developer, I want to run the setup process in quiet mode so that I setup devonfw ide without questions.

### Actual behavior

Neither `setup --batch --quiet` nor `setup --quiet` avoids the "Settings url" question:
MicrosoftTeams-image

### Steps to reproduce bug

1. run `setup --batch --quiet` in a powershell console.

### Related/Dependent Issues

#316

### Comments/Hints:

When setup finishes it displays "Press any key to continue..." because the script has a harcoded "pause"

### Affected version:

* OS: Windows 10
* Console: Powershell

hohwille:
Fixes:
* #316 
* #315 
* #261 
* #177

### DIFF
--- a/scripts/src/main/resources/scripts/command/help
+++ b/scripts/src/main/resources/scripts/command/help
@@ -31,6 +31,7 @@ else
     doDevonCommand "${1}" help
     echo " -b | --batch         run in non-interactive mode (do not ask any questions)."
     echo " -q | --quiet         be quiet and avoid output."
+    echo " -d | --debug         be extra verbose and print more output for debugging."
     echo " -f | --force         force proceeding even in case of warnings."
   fi
 fi
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -29,21 +29,26 @@ function doUpdateSettings() {
     if [ -n "${1}" ]
     then
       SETTINGS_URL="${1}"
-    fi    
+    fi
     if [ -z "${SETTINGS_URL}" ]
     then
-      echo "Missing your settings at ${SETTINGS_PATH} and no SETTINGS_URL is defined."
-      echo "Further details can be found here:"
-      echo "https://github.com/devonfw/ide/blob/master/documentation/settings.asciidoc"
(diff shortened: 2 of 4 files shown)


jambulud:
Okay, I have checked "setup -b -q" a second time and I'm not longer getting any question, which is nice.

image


But there is still the hardcoded pause in setup.bat

I've also checked `devon ide setup -b -q`  (after running setup script )but I get

```
/usr/bin/bash: ./devon: No such file or directory
```

running `path-to-devon-ide-folder/scripts/devon ide setup -b -q` works fine, so I think that the environment variables are not being correctly set.

jambulud:
Sorry, I accidentally closed it.

hohwille:
> But there is still the hardcoded pause in setup.bat

Many users will just double-click on this script in windows. Without the pause a potential error message just dissapears and the user is lost.
What I could try is to also check if say `-b` was given as first argument and in that case omit the pause command as a specialy feature for your use-case. I will not consider if `-b` is given as second arg or if `--batch` is provided instead as actually the `setup` script is more or less just a delegation to `devon ide setup` and helps as a single click solution for newbee users to get started.
I will update this PR and think then it is complete.

hohwille:
Any further comments? Can I merge this PR now?

hohwille:
> so I think that the environment variables are not being correctly set.

Did you run the `devon` command within an exisitng shell?
The `devonfw-ide` can not change the environment variables of processes that have already been started. When you open a new terminal (e.g. git bash or CMD) after `setup` then all should work and `devon` command should be available. There is IMHO nothing I can do about this in `devonfw-ide`.

jambulud:
> > so I think that the environment variables are not being correctly set.
> 
> Did you run the `devon` command within an exisitng shell?
> The `devonfw-ide` can not change the environment variables of processes that have already been started. When you open a new terminal (e.g. git bash or CMD) after `setup` then all should work and `devon` command should be available. Did you try restarting dashboard? There is IMHO nothing I can do about this in `devonfw-ide`.

You are right, after the restart it works. 

I don't have any other comments, you can merge it

Issue #310, opened 2019-11-14
honor maven skip.tests option


hohwille:
As a developer, I want to have optimized builds so that I have fast CI.

As we did not find a unit- and integration-test framework for bash, we wrote this on our own. As this is not honoring the `-DskipTests` or `-Dmaven.test.skip=true` flag. 

This results in the tests performed twice by `travis-ci`. As this is a pure waste of resources, we want to change this and honor the flags to skip tests.

hohwille:
The issue with travis and duplicated tests can be resolved easily by setting `install: skip` in travis.
Actually this is something that we should do in any travis config invoking maven as this default of travis can lead to various confusions and errors.

hohwille:
Done with PR #319

PR #319, opened 2019-12-12
#310: skip duplicate tests in travis and honor skipTests variable

hohwille:
Fix for #310


hohwille:
Nice, I checked the travis log and tests are now executed only once.
Next is to omit the tests following maven standards.

Issue #325, opened 2019-12-18
Do not respect global Java installation


hohwille:
As reported by @markusschuh on Yammer, the `setup` of Java may find a globally installed Java. As `devonfw-ide` is supposed to work isolated and portable and guarantee that a project can use exactly the Java version that is configured for that project, this is to be changed.

markusschuh:
I checked, what I wrote in yammer on Dec 16, 2019. This was a comment to some issue, where the download URL for java setup was calculated as "-"  and this may have been related to a java preexisting outside of the devonfw-ide base.  As far as I can remember, it was more likely, this external java was introduced via PATH - not via explicit call via JAVA_HOME. But I may miss something.

Anyway I agree, this issue should be close for now. If there exists a specific log, when an externally found component overwrites an internal devonfw-ide software component and causes issues, another issue can be opened.

hohwille:
Closing as invalid then.

PR #326, opened 2019-12-18
WIP: #325: ensure to use Java from software folder ignoring global installation

hohwille:
See #325


hohwille:
Maybe I should even remove this line:
https://github.com/devonfw/ide/blob/b1d9fda269c4c3fb0f4009b5f3cc12b01db36029/scripts/src/main/resources/scripts/devon.properties#L12
And instead extend `environment-project[.bat]` to do this `export JAVA_HOME=${DEVON_IDE_HOME}/software/java` if that folder is present of `java` is included in `${DEVON_IDE_TOOLS}`.
WDYT?
At least configuring JAVA_HOME to something else might only cause problems. Or might that be a feature for a project that needs different versions of Java in parallel and could configure the JAVA_HOME per workspace???

hohwille:
I am still unsure about this. Still I can not see how to reproduce #325. Maybe this ticket can even be closed as invalid. If all is working, why would we need to take away flexibility? E.g. what if for some reason a project would require OracleJDK that may be preinstalled under a given path for every developer in the team working on the same OS. Currently they could overwrite JAVA_HOME to that location.

markusschuh:
I also would suggest to close this PR, since the original issue is a bit vague and the consequences of this concrete change might need more testing - but also more documentation.
At the end it is of course an advantage, when also some externally installed tools could be used by devonfw-ide. It is just important, that the needed control and transparency exists. This should not happen by chance and the developer needs some kind of feedback / debug method to identify the difference in case, something unexpected happens.

hohwille:
Closing. See #325

Issue #332, opened 2020-01-09
Can't put setting_url manually in the case of swtbot automation


Rohit90700:
Ref link  -https://github.com/devonfw/swtbot/issues/1

Reason of this issue creation is , if we make devon-ide setup automated, for example -

1.  Download, complete devon-ide package
2. its extraction
3. run setup.bat       ( these all things should be automated)

In the case of step 3) - run setup.bat ,its need  setting_url to be entered manually,  we cannot provide manual intervention in jenkins.
Then idea is, we will pass extra parameter to setup.bat  for swtbot automation and for regular setup if parameter is not there then it will work as it is.

hohwille:
This is wrong. You can supply "-" as default. Can you just have a look at the existing integration tests that already solve all such issues?

hohwille:
https://github.com/devonfw/ide/blob/master/scripts/src/test/bash/test-setup#L19

PR #333, opened 2020-01-09
setting_url is default in the case of swtbot automation

Rohit90700:
This PR is created for the issue - https://github.com/devonfw/ide/issues/332

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -42,7 +42,13 @@ function doUpdateSettings() {
         echo "Please contact the technical lead of your project to get the SETTINGS_URL for your project."
         echo "In case you just want to test devonfw-ide you may simply hit return to install default settings."
         echo
-        read -r -p "Settings URL: " answer
+        if [ "${DEFAULT_SETTING_URL}" == "default" ]
+        then
+          echo  "Settings URL:"  "${DEFAULT_SETTING_URL}"
+          SETTINGS_URL="-"
+        else
+          read -r -p "Settings URL: " answer
+        fi
         if [ -z "${answer}" ]
         then
           SETTINGS_URL="-"


hohwille:
> testMigrations(com.devonfw.tools.ide.migrator.MigrationsTest): Server returned HTTP response code: 403 for URL: https://repo1.maven.org/maven2/com/devonfw/java/modules/devon4j-basic/maven-metadata.xml

Looks not like an error in your changes but some stability issue with travis.
I will rerun the build in travis and we will see if the error goes away...

hohwille:
Travis build fixed. I will review and merge if fine...

scripts/src/main/resources/scripts/command/ide
```diff
@@ -42,7 +42,13 @@ function doUpdateSettings() {
...
         echo "In case you just want to test devonfw-ide you may simply hit return to install default settings."
         echo
-        read -r -p "Settings URL: " answer
+        if [ "${DEFAULT_SETTING_URL}" == "default" ]
+        then
+          echo  "Settings URL:"  "${DEFAULT_SETTING_URL}"
+          SETTINGS_URL="-"
+        else
+          read -r -p "Settings URL: " answer
+        fi
```
hohwille: why do you need this change after all?
Instead of supplying "default" you can already supply "-" and you do not need this change at all.

hohwille:
I closed #332 as invalid. Can I also close this PR or am I missing something?

scripts/src/main/resources/scripts/command/ide
```diff
@@ -42,7 +42,13 @@ function doUpdateSettings() {
         echo "Please contact the technical lead of your project to get the SETTINGS_URL for your project."
         echo "In case you just want to test devonfw-ide you may simply hit return to install default settings."
         echo
-        read -r -p "Settings URL: " answer
+        if [ "${DEFAULT_SETTING_URL}" == "default" ]
+        then
+          echo  "Settings URL:"  "${DEFAULT_SETTING_URL}"
+          SETTINGS_URL="-"
+        else
+          read -r -p "Settings URL: " answer
```
Rohit90700: As we know when control comes on the line 50  'read -r -p "Settings URL: " answer' its need manual setting_url but in the case of automation we can't put manual anything that's why i am trying to bypass this line with the lines of code which i have added.
Parameter 'DEFAULT_SETTING_URL' and its Value 'default' i am sending from other bat file which have added through java( swtbot automation java part). Line 50 only will be bypass in case of swtbot automation ,In other case work as it is.

Rohit90700: Can we do like this -
echo
        if [ "${answer}" == "-" ]
        then
          SETTINGS_URL="-"
        else
          read -r -p "Settings URL: " answer
        fi
instead of above code ,is it ok?

hohwille: There is nothing you need to change at all. It is all already working.
Just call setup with "-" as parameter as I already stated.
Could you read my comment in the issue and follow the link to the existing integration test.
Wenn discussion started about swtbot for devonfw-ide I pointed out in daily standup that people should first have a look at the existing integration tests. It would be lovely if you could first have a look at this so we can avoid problems that are already solved. Thanks.

Issue #259, opened 2019-09-25
mvn package runs into test failures on Windows


markusschuh:
### Expected behavior

As a developer, I want to generate my own package including execution of included tests so that I can rely on a behavior of the software being compliant to the realized tests.

### Actual behavior

A `mvn package` runs into test failures due to use of different conventions for path names during com.devonfw.tools.ide.configurator.ConfiguratorTest

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Use windows 10 as build environment
1. use `mvn 3.6.1`
1. `git clone ...`
1. `cd ide`
1. `mvn clean package`

This results in errors as listed at end of issue.

### Comments/Hints:

As a workaround I can skip the tests.  ( `-Dmaven.test.skip=true`)

As far as I understand the ConfiguratorTest code, this mismatch originates from the following difference:

- The expecting value is generated from the Java code using the value as read from
```
private static final String DEVON_IDE_HOME = new File("").getAbsolutePath();
```
- The compared value is generated by maven via expansion of the falling variable from the template for settings.json: `${client.env.home}`

The first one generates - on windows - pathname using a single back slash, the second pathname (by maven) using double back slashes- valid in context, where back slash is an escape string.  If there were some "compare canonical pathname" instead of a string compare, both would match. Currently they do not - on Windows only.

### Affected version:

* OS: Windows/Linux/Mac?
  Windows 10 (1903)
* Browser: Chrome/Firefox/Safari?

```
Apache Maven 3.6.1 (d66c9c0b3152b2e69ee9bac180bb8fcc8e6af555; 2019-04-04T21:00:29+02:00)
Maven home: ...
Java version: 11.0.4, vendor: AdoptOpenJDK, runtime: C:\Program Files\AdoptOpenJDK\jdk-11.0.4+11
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"
```

An except of the output:

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.devonfw.tools.ide.configurator.ConfiguratorTest
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.312 sec <<< FAILURE!
testConfigurator(com.devonfw.tools.ide.configurator.ConfiguratorTest)  Time elapsed: 0.24 sec  <<< FAILURE!
java.lang.AssertionError:
File:
[...]

PR #334, opened 2020-01-09
#259: fix for windows backslash vs. slash issues

hohwille:
Fix for #259


hohwille:
I fixed the JUnits by normalizing paths. However, I am not yet 100% sure if all tools (Eclipse and all plugins) will properly abstract between `/` and `\\`. We would IMHO need some manual testing of this change.

hohwille:
Manually tested in windows VM. Looks good.

PR #353, opened 2020-02-14
Implementaton of automated git checkout and eclipse import

maihacke:
Implementaton of automated git checkout and eclipse import
see https://github.com/devonfw/ide/issues/175

PR #357, opened 2020-02-20
#175: improvement for import

hohwille:
Improvement for #175 and #353

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -325,14 +325,14 @@ function doUpdateProjects() {
     doDebug "Pull or clone git project ${project_path}..."
     project_workspace_path="${DEVON_IDE_HOME}/workspaces/${project_workspace}"
     mkdir -p "${project_workspace_path}"
-    cd "${project_workspace_path}"
+    cd "${project_workspace_path}" || exit 255
     doGitPullOrClone "${project_path}" "${project_git_url}" "${project_git_branch}"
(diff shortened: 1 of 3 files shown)


scripts/src/main/resources/scripts/command/ide
```diff
@@ -306,27 +306,28 @@ function doUninstall() {
...
     
     [ -z "${project_path}" ] && doFail "Config is missing required parameter 'project.path'"
-    [ -z "${project_workspace}" ] && local project_workspace=main
+    [ -z "${project_workspace}" ] && project_workspace=main
     [ -z "${project_git_url}" ] && doFail "Config is missing required parameter 'project.git.url'"
-    [ -z "${project_eclipse}" ] && local project_eclipse=import
+    [ -z "${project_eclipse}" ] && project_eclipse=import
     
-    echo "Project (Path: ${project_path}, Url: ${project_git_url}, Branch: ${project_git_branch}, Eclipse: ${project_eclipse}, Workspace: ${project_workspace}, Workingsets: ${project_workingsets})"
-    if [ "${project_workspace}" == "${WORKSPACE}" ]
```
maihacke: why did you remove the if-statement?  now all projects will be imported into all workspaces...

hohwille: No each project is configured into the workspace that is configured.
Otherwise only the projects for the main workspace where perocessed during setup and all others had been ignored.

hohwille: OK. We now have clarified this in a call. Thanks as this revealed gaps on both of our sides so we can now fix everything properly.

PR #364, opened 2020-03-15
#363: Add git pull for already existing ide settings fixes #363

### LINKED ISSUE #363: devon ide update does not pull latest project settings
Expected:
1. Architect creates project settings repository
2. Developer uses "setup" command  with settings repository url to setup devon ide with project settings
3. Architect updates project settings repository
4. Developer updates devon ide using "devon ide update"
5. Devon ide pulls latest settings from project settings
6. Devon ide updates components according latest settings defined by architect

Actual: 
1. Architect creates project settings repository
2. Developer uses "setup" command  with settings repository url to setup devon ide with project settings
3. Architect updates project settings repository
4. Developer updates devon ide using "devon ide update"
5. Devon ide does not pull latest settings from project settings
6. Devon ide updates components according initial settings state set defined by architect

### Affected version:
devonfw-ide-scripts 3.2.4
devonfw 3.2.1

* OS: Windows

markusschuh:
The current code for `devon ide update settings` only covers the case, there is no settings sub-directory yet.  This change adds handling for the case, it already exists:  The already existing function `doGitPullOrClone` is called now and will ensure, a `git pull` is executed to update the settings from the origin repository, when possible.

Since I don't actively host a real settings repository, I only have executed a simple test. Technically this change covers the "update settings" (with already existing settings dir) now. I'm just a bit unsure, if there might be some corner case - e.g. specific changes in settings - that may need some additional handling after the pure `git pull`. 

E.g. some of the *.properties files are copied to location outside of the settings sub-directory on initial setup. Does that needs to be re-executed after this update also? How to handle potential overwrites of a user-only change within the target file in this case?

I assume, this needs a bit more work to make this PR more complete.

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -73,6 +73,8 @@ function doUpdateSettings() {
       git commit -m "Initial settings downloaded from ${SETTINGS_URL}"
       cd - || exit 255
     fi
+  else
+    doGitPullOrClone "${SETTINGS_PATH}"
   fi
 }
 


hohwille:
OMG, I was not aware that this is missing. Acutally I always thought that the settings are already updated via git pull on `devon ide update`.
Thanks for this PR. I took this out of Draft state, as I think this is an important bugfix.

One thing to notice for further improvement/correctness:
After updating the settings the config has to be reloaded as `settings/devon.properties` may have changed. See here:
https://github.com/devonfw/ide/pull/341/files

hohwille:
To complete the release today, I created PR #367 that replaces this PR.
Thanks @markusschuh for this PR. I took your solution and improved it.

PR #373, opened 2020-03-25
Fix project import on windows

maihacke:
This fixes the probelm that eclipse could not be started to import projects for some users on windows.


hohwille:
can you keep this line:
```
# shellcheck disable=SC2086
```

maihacke:
> can you keep this line:
> 
> ```
> # shellcheck disable=SC2086
> ```

Yes, sure.

Issue #372, opened 2020-03-21
Do not setup devonfw projects by default


maybeec:
### Expected behavior

As a user, I want to use the IDE for a specific devonfw project so that I can start developing easily.

### Actual behavior

Downloading the devonfw ide and installing it results in an endless processing of multiple devonfw projects like devon4j, mts, ... although I might not want to have any or just one of the projects to be checked out for editing

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Download devon ide 3.3.0
2. run setup.bat

### Comments/Hints:

It might be worth to ask the user after initialization, whether or which projects he wants to be setup.

### Affected version:

* OS: Windows

hohwille:
@maybeec Thanks for this feedback. 
First of all devonfw-ide is fully configurable for each project via the settings git.
So in the end the question is about the default settings.
Here we have contradicting demands as it seems:
* Some people told us to work on supporting that devonfw-ide has MTS imported automatically by default
* You are now telling to do the opposite.

I am not biased for any of these choices but I think we have to discuss this with the entire community to make the decision or to even think of a compromis (e.g. asking the user or whatever).

hohwille:
> Downloading the devonfw ide and installing it results in an endless processing 

It is not "endless" but surely it takes quite some time. We have already been discussing about omtimizing the import of multiple projects so that eclipse in not launched multiple times.

> of multiple devonfw projects like devon4j, mts, ... 

Further by default exactly two projects are imported: "devon4j" and "mts".

maybeec:
Yes you are right and I Totally agree that the default should be devonfw focused.
I just raised this issue as currently it takes ages to setup the ide as Devon4J and mts is build and tested by maven during the process, which btw also raises errors if tests are failing.
In my opinion, tests execution should definitely not be part of the provisioning.

hohwille:
So what I could offer as a quickfix:
* Create a fork of `ide-settings` repository without the projects configured
* Add the URL to that forked git to the documentation. Then users who do not want to setup their own settings (what is still the recommended way to start) can easily use that URL to avoid the "endless setup".

maybeec:
Yes, that's the general way to go.
But as a general question, why to you want to run all tests of all repositories on ide provisioning? I don't see the reason and it even does not scale

maihacke:
We could add -DskipTests=true to the initial IDE setup

hohwille:
> We could add -DskipTests=true to the initial IDE setup

[...]

PR #374, opened 2020-03-25
Rework build on import

maihacke:
Proposal to fix https://github.com/devonfw/ide/issues/372

Users have to provide params for the build command (e.g. dependency:go-offline to just download libs). Without params build is skipped.

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -320,14 +320,14 @@ function doUpdateProjects() {
     last_cwd="${PWD}"
     for projectconfig in "${SETTINGS_PATH}"/projects/*.properties; do
       doEcho "Importing from ${projectconfig}..."
-      project_path=$(grep "^path" "$projectconfig" | cut -d'=' -f2)
-      project_workingsets=$(grep "^workingsets" "$projectconfig" | cut -d'=' -f2)
-      project_workspace=$(grep "^workspace" "$projectconfig" | cut -d'=' -f2)
-      project_git_url=$(grep "^git.url" "$projectconfig" | cut -d'=' -f2)
-      project_git_branch=$(grep "^git.branch" "$projectconfig" | cut -d'=' -f2)
-      project_build_path=$(grep "^build.path" "$projectconfig" | cut -d'=' -f2)
-      project_eclipse=$(grep "^eclipse" "$projectconfig" | cut -d'=' -f2)
-      project_build_params=$(grep "^build.params" "$projectconfig" | cut -d'=' -f2)
+      project_path=$(grep "^path" "$projectconfig" | cut -d'=' -f2-)
+      project_workingsets=$(grep "^workingsets" "$projectconfig" | cut -d'=' -f2-)
+      project_workspace=$(grep "^workspace" "$projectconfig" | cut -d'=' -f2-)
+      project_git_url=$(grep "^git.url" "$projectconfig" | cut -d'=' -f2-)
+      project_git_branch=$(grep "^git.branch" "$projectconfig" | cut -d'=' -f2-)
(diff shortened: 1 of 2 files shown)


maihacke:
@hohwille  there are changes in the ide script I did not intend 3.3.0 vs variable. Is there any magic about this aspect in place I do not know?

scripts/src/main/resources/scripts/command/ide
```diff
@@ -320,30 +320,35 @@ function doUpdateProjects() {
...
       mkdir -p "${project_workspace_path}"
       cd "${project_workspace_path}" || exit 255
       doGitPullOrClone "${project_path}" "${project_git_url}" "${project_git_branch}"
-      if [ -n "${project_build_path}" ]
+      doDebug "Building project with params: ${project_build_params}"
+      if [ -n "${project_build_params}" ]
       then
         cd "${project_path}/${project_build_path}" || exit 255
-        doDevonCommand build
+        doDevonCommand build ${project_build_params}
```
maybeec: But that's just passing parameters.
As of our last discussion, I even thought of running doDevonCommand ${project_post_build_command} and let the user specify what to do. So a user could even decide to do something like `py doSomeGreatSetup`.

Even the question is here whether we should also support anything like also git bash standard commands, such that I could process something like an sed command or anything like this.

maihacke: The problem is, that a user could also provide `rm -rf /` as build command then...

maybeec: Ok. My view on that is simple:
With great power comes great responsibility. I don't think that we have to discuss about attackers but just about faulty configurations. But in my opinion it's quite obvious that you should not put here any cd / && rm -Rf :)
But for me it would be even ok if we simply skip execution possibilities of any non devonfw command.

maihacke: Then you would propose:
`doDevonCommand ${project_build_cmd}`

maybeec: yes

maihacke: et voilà!

scripts/src/main/resources/scripts/command/ide
```diff
@@ -320,30 +320,36 @@ function doUpdateProjects() {
...
+      [ -z "${project_build_path}" ] && project_build_path=.
     
       doDebug "Project (Path: ${project_path}, Url: ${project_git_url}, Branch: ${project_git_branch}, Eclipse: ${project_eclipse}, Workspace: ${project_workspace}, Workingsets: ${project_workingsets})"
       doDebug "Pull or clone git project ${project_path}..."
       project_workspace_path="${DEVON_IDE_HOME}/workspaces/${project_workspace}"
       mkdir -p "${project_workspace_path}"
       cd "${project_workspace_path}" || exit 255
       doGitPullOrClone "${project_path}" "${project_git_url}" "${project_git_branch}"
-      if [ -n "${project_build_path}" ]
+      doDebug "Building project with params: ${project_build_params}"
```
maybeec: old reference in debug log

hohwille:
@maihacke 
What to do about the default settings? Is the import with m2e working for a fresh IDE installation with devon4j and mts? I expect it to fail so we would still need a change in `ide-settings` to add the `build.command`, right?

maihacke:
> @maihacke
> What to do about the default settings? Is the import with m2e working for a fresh IDE installation with devon4j and mts? I expect it to fail so we would still need a change in `ide-settings` to add the `build.command`, right?

Yes that's true. I checked a fresh install, without "pre-loading" libgs via `mvn`, m2e fails to download the deps...

maihacke:
I changed my local settings, should I push that now the PR is merged?

Issue #377, opened 2020-03-27
Imporve eclipse import


hohwille:
As a devonfw-ide user and maintainer, I want to use the Eclipse import feature so that the setup is ready for the project. However, currently we are facing the following drawbacks and limitations:

* Importing multiple projects is very slow as for each project the entire Eclipse IDE is launched and initialized with all plugins, etc. then one project is imported and Eclipse is shutdown. Instead it would be great to only launch Eclipse once and import all projects in one go.
* Importing a project is not possible while the IDE is already launched. When I create a new project and want to add it to Eclipse, I have to shutdown Eclipse, do the import what launches Eclipse again, performs import, shuts down Eclipse and then I have to start Eclipse again to continue working. This is taking too much time.

So instead I think we should still consider the original design I proposed in https://github.com/devonfw/ide/issues/175#issuecomment-574191167

hohwille:
The original design linked above could solve the first point.
For the second point, we would process the queued imports whenever eclipse is launched.
To immediately trigger the import even when Eclipse is already open, we IMHO would need to write our own Eclipse plugin.

hohwille:
I also tested the current import and still the projects are broken and have red error icons when you start Eclipse.
I noticed this output:
```
Total time: 11 seconds
Job found still running after platform shutdown.  Jobs should be canceled by the plugin that scheduled them during shutdown: org.eclipse.mylyn.internal.monitor.ui.MonitorUiPlugin$3
Job found still running after platform shutdown.  Jobs should be canceled by the plugin that scheduled them during shutdown: org.eclipse.ui.internal.views.markers.UIUpdateJob
Job found still running after platform shutdown.  Jobs should be canceled by the plugin that scheduled them during shutdown: org.eclipse.mylyn.internal.tasks.ui.TasksUiPlugin$TasksUiInitializationJob
Job found still running after platform shutdown.  Jobs should be canceled by the plugin that scheduled them during shutdown: org.eclipse.jdt.internal.ui.InitializeAfterLoadJob
Job found still running after platform shutdown.  Jobs should be canceled by the plugin that scheduled them during shutdown: org.eclipse.egit.ui.Activator$1
Job found still running after platform shutdown.  Jobs should be canceled by the plugin that scheduled them during shutdown: com.genuitec.eclipse.startup.workspace.WorkspaceTracker$EarlyStartup$1
[...]

PR #392, opened 2020-04-03
Wait for eclipse jobs to finish

maihacke:


maihacke:
This addresses problems described in #377

scripts/src/main/resources/scripts/lib_script/import.groovy
```diff
@@ -61,12 +76,44 @@ class MyWorkbenchAdvisor extends org.eclipse.ui.application.WorkbenchAdvisor {
...
+      do {
+        importantJobsRunning = false;
+        System.out.println("Still running:");
+        Job[] jobs = jobMan.find(null);
+        for (Job job : jobs) {
+          importantJobsRunning |= (importantJobClassNames.contains(job.getClass().getName()));
+          System.out.println(job.toString() + ": " + job.getClass().getName());
+        }  
+        Thread.sleep(3000);      
+      } while (importantJobsRunning);
```
maybeec, line 116: I think it's a smart idea to wait for specific job names, but in my opinion, that also makes the maintenance harder in case there are some classes renamed/packages renamed of the plugins, which is hard to detect.
Also eclipse jobs are triggered after another and even by other jobs, so it might be, that any of the jobs might even just being triggered when actually being checked for the running status. This might be some very rare cases. But nevertheless.

My proposal wouldbe to simply wait until the jobmanager has processed all jobs: `Job.getJobManager().isIdle()`. This might take a little longer, but you could be sure, that also future "jobs" to fix some other issues or bring some new functionality into eclipse will be executed properly.

Also this might fix the refresh problem, you faced as maybe some finishing jobs are taking care of workspace polishing?

maihacke: I tried `Job.getJobManager().isIdle()` but it seems that it is never `true` meaning there is always some job running.
Maybe I can check what eclipse is normally doing to assure a clean shutdown. Currently we call `PlatformUI.getWorkbench().close();` maybe there is something more todo...

maybeec: Ok, that's strange, we are using this in our swtbot tests all the time: https://github.com/devonfw/cobigen/blob/master/cobigen-eclipse/cobigen-eclipse-test/src/main/java/com/devonfw/cobigen/eclipse/test/common/swtbot/AllJobsAreFinished.java 

It has been turned out, that this is the most stable solution, but it honestly takes a little longer. Eclipse seems to do quite some stuff. Our timeouts for that are at 10s, working stable to wait for eclipse. But obviously, it also depends on the number of plugins you installed what the list of tasks is. :(

maihacke:
I tried @maybeec suggestion to use `Job.getJobManager().isIdle()`. It turned out that this is never `true` in this context, so the import wouldn't finished. The current solution is not an object lesson for maintainability but it works. So from my point of view this can be merged.

maihacke:
This PR should also address the issues in https://github.com/devonfw/ide/issues/377

hohwille:
I consider this PR as a nice-to-have improvement for the eclipse import. It IMHO will not make anything worse but only better. Hence I will include and merge it.

PR #398, opened 2020-04-06
Fixed IntelliJ launch location for Windows

phschimm:
Tested on Windows 7, 10.


hohwille:
Sorry for the bug and thanks for this PR with a fix 👍 
I have to double-check but my assumption is that IntelliJ version has been updated in devonfw-ide without proper testing and the directory layout with bin folder has meanwhile changed. Also if the start script is to be expected in `bin` folder, I would also change the MacOS quirk to create the script in the same folder also on MacOS then we do not need to distinguish the location per OS.

hohwille:
@phschimm thanks for the fix. I have created PR #402 from it that replaces this PR.

PR #402, opened 2020-04-20
#398: fixed idea script location, normalized to bin folder across all OS

hohwille:
Replacement for PR #398.


hohwille:
I make my fix downward compatible but assure that we will have the executable always on `${IDEA_HOME}/bin/idea` for all OS.
Still it would be great if someone could test this on linux. I have never used IntelliJ on linux at all...

PR #403, opened 2020-04-23
Updated node.js and VS Code versions and Ionic support added

sjimenez77:
- node.js 12.16.2 LTS
- VS Code 1.44.2
- Ionic support included
- Some bug fixes


scripts/src/main/resources/scripts/devon.properties
```diff
@@ -4,7 +4,7 @@
 # For details see https://github.com/devonfw/ide/blob/master/documentation/configuration.asciidoc
 #************************************************************************************************
 
-DEVON_IDE_TOOLS=(java mvn eclipse vscode node npm ng)
+DEVON_IDE_TOOLS=(java mvn eclipse vscode node npm ng ionic)
```
hohwille, line 7: we are extending the defaults more and more while we get the feedback that we should rather follow opt-in than opt-out approach.

hohwille:
Maybe I merge this and afterwards remove ionic from tools variable to complete the release as fast as possible. Many projects wont be using ionic so we should not overbloat the setup with tools never needed. Any user can at any time just type `devon iconic ...` and it will automatically be installed. So actually tools variable should be rather minimal... And projects can still customize the tools variable to their needs in settings git. However, most interested users will use the default settings when they try devonfw-ide for the first time and already installing eclipse and vscode seems to be too much and take too long... Same for preinstalled projects in Eclipse...

sjimenez77:
Ok, perfect. In that case we could remove it. Maybe in the documentation we should add the steps to enable `ionic` in the IDE.

PR #417, opened 2020-05-22
Update yarn

OliverNeff:
`install` has been replaced with `add` to add new dependencies.

### DIFF
--- a/scripts/src/main/resources/scripts/command/yarn
+++ b/scripts/src/main/resources/scripts/command/yarn
@@ -55,9 +55,9 @@ function doCheckTopLevelProject() {
 function doRunBuild() {
   doSetup
   local extra_arg=""
-  if [ ! -d node_modules ] && [[ "${*}" != *install* ]]
+  if [ ! -d node_modules ] && [[ "${*}" != *add* ]]
   then
-    extra_arg="install"
+    extra_arg="add"
   fi
   doEcho "Running: yarn ${extra_arg} ${*}"
   if doIsQuiet


hohwille:
Can you give a link or more details? Don't we need downward compatibility?
So would we need to check `YARN_VERSION` and only if newer than `«x.y.z»` apply this change?

hohwille:
@OliverNeff I assume this is a misunderstanding.
* `yarn add` is for adding a dependency: https://classic.yarnpkg.com/en/docs/cli/add/
* `yarn install` is for building: https://classic.yarnpkg.com/en/docs/cli/install/

I think this PR is invalid and can be closed.
Agreement or do you have insights what I might be missing?

OliverNeff:
@hohwille,
I installed the IDE with YARN and not NPM and become this installation in the installation process. When I changed it to add it worked for me.

hohwille:
Feedback from @dario-rodriguez on teams:
> What is the purpose of the doRunBuild function? Build the application? In that case you need to call to yarn build (build script must be present in package.json). If node_modules folder is not present you must call yarn install (or only yarn) before you execute the yarn build command. So, in my opinion, both implementations (the PR and the previous code) are wrong.

I fully agree and therefore my suggestion is to close this PR as invalid and create a new issue/PR to fix this properly.

hohwille:
Created issue #431 and closing this PR. We have to create a new PR to fix #431.

hohwille:
For the record. I fully understood the problem only while implementing the fix. So unlike in maven (`mvn`) the JavaScript build tools like `npm` and `yarn` do not allow to simply concatenate goals as the goals may take parameters.
Therefore `devon install build` is not first running `install` and running `build` script, but will add and install the dependency `build` as it is provided as argument to `install`. This is also leading to the message that instead `add` shall be used and this fully explains the change proposed in this PR.
However, the correct fix was:
* prevent to implicitly add `install` goal if `node_modules` is not present
* for build command simply first run `install` and then in a second invocation call `run build` (but only if `build` section is present in `package.json` under `scripts`).

So again thanks for the PR and contribution. It took a while but it finally lead to a fix and helped improving `devonfw-ide` 👍

Issue #418, opened 2020-05-25
Make projects optional


hohwille:
For plugins we have the ability to make them optional.
The same should apply for projects what is currently not yet supported. This way we should make devon4j and mts optional projects as discussed in #377

hohwille:
Done with PR #419

PR #419, opened 2020-05-25
#418: #377: added project commandlet, allow project to be inactive

hohwille:
Implementation of #418

### DIFF
--- a/scripts/src/main/resources/scripts/command/project
+++ b/scripts/src/main/resources/scripts/command/project
@@ -1,0 +1,107 @@
+#!/usr/bin/env bash
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+# $1: project to setup (omit for all all projects)
+function doSetup() {
+  if [ -n "${1}" ]
+  then
+    local project_file="${1}"
+    if [ ! -f "${project_file}" ]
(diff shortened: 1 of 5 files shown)


scripts/src/main/resources/scripts/command/project
```diff
@@ -0,0 +1,107 @@
...
+  then
+    return
+  fi
+  project_path=""
+  project_workingsets=""
+  project_workspace="main"
+  project_git_url=""
+  project_git_branch=""
+  project_build_path="."
+  project_eclipse="import"
```
maybeec, line 48: I was wondering about this default by the way. It was making much trouble for us finding out how to deactivate eclipse import without reading the code. Even as this feature is not reliably working. I would suggest to remove the default = import.

Anyhow if you like to have a separate pr for that I will go for it.
Obviously, this will cause a changed behavior for ide upgrades.

hohwille: Default changed as requested.

maihacke:
I think the active flag is a bit redundant, since disabling build and eclipse-import should be the same as active=false. But that is more a matter of taste.
Thanks for working on this feature. I like that is now separate command and not part of the ide script.

Issue #395, opened 2020-04-03
variable from devon.properites unset if value is in double quotes


hohwille:
### Expected behavior

I can write something like this in my `devon.properties`:
```
key="my value"
```
And when I run `devon` command in that IDE the variable `key` should be set to `my value`.

### Actual behavior

The variable `key` will be unset (so if `key` is actually `M2_REPO` this will remove the variable that was set before in `scripts/devon.properties`).

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. do the change above to your `conf/devon.properties` on windows
2. open CMD in that IDE
3. call `devon` command
4. do `echo %KEY%`

### Related/Dependent Issues

#394

### Affected version:

* `3.3.1`
* OS: Windows

hohwille:
Done with PR #430

PR #426, opened 2020-06-18
Fixed variable double qoutes issue #395

Rohit90700:
Fixed issue - https://github.com/devonfw/ide/issues/395

### DIFF
--- a/scripts/src/main/resources/scripts/environment-project.bat
+++ b/scripts/src/main/resources/scripts/environment-project.bat
@@ -79,10 +79,11 @@ rem %~1: variable name
 rem %~2: variable value (may contain ~ or ${var})
 :set_variable
 setlocal EnableDelayedExpansion
-set "value=%~2"
+set "value=%~2 %~3"
 rem replace ${var} variable syntax with windows %var% syntax
-set "value=!value:${=%%!
-set "value=!value:}=%%!
+set value=!value:${=%%!
+set value=!value:}=%%!
+set value=!value:"=%!
 rem resolve ~ to user home (USERPROFILE)
 if "!value:~0,1!" == "~" (
   set "value=%USERPROFILE%!value:~1!"


scripts/src/main/resources/scripts/environment-project.bat
```diff
@@ -79,10 +79,11 @@ rem %~1: variable name
...
 :set_variable
 setlocal EnableDelayedExpansion
-set "value=%~2"
+set "value=%~2 %~3"
 rem replace ${var} variable syntax with windows %var% syntax
-set "value=!value:${=%%!
-set "value=!value:}=%%!
+set value=!value:${=%%!
+set value=!value:}=%%!
+set value=!value:"=%!
```
hohwille, line 86: great. So this line replaces double quotes with nothing (empty string) if I get it right.

scripts/src/main/resources/scripts/environment-project.bat
```diff
@@ -79,10 +79,11 @@ rem %~1: variable name
 rem %~2: variable value (may contain ~ or ${var})
 :set_variable
 setlocal EnableDelayedExpansion
-set "value=%~2"
+set "value=%~2 %~3"
```
hohwille, line 82: Ah, if I get this correctly you were stumbling over situations like:
```
property="word1 word2"
```
And here `%~2` would be `"word1"` and `%~3` would be `word2"`, correct?

If that is the case we might need to do a `shift` command and then use the syntax to insert all parameters so it would also work for more words:
```
property="word1 word2 word3 word4 word5 word6 word7 word8 word9 word10 word11 word12"
```

WDYT?

Rohit90700: Hi @hohwille , You are right. For more then two words separated by spaces need to add more numbers of formal parameter like %~4 ,%~5..... etc.This is the fix only for a double quotes String like property="word1 word2" separated by space.

scripts/src/main/resources/scripts/environment-project.bat
Rohit90700: Yes

scripts/src/main/resources/scripts/environment-project.bat
hohwille: The MSDos/CMD Syntax for all parameters is `%*`. This will include also the first one.
To prevent including the first one (the property name), use the `shift` command that shifts all arguments to the left. The first parameter is lost and becomse the second that becomes the third, etc.

hohwille: The `shift` command in MS CMD is entirely broken and does not affect `%*`. As always with CMD simple things are pain. I am still working on a solution and will redo this PR.

hohwille: I tried something like this:
https://stackoverflow.com/questions/935609/batch-parameters-everything-after-1
However, the shift command seems as a side effect to manipulate the arguments of the main program while we are in a sub-routine that has its own arguments for the variable assignment. CMD is nothing but ugly hacking...

hohwille:
@Rohit90700 thanks for your work and hits how to fix #395. As there was no feedback for a longer while, I took over and created PR #430 as a replacement for this PR.

Issue #416, opened 2020-05-18
Meta information about all available devonfw instances in a system


suprishi:
As a developer of [Devon Dashboard](https://github.com/devonfw/devonfw-dashboard), I would like to have a feature where I can list all available devon instances in a system so that I can display the workspaces and the projects they hold.
Could it be possible to have some kind of meta file containing this information in a common directory (like the **username** directory available in windows, mac and linux systems). The file should be updated every time a devon instance is installed or uninstalled.

hohwille:
Should definitively be in `$DEVON_HOME/.devon/` (`.devon` folder in users home directory).
For the design make a suggestion. IMHO I see two obvious options:
1. a subfolder `installations` with a file for each installation containing the absolute path
2. a single file containing the absolute path of each project one on each line

IMHO the 2. option makes more sense to me.

This config should IMHO be maintained during setup in `ide` commandlet:
https://github.com/devonfw/ide/blob/2a096e33af58b75427c992a363f6b40728385288/scripts/src/main/resources/scripts/command/eclipse#L24

Will you try to provide a PR?

suprishi:
Based on what I could understand, the solution if implemented could work for future versions of Devon. We can have a single file, could be XML or JSON (preferably JSON, as it will be easier to read in a JavaScript environment), which is updated each time the user sets up a new Devon instance and also when projects and workspaces are created using the IDE. 
Currently, do we have any way of querying the instances/workspaces/projects already setup in a system so that we can get a list of earlier versions as well?

hohwille:
> Currently, do we have any way of querying the instances/workspaces/projects already setup in a system so that we can get a list of earlier versions as well?

Nope. There is one instance that can be found in `~/.devon/home` but other instances could only be found by scanning the disc (optimized with some heuristics, e.g. assuming that all IDEs are installed in the same toplevel folder or share the same parent folder, this could be fast, but otherwise it will be pain).

So ideally some dashboard driver could start working on this story so we can push this new feature our sooner and have it spread out in the wild for future dashboards.

PR #428, opened 2020-07-01
added script to create a meta file in the users directory after setup

suprishi:
In this PR I've added a function in the ide script to create a meta file on setup. Please let me know if it will work on Windows system as well or do I need to call a batch file for Windows specifically. In that case I will update this PR

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -3,6 +3,18 @@
 source "$(dirname "${0}")"/../functions
 DEVON_IDE_REPO_URL="https://repo.maven.apache.org/maven2/com/devonfw/tools/ide"
 
+function doUpdateMetaFile() {
+  if [ -f "${USERPROFILE}/devon.txt" ]
+  then
+    if ! grep -q "${DEVON_IDE_HOME}" "${USERPROFILE}/devon.txt"
+	then
+      echo "${DEVON_IDE_HOME};" >> "${USERPROFILE}/devon.txt"
+	fi
+  else
+    echo "${DEVON_IDE_HOME};" > "${USERPROFILE}/devon.txt"
+  fi
+}
+
 function doUpdateScripts() {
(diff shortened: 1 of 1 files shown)


scripts/src/main/resources/scripts/command/ide
```diff
@@ -3,6 +3,18 @@
...
+  if [ -f "${USERPROFILE}/devon.txt" ]
+  then
+    if ! grep -q "${DEVON_IDE_HOME}" "${USERPROFILE}/devon.txt"
+	then
+      echo "${DEVON_IDE_HOME};" >> "${USERPROFILE}/devon.txt"
+	fi
+  else
+    echo "${DEVON_IDE_HOME};" > "${USERPROFILE}/devon.txt"
+  fi
+}
```
hohwille, line 16: Excellent solution. However, suggestion for improvement:
* Please follow our guidelines and link commits and PRs with the according issues: https://github.com/devonfw/.github/blob/master/CONTRIBUTING.asciidoc#committing-changes - The issue this PR is implementing seems to be #416
* Also you should always create a fork to work on. On that create feature branches and from that create your PR. I have no clue why you have direct permission to commit on this repo, but I will try to restrict those for all users to prevent accidents in the future. See https://github.com/devonfw/.github/blob/master/CONTRIBUTING.asciidoc#forking
* We already have a `.devon` folder that we create in the users home directory. Hence we should write the medata data into that directory as already described in #416
* The filename `devon.txt` says nothing about its content. Instead it should be called `installations`, `ide-paths`, or something like this.
* Why do you terminate each line with a semicolon? I think this is a bad idea. It has to be removed when parsed and in odd cases a path may for whatever reason contain a semicolon.
* To improve, I would suggest using a local variable for the metafile to avoid repeating it four times redundantly.

suprishi:
Closing this PR as I raised [another one](https://github.com/devonfw/ide/pull/429) from my fork.

PR #429, opened 2020-07-02
Added script to create a meta file in the users directory after setup

suprishi:
I've updated the PR according to the suggestions provided for rework.

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -4,14 +4,15 @@ source "$(dirname "${0}")"/../functions
 DEVON_IDE_REPO_URL="https://repo.maven.apache.org/maven2/com/devonfw/tools/ide"
 
 function doUpdateMetaFile() {
-  if [ -f "${USERPROFILE}/devon.txt" ]
+  local ide_paths="${USERPROFILE}/.devon/ide-paths"
+  if [ -f "${ide_paths}" ]
   then
-    if ! grep -q "${DEVON_IDE_HOME}" "${USERPROFILE}/devon.txt"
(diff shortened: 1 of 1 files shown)


scripts/src/main/resources/scripts/command/ide
```diff
@@ -3,6 +3,19 @@
 source "$(dirname "${0}")"/../functions
 DEVON_IDE_REPO_URL="https://repo.maven.apache.org/maven2/com/devonfw/tools/ide"
 
+function doUpdateMetaFile() {
+  local ide_paths="${USERPROFILE}/.devon/ide-paths"
```
hohwille, line 7: In order to work also properly when `setup` has been invoked from cygwin, you need to use `${DEVON_HOME_DIR}` instead of `${USERPROFILE}`.
See here for reason (as cygwin for historic reasons has its own home directory and then cygwin would write/append a different file than CMD):
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/environment-project#L70

suprishi: So shall I simply replace `${USERPROFILE}` with `${DEVON_HOME_DIR}` or do a conditional check for `${OSTYPE}` and use `${DEVON_HOME_DIR}` if it is `"cygwin"`?

hohwille: First option: Just use `${DEVON_HOME_DIR}` instead as this is set already be the IDE and we can avoide code duplication.

suprishi: Done.

Though I want to clarify when I tested this changed function it did not give expected results on my system. Probably there are other factors which can't be replicated while testing?

Issue #433, opened 2020-07-23
Windows 10: devon command line sets wrong environment variables (with tilde symbol)


yoktobit:
### Expected behavior

Running the devon command line by calling it from the windows explorer context menu or by calling "devon" in any other command line, I would expect devon to set the environment variables for M2_REPO, MAVEN_HOME, MAVEN_OPTS and JAVA_HOME.

### Actual behavior
The command sets the variables M2_REPO~, MAVEN_HOME~, MAVEN_OPTS~ and JAVA_HOME~
This leads to a lot of other errors like checking out the repo to a wrong folder or accessing wrong remote repos with wrong credentials.

### Steps to reproduce (bug)

1. update to latest devon
2. open command line in some workspace
3. run devon command
4. run set
5. see how environment variables are set wrong

### Related/Dependent Issues
#395 
#430 

### Comments/Hints:
By undoing #430 I could solve it temporarily (I have no whitespaces in my project folders).

### Affected version:
2020.04.003

* OS: Windows 10

markusschuh:
Good catch!
 #430 includes a typo in [line 91 of `environment-project.bat`](https://github.com/devonfw/ide/blob/ea9b1fcf3987eba68eb10488e901cf6969a5aa55/scripts/src/main/resources/scripts/environment-project.bat#L91) 

The line should be

`set "replacement=%~2%="`

which would remove any surrounding quotes. But the current script line

`set "replacement=%2~%="`

adds a  "~" instead, what is the issue.

N,B.:  This issue is a good example, why my enhancement proposal #261 for a simple DEVONFW_IDE_TRACE hook should be integrated into the scripts of devonfw-ide. 

With that in place (I had to re-add it into my local - just updated - devonfw-ide again) I could identify the root cause of this issue by a few simple enough steps:

1. `set DEVONFW_IDE_TRACE=yes`
1. execute `devon`
1. locate the first appearance of `M2_REPO~` in the trace output ( => line 91)
1. understand the code - and such identify the error
1. `set DEVONFW_IDE_TRACE=`

hohwille:
@yoktobit thanks for reporting this issue.
@markusschuh thanks for your investigation.
I am quite confused how this severe bug was neither detected by me nor be the two core-team members that claimed to have tested PR #430 
However, murphies law is with us of course and Windows CMD is one of the most confusing and error-prone lang so lets look ahead and get this fixed asap.

hohwille:
[...]

PR #434, opened 2020-07-27
#433: fix for variable resolution on windows (prevent title symbol in variable name when export is used)

hohwille:
Fix for #433 as proposed by @markusschuh


hohwille:
I was able to reproduce the error of #433 and was now able to confirm this error is fixed with the change of this PR.

PR #430, opened 2020-07-07
#395: windows fix for variables containing whitespaces or quotes

hohwille:
Fix for #395 and replacement for PR #426

PR #435, opened 2020-07-27
#435: fix variable solution on bash

hohwille:
issues with quotes and whitespaces in variable values also on bash (linux/mac/win)
similar to #434 and #430


scripts/src/main/resources/scripts/environment-project
```diff
@@ -23,6 +23,7 @@ function doLoadProperties() {
     then
       key="${key//$'\r'}"
       value="${value//$'\r'}"
+      echo "$key=$value"
```
hohwille: ups. Echo statements have to be removed. Thats why we create PRs and review (even if I review myself ;)).

Issue #427, opened 2020-06-26
release: keep leading zeros


hohwille:
As a user of devonfw-ide release feature, I want it to keep the syntax of the versioning schema so that I do not have to type the next version in case I only need to increas the minor version:

Current behaviour:
```
Trying to determine current version of your project...
Current version: 2020.04.002-SNAPSHOT
Release version: 2020.04.002
Next version: 2020.04.3-SNAPSHOT

Is next version correct?
(yes/no): no
No...
Please enter next version: 2020.04.003-SNAPSHOT

Current version: 2020.04.002-SNAPSHOT
Release version: 2020.04.002
Next version: 2020.04.003-SNAPSHOT

Is next version correct?
(yes/no): yes
```

Expected behaviour:
```
Trying to determine current version of your project...
Current version: 2020.04.002-SNAPSHOT
Release version: 2020.04.002
Next version: 2020.04.003-SNAPSHOT

Is next version correct?
(yes/no): yes
```

PR #436, opened 2020-07-30
#427: keep leading zeros

hohwille:
Fix for #427


hohwille:
As there are tests proving the solution works and it is a little change only, I wont beg and wait for reviews...

PR #438, opened 2020-07-31
Add DEVON_IDE_TRACE to scripts

markusschuh:
for #261

Variable renamed to start with DEVON_ (not DEVONFW_) in order to not introduce another namespace prefix.

The trigger line could also be moved into `functions` instead  (for most scripts) - but then the call of `functions` itself would not be traced - which makes it harder to read the generated trace logs.

### DIFF
--- a/documentation/variables.asciidoc
+++ b/documentation/variables.asciidoc
@@ -32,4 +32,5 @@ Please note that we are trying to minimize any potential side-effect from `devon
 |`«TOOL»_VERSION`|`-`|The version of the tool `«TOOL»` to install and use (e.g. `ECLIPSE_VERSION` or `MAVEN_VERSION`).
 |`«TOOL»_BUILD_OPTS`|e.g.`clean install`|The arguments provided to the build-tool `«TOOL»` in order to run a build.
 |`«TOOL»_RELEASE_OPTS`|e.g.`clean deploy -Dchangelist= -Pdeploy`|The arguments provided to the build-tool `«TOOL»` in order to perform a release build.
+|`DEVON_IDE_TRACE`||If value is not an empty string, the `devonfw-ide` scripts will trace each script line executed. For bash two lines output: before and again after expansion.
 |=======================
--- a/scripts/src/main/resources/scripts/command/build
+++ b/scripts/src/main/resources/scripts/command/build
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
 
--- a/scripts/src/main/resources/scripts/command/cicdgen
+++ b/scripts/src/main/resources/scripts/command/cicdgen
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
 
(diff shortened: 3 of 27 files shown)


scripts/src/main/resources/scripts/command/build
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
```
hohwille, line 2: as this line is already included in `functions` that is sourced one line after, why do we need to add it redundantly to all scripts sourcing `functions`?
Will not really hurt but seems pointless clutter to me.

documentation/variables.asciidoc
```diff
@@ -32,4 +32,5 @@ Please note that we are trying to minimize any potential side-effect from `devon
 |`«TOOL»_VERSION`|`-`|The version of the tool `«TOOL»` to install and use (e.g. `ECLIPSE_VERSION` or `MAVEN_VERSION`).
 |`«TOOL»_BUILD_OPTS`|e.g.`clean install`|The arguments provided to the build-tool `«TOOL»` in order to run a build.
 |`«TOOL»_RELEASE_OPTS`|e.g.`clean deploy -Dchangelist= -Pdeploy`|The arguments provided to the build-tool `«TOOL»` in order to perform a release build.
+|`DEVON_IDE_TRACE`||If value is not an empty string, the `devonfw-ide` scripts will trace each script line executed. For bash two lines output: before and again after expansion.
```
hohwille, line 35: Excellent, that you also added the documentation.
However, the way you implemented the feature this is not a regular variable like other variables in `devonfw-ide` as it can only be set upfront via `export DEVON_IDE_TRACE=true` while setting it in `devon.properties` as `DEVON_IDE_TRACE=true` will IMHO not work.
If you see this now and can quickly provide an update, that would be highly appreciated. Otherwise I might need to redo this PR to take over and add small changes.

scripts/src/main/resources/scripts/command/build
markusschuh: As already mentioned in my PR comment. I considered the alternative to only start tracing within _function_ -  but decided against it, since this makes reading the trace log out more difficult that it already is. Imho it should trace all executed lines - and not skip some important context switches just because some may consider the related code as pointless clutter.

Imagine a stacktrace, where you skip some function call - makes not much sense to me on TRACE level.

There could even be errors in the call of _functions_ itself. Or imagine the _functions_ file were within a new commandlet called - not sourced. This would stop tracing at all on return if the trace flags have not been activated in the original calling shell.

Sure - the PR is a poor mans tracing. But it is a quick win - even with the complete solution, which adds the trace on detection at each script begin.

documentation/variables.asciidoc
markusschuh: You are right - I just focussed on the table header. Using it with `export` inside `devon.properties` - but of course, this starts tracing only after variable import - which might be already too late.

I still think, that the variables documentation is a good place to mention this variable - but another new documentation file that describes the  tracing would of course also help. I'll provide an update soon.

scripts/src/main/resources/scripts/command/build
hohwille: OK, I understood. I was not aware that this actually makes a difference in the resulting trace log.
So I just close this as resolved. Thanks for clarification.

documentation/variables.asciidoc
hohwille: We are in the process of releasing devonfw 2020.08 roughly next week. If you can not provide the update soon enough, I will take over. Maybe I will just merge and then quickly edit the docs immediately after via browser to keep it simple.

markusschuh: please do so

Issue #432, opened 2020-07-15
vsCode settings are not updated


maybeec:
### Expected behavior

As a developer, I want to share vsCode settings so that everyone in my team works for example with the same linting settings.

### Actual behavior

I updated the settings in `settings/vscode/workspace/update/.vscode/settings.json` and ran `devon ide update all`, but the settings are not updated. Even on a fresh installation, the settings are not propagated to the user settings.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Create a settings file in `settings/vscode/workspace/update/.vscode/settings.json` of your current IDE installation with the content `{ "tslint.autoFixOnSave": true }`.
2. Run `devon ide update all`
3. The user settings file in `conf/vscode/User/settings.json` will not contain the new or update value.

### Affected version:

* OS: Windows

hohwille:
@maybeec thanks for reporting the issue.
Can you check some things:
* Do you have `vscode` configured in `DEVON_IDE_TOOLS`?
* Is the configurator (the Java programm that updates/merges the settings) actually called by `devon ide update all`?

maybeec:
Hi @hohwille, we configured `DEVON_IDE_TOOLS=(java mvn eclipse vscode node npm ng)`

I am not sure whether this progam is running. At least I cannot see anything on the logs:
```
*** Software Update of devonfw-ide-scripts ***
Updating /c/project from https://repo.maven.apache.org/maven2/com/devonfw/tools/ide/devonfw-ide-scripts
Trying to determine the latest available version from https://repo.maven.apache.org/maven2/com/devonfw/tools/ide/devonfw-ide-scripts/maven-metadata.xml
The devonfw-ide-scripts package is already at the correct version 2020.04.004
Already up to date.

*** Setting up java ***
Version 1.8.0_181 of java is already installed.
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.181-b13, mixed mode)

*** Setting up mvn ***
Version 3.6.2 of maven is already installed.
Apache Maven 3.6.2 (40f52333136460af0dc0d7232c0dc0bcf0d9e117; 2019-08-27T17:06:16+02:00)
Maven home: C:\project\software\maven
Java version: 1.8.0_181, vendor: Oracle Corporation, runtime: C:\project\software\java\jre
Default locale: en_US, platform encoding: Cp1252
[...]

Issue #384, opened 2020-04-02
VSCode auto-installation of plugins not working


hohwille:
### Expected behavior

When I setup devonfw-ide it should install extensions configured in `settings/plugins/*.properties` into VSCode automatically. Using default settings this should install `devonfw-extension-pack`.

### Actual behavior

No extension is installed automatically. As it seems the `vscode` script triggers the command to do so but it seems to have no effect:
```
******** ATTENTION ********
Installing plugin (devonfw.devonfw-extension-pack) into VS code...
Please wait until VSCode has successfully installed all plugins.
You may then use or close VSCode as you like.
(node:20821) Electron: Loading non context-aware native modules in the renderer process is deprecated and will stop working at some point in the future, please see https://github.com/electron/electron/issues/18397 for more information
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. build the latest release of `devonfw-ide` from `develop`
2. extract it to new folder and run `setup` with defaults (e.g. provide `-` as arg or hit return for settings).

### Related/Dependent Issues
* #380 

### Comments/Hints:
Currently clueless.

### Affected version:

* OS: Only tested on Mac so far

hohwille:
Further `devon vscode setup` will trigger the installation of the configured extensions every time no matter if they are already installed. Is that good because it is KISS and vscode should handle this idem-potent?

hohwille:
I ran the command manually:
```
/Users/hohwille/projects/ide/software/vscode/bin/code --user-data-dir '/Users/hohwille/projects/ide/conf/vscode' --extensions-dir '/Users/hohwille/projects/ide/software/vscode-extensions' --force --install-extension devonfw.devonfw-extension-pack
```
Same effect. It more or less seems to do nothing. I also omitted `--force` and tried other things but somehow there is a bug in VSCode itself and not in `devonfw-ide`.
I need help here as I can not fix the issue.

hohwille:
To me it seems as if there is a bug in VSCode and `--extensions-dir` CLI option does not work together with `--install-extension`.

PR #439, opened 2020-08-02
fix vscode extension installation + introduced extension cleanup

maybeec:
fixing #384 (installation of vscode extensions) and introduced the new ability to cleanup extensions.
The latter is needed to allow properly replacing extensions, which could be conflicting at some time.
One example for this is the replacement of tslint by eslint with some more impacts on autoimporters and similar extensions.
I tried to run `--uninstall-extension` parameter to uninstall specific plugins, but unfortunately the command was not able to terminate successfully in case the intended extension was not available in the current vscode instance. So finally, I wrote a way to cleanup the extension folder automatically by the update process (optionally) to allow a proper plugin replacement in case it is necessary.

This PR also relates to #432 with respect to the error log of failing extension installations, but it does not fix the settings update.

### DIFF
--- a/scripts/src/main/resources/scripts/command/vscode
+++ b/scripts/src/main/resources/scripts/command/vscode
@@ -51,14 +51,12 @@ function cleanupPlugins() {
 function doAddPlugins() {
   local file
   local pluginsToInstall
-  local pluginsToUninstall
   for file in "${SETTINGS_PATH}"/vscode/plugins/*.properties
   do
     if [ -f "${file}" ]
     then
       plugin_id=""
       plugin_active="true"
-	  force_uninstall="false"
       doLoadProperties "${file}"
       if [ -z "${plugin_id}" ]
       then
@@ -74,7 +72,7 @@ function doAddPlugins() {
 }
 
 function doInstallPlugins() {
-  if [ ! -z "${1}" ]
+  if [ -n "${1}" ]
   then
 	doEchoAttention "Execute plugin installation with following parameters (${1}) into VS code...\nPlease wait until VSCode has successfully installed all plugins.\nYou may then use or close VSCode as you like."
     doRunVsCode --force "${1}"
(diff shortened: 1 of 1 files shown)


maybeec:
Basically the execution of all extension installations in one command was fixing the problem of the file lock, which was caused by the issue that all instances ran on the same log file and the installation command was terminated before the lock was dropped. I could not find any other reliably working solution. 

Regarding the remove feature, I also searched for reliable solutions which would work in every case of a project state. I finally even found this solution to be the most reliable in the web as the uninstall does not have a force and fails in case the extension to be removed is not installed....

So finally that's the outcome :D

hohwille:
@maybeec Excellent and thanks for your feedback.
What I would love to have is that we also get some documentation about all this for `vscode` as we already have for `eclipse`:
https://github.com/devonfw/ide/blob/master/documentation/eclipse.asciidoc#plugins

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -47,15 +63,22 @@ function doAddPlugins() {
...
+  doInstallPlugins "${pluginsToInstall}"
 }
 
-function doAddPlugin() {
-  doEchoAttention "Installing plugin (${1}) into VS code...\nPlease wait until VSCode has successfully installed all plugins.\nYou may then use or close VSCode as you like."
-  doRunVsCode --force --install-extension "${1}" &
+function doInstallPlugins() {
+  if [ -n "${1}" ]
+  then
+	doEchoAttention "Execute plugin installation with following parameters (${1}) into VS code...\nPlease wait until VSCode has successfully installed all plugins.\nYou may then use or close VSCode as you like."
```
hohwille, line 77: was a tab, I am fixing it locally and add the missing documentation.

hohwille:
Actually revisiting this, I have one concern:
The new property `clean_plugins_on_update` is configured per plugin/extension. However, this does not make any sense.
Instead this should rather be kind of a global configuration variable for a project defined only once per settings.
Assuming you have two plugins for VS Code configured:
* a.properties with `clean_plugins_on_update=false`
* b.properties with ``clean_plugins_on_update=true`
What should happen?
This IMHO does not make any sense.
The option should rather go to `devon.properties`.
Do you agree? Or am I missing something?

hohwille:
Replaced by PR #443

Issue #431, opened 2020-07-07
devon build command not working correct for yarn or npm


hohwille:
As discussed in PR #417 the command `devon build` for yarn or in other words the command `devon yarn build` is not working correct.
It invokes `yarn build`. However, dependencies may not be installed and not every project may have a `build` script.
IMHO the same applies for `npm` as well.

The JavaScript ecosystem is less standardized so there are no clear goals and guidelines for this (other than for `maven` or `gradle`).
To me the build of a project should behave a following:
*  install dependencies (`yarn install` or `npm install`)
* run build script if present (`yarn run build` or `npm run build` if `build` section is present in `scripts` section of `package.json`)

hohwille:
So to figure out if the `package.json` file actually contains a `build` script, I came up with this (ugly) bash pipe:
```
cat package.json | tr '\n' '§' | tr -d '\r' | grep -o "[\"']scripts[\"']\s*:\s*{\s*§.*[\"']build[\"']\s*:"
```
As we need support on all OS, we can not expect `grep` to support options like `-z` or `-P`.
I know we should avoid such ugly pipes as discussed in #358 but in this case:
* it is only called in this very specific situation (so performance is not really an issue)
* I do not see any alternative (except assuming that there is a `build` script included in `package.json` but that may not always be the case).

PR #448, opened 2020-08-24
#431: fixed build command

hohwille:
Fix for #431

### DIFF
--- a/scripts/src/main/resources/scripts/command/npm
+++ b/scripts/src/main/resources/scripts/command/npm
@@ -49,17 +49,12 @@ function doCheckTopLevelProject() {
 
 function doRunBuild() {
   doSetup
-  local extra_arg=""
-  if [ ! -d node_modules ] && [[ "${*}" != *install* ]]
-  then
-    extra_arg="install"
-  fi
-  doEcho "Running: npm ${extra_arg} ${*}"
+  doEcho "Running: npm ${*}"
   if doIsQuiet
   then
-    npm --quiet ${extra_arg} "${@}"
+    npm --quiet "${@}"
   else
-    npm ${extra_arg} "${@}"
+    npm "${@}"
   fi
 }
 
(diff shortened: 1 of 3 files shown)


hohwille:
As discussed offline there have been issues when `node_modules` have not been present. This was previously silently adding `install` as goal to the build-tool. However, unlike tools like `mvn` for `npm` or `yarn` this was entirely incorrect as
```
npm install build
```
means something totally different than
```
npm install
npm build
```
In JS build tools goals can also be followed by parameters such as
```
npm add «pacakge»
```
Therefore `npm install build` is thinking that the dependency `build` should be installed and even adds this to the `package.json` as dependency leading to silly results. It took me while to understand these worlds of JS. Would have been easier if some domain expert had created the commandlets initially.

hohwille:
OK, for some odd reason `tr '\n' '§'` is not working in git-bash (on windows).
Therefore we had to use `sed` magic. To avoid redundancy of this kind of tricky detection inside `package.json`, I moved it to `functions` and share it between `npm` and `yarn` commandlets.

Also I found and fixed another small bug that debug option (`-d` or `--debug`) was not passed by `doDevonCommand`. Therefore without this fix `devon build -d` was not printing debug info while `devon yarn -d build` did. With the fix both behave the same.

markusschuh:
Introducing a non-ASCII char like '§' into the stream has an influence on the "binary detection" logic built into `grep` - and this is dependent on the shell environment, since the intention of this check is to avoid output of binary as potentially non-printable chars inside `grep` output. So an additional "-a" option (treat all files as ASCII text) to `grep` would harden the code more against such environment differences.

But since both `npm` and `yarn` have `node` as prerequisite - hence a more powerful scripting language is available - it might make more sense to do the check with javascript:

```
node -e "process.exit(require('./package.json').scripts.hasOwnProperty('build'));"
```

Tested successfully also with an a bit older node 10.x

PR #455, opened 2020-08-31
Fix args quoting for node commandlet

markusschuh:
Fix quoting of args in array context.

Use "$@" instead of "$*" for correct quoting of each single command option. Otherwise "devon node" can't take more than one argument.

### DIFF
--- a/scripts/src/main/resources/scripts/command/node
+++ b/scripts/src/main/resources/scripts/command/node
@@ -45,9 +45,9 @@ function doRun() {
   doEcho "Running: node ${*}"
   if doIsQuiet
   then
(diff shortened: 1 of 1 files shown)


hohwille (approved):
@markusschuh thanks for pointing this out and fixing it via PR. Indeed this is the right way to do it and in other commandlets (e.g. `mvn`) we also use `${@}` instead of `${*}`. However, I also have to admit that bash confuses me with these two variants and shellcheck wants us to use `${*}` in the `doEcho` statement (see above) since it is within double quotes.
Thanks for the fix 👍

Issue #452, opened 2020-08-25
Consider support for IntelliJ Ultimate


hohwille:
As a user of devonfw-ide, I want to be able to use the ultimate edition of IntelliJ so that I can make use of the payed features in my project.

However, currently only the community edition is supported.

PR #460, opened 2020-09-14
#452: IntelliJ add Ultimate Edition

jensschirmer:


jensschirmer:
@hohwille : Is this okay?

documentation/intellij.asciidoc
```diff
@@ -22,3 +22,14 @@ You may also supply additional arguments as `devon intellij «args»`. These are
...
 |`create-script` |create launch script for this IDE, your current workspace and your OS
 |=======================
+
+There are variables that can be used for IntelliJ. These are explained by the following table:
+
+.Variables of devonfw-ide for intelliJ
+[options="header"]
+|=======================
+|*Variable*|*Default-Value*|*Meaning*
+|*`INTELLIJ_VERSION`*|`2019.2`|The version of the tool IntelliJ to install and use.
```
hohwille, line 32: adding the default value for the version is dangerous as we will typically not be able to maintain that and forget to update this documentation. So I would rather remove this column or write something like «latest_tested_version» here.

documentation/intellij.asciidoc
```diff
@@ -22,3 +22,14 @@ You may also supply additional arguments as `devon intellij «args»`. These are
...
+|*`INTELLIJ_EDITION_TYPE`*|`C`|The edition of the tool IntelliJ to install and use. The value `C` mean Community edition and the value `U` mean Ultimate edition. The Ultimate edition requires a license. The user has to buy the license separately and it is not part of devon-ide. The devon-ide only supports download and installation.
```
hohwille, line 33: can you write `devonfw-ide` instead of `devon-ide` to be aligned with the officials. Thanks.

scripts/src/main/resources/scripts/command/intellij
```diff
@@ -9,6 +9,7 @@ function doSetup() {
   if [ -n "${1}" ] || [ ! -d "${IDEA_HOME}" ]
   then
     software_version="${INTELLIJ_VERSION:-2019.2}"
+	software_edition_type="${INTELLIJ_EDITION_TYPE:-C}"
```
hohwille, line 12: Can you check indendation? Using tabs?

hohwille (approved):
@jensschirmer Wow! Thanks for your great contribution. I mark as approved as your solution is already great 👍 I just have some nice to have improvements suggested in my review comments. Would you have some minutes to update your PR? Otherwise I would take over but I am also going on vacation next monday.

documentation/intellij.asciidoc
jensschirmer: I have corrected it with commit 3.

hohwille (approved):
great. Thanks for the quick rework. Going to merge. We still might need some additional PRs to make intelliJ support work suitable, but this is again a nice imporvement and step ahead.

PR #465, opened 2020-09-18
Security update for node.js 

sjimenez77:
- Node.js LTS 12.18.4 with security fixes. 
- Also includes the VS Code latest update 1.49.1.


sjimenez77:
- VS Code updated to 1.49.2.

@hohwille, I think that we can merge this as it does not affect anything else and I already removed the previous _ionic_ CLI from toolset.

hohwille (approved):
@sjimenez77 thanks for the update. Highly appreciated that you keep these versions up-to-date and secure. Thanks 👍

Issue #456, opened 2020-09-02
Property values are truncated


fdcg:
If a property value in a devon.properties file contains certain special characters (e.g. } or !} then the property value is truncated there and thus incorrectly transferred to the environment. This is at least the case with Windows 10.

hohwille:
@fdcg thanks for reporting this.
Is there a real use-case behind this or is this just a theoretical observation?
If there is a real use-case please give an example or explanation so we can prioritize this issue.

fdcg:
Yes, there is a real use case.
Each member of my team has their own account on a central infrastructure for a Maven repo, etc.
I want to create centralized settings for the team that use fixed environment variables where credentials are required.
These environment variables must be set by the developer in the private devon.properties.
For example, if the password contains the special characters mentioned above, it will be truncated.

markusschuh:
Indeed the current `environment-project.bat` replaces `}` with `%%` even in cases, where this is not needed. I'll provide a PR to fix this.

Beyond this detail - which is easy to fix - your use case IMHO raises some more open questions. This is somehow also related to #415 :

Just from a practical point of view use of clear-text passwords within config files is always quite a challenge, since some characters - valid for passwords - will anyway need some quoting in common config files. Imagine the following were - even when quite unlikely - a user's password. It will be hard to enable the user, to define this within a properties file - even in unix and outside of `devonfw-ide`:

```
~${PATH}"''"
```

Also from view point of security sensitive details like a user password <b>should be protected by additional controls</b>.

In practice this is of course more simpler to state than to realise. So what is possible with `devonfw-ide` out of the box - as far as I can say?

If the password were only used for access with maven to a repo, protected by basic authentication, the [Maven --encrypt-password](https://maven.apache.org/guides/mini/guide-encryption.html) method can and should be used within `devonfw-ide`.  The needed master password is seeded automatically during `setup` of `devonfw-ide` and the per default installed `${DEVON_IDE_HOME}/conf/.m2/repository` contains [helpful guidance](https://github.com/devonfw/ide-settings/blob/da2e7c825828748c3d3bd14e403d9a995b644b27/devon/conf/.m2/settings.xml#L7), how to  protect the password such against the simpler spy out methods.

[...]

PR #487, opened 2020-10-17
#456: add condition around variable expansion

markusschuh:
The closing curly brace should only be replaced, when a variable expansion is necessary.


hohwille (approved):
@markusschuh Looks good and functionally correct to me. Thanks for this PR 👍 
Did you check performance (that is already a little slow on windows)?
Assuming no `${...}` will be found will this be as fast or even faster as before?

hohwille:
I see this as ready to be merged but give some time for @markusschuh in case he wants to anwer to my questions from review.

hohwille:
Regarding `findstr` I found some comments about performance here:
https://stackoverflow.com/questions/34058575/improve-windowss-batch-script-performance
So it seems not very fast and is executed for every line and always has to parse and apply the regex.

mperezor19:
Hi,  I tested executing 'devon' command in  this feature branch and master branch and the executions time have significant difference:

Feature branch:
measure command 1
measure command 2
measure command 3
measure command 4
measure command 5


Master branch:
measure-master-5
measure-master-1
measure-master-2
measure-master-3
measure-master-4

Also mention that there is a test in this branch which is causing error when build, the regex should be modified like this:
test-fix

hohwille:
Thanks  @mperezor19 for your tests and analysis.
Do I read your results properly such that the result is the execution of running plain `devon` command (parsing properties and setting env variables) takes more than twice as much execution time with this PR applied?
Then my gut feeling was correct and we should not merge without further action/decision making.

Your second point IMHO does not seem to be related to this PR at all. Anyway also thanks for that feedback. So you are talking about this:
https://github.com/devonfw/ide/blob/4328c88f087d671889b2273f34ee01fa537a7cc1/configurator/src/test/java/com/devonfw/tools/ide/configurator/SortedPropertiesTest.java#L27

If I am not mistaken, the problem relates to different EOL encoding and only happens on windows. As I am using MacOS and our travis CI build is using Linux both are not affected. Your fix will only work on windows. So we should add an optional carriage return here so it works on all plattforms. I will file a new issue for this.

markusschuh:
Let's just close this PR. 
I don't share the interpretation that the above time series would show a slow down of "2 times". But sure - forking external commands is insane for performance in any case. Which disqualifies `CMD `- and by the way `bash` as well, even when less - for too complex tasks..

Issue #494, opened 2020-11-26
SortedPropertiesTest fails on windows


hohwille:
As identified in PR #487 the `SortedPropertiesTest` does not seem to work on windows.

Problem is this line:
https://github.com/devonfw/ide/blob/4328c88f087d671889b2273f34ee01fa537a7cc1/configurator/src/test/java/com/devonfw/tools/ide/configurator/SortedPropertiesTest.java#L27

We need to consider that windows EOL style also contains a carriage return that we should add as optional to this regex.

hohwille:
So `.*` does not match `\r`.
Do we need `"#.*(\r)?\n"` then to make it work on all plattforms?
As an alternative we could retrieve the OS specific EOL sequence from Java and append it to the regex (but IMHO only if the above regex does not work).

Gabyk0:
End-of-line code is:
- `\n` on Unix/Unix-like platform;
- `\r` on old Mac platform (pre-Mac OS X);
- `\r\n` on Windows platform.

`#.*\r?\n` (or `#.*(\r)?\n`) will work on Unix and Windows, to cover all three cases, `(\r\n|\r|\n)` should do the trick.

So, the following line will solve the issue:
 ` String result = buffer.toString().replaceAll("#.*(\r\n|\r|\n)", ""); // remove invariant date header `

hohwille:
Done with PR #496

Issue #471, opened 2020-09-25
Preconfigure Project Explorer with Hierarchical Project Presentation


hohwille:
A cool feature of the `Project Explorer` view in Eclipse is the `Hierarchical` project presentation that is unfortunately not enabled by default. It can be manually enabled like this:
![grafik](https://user-images.githubusercontent.com/1923119/94295720-f412ba00-ff61-11ea-9f21-d1b0e5707c31.png)

In our settings in setup we should preconfigure this what will enable it automatically when the workspace is created but will leave the user with the freedom to change as desired.

hohwille:
Needs to go somewhere here:
https://github.com/devonfw/ide-settings/tree/master/eclipse/workspace/setup/.metadata/.plugins

hohwille:
Last thing to do is to update the submodule for the `settings` and then we can close this issue.

TheTisiboth:
In one of my eclipse instance, I have this menu option, and in one other eclipse instance (in an other workspace), I don't see the Projects presentation option. How can I have access to it?

PR #496, opened 2020-12-04
Completion of release 2020.12.001

hohwille:
* fix for #471 
* fix for #494 
* added 2020.12.001 to changelog


hohwille:
The Travis build will fail due to #497. As the local build works fine, I will merge this anyhow and we can do additional QA and testing before we publish the release officially.

Issue #505, opened 2021-01-03
devon java create - ERROR: Missing arguments for doMavenArchetype


maybeec:
### Expected behavior

As a user, I want to create a new project so that it automatically is created in the default wokspace.

### Actual behavior

```
PS D:\tmp\devonfw-ide-test> devon java create
WARNING: You are currently not directly in a workspace folder. Current location is:
/d/tmp/devonfw-ide-test
Create project in main workspace instead?
(yes/no): yes

******** ATTENTION ********
ERROR: Missing arguments for doMavenArchetype
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Install clean devonfw-ide
2. run `devon java create` in root dir of the installed IDE
3. Answer `Create project in main workspace instead?` with `yes`

### Comments/Hints:

Same is happening if you run `devon java create` within a workspace:

```
******** ATTENTION ********
ERROR: Missing arguments for doMavenArchetype
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
```

### Affected version:

* OS: Windows
* Console: PowerShell / Git Bash / CMD

maybeec:
ok, this is invalid as you could simply provide the project name as parameter to make it work, but I feel, this can be improved to prevent from stopping users to proceed here as they might think they faced a bug.

PR #506, opened 2021-01-03
#505 adapted error message to guide the user properly

maybeec:
proposal to improvement for #505

### DIFF
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -545,7 +545,7 @@ function doArrayContainsItemWithPrefix() {
 function doMavenArchetype() {
   if [ -z "${1}" ]
   then
-    doFail "Missing arguments for doMavenArchetype ${*}"
+    doFail "Missing arguments for doMavenArchetype ${*}. Please find the documentation and examples of the parameters at https://devonfw.com/website/pages/docs/devonfw-ide-usage.asciidoc.html#java.asciidoc"
   fi
   local old_args
   old_args="${*}"


scripts/src/main/resources/scripts/functions
```diff
@@ -545,7 +545,7 @@ function doArrayContainsItemWithPrefix() {
 function doMavenArchetype() {
   if [ -z "${1}" ]
   then
-    doFail "Missing arguments for doMavenArchetype ${*}"
+    doFail "Missing arguments for doMavenArchetype ${*}. Please find the documentation and examples of the parameters at https://devonfw.com/website/pages/docs/devonfw-ide-usage.asciidoc.html#java.asciidoc"
```
hohwille, line 548: `doMavenArchetype` is used for different purposes. If called from a commandlet other than `java` this change does not seem to make sense to me. Furhter, we should never link directly to our website. Sorry to say so but we have perfectly proven of the last years that our website URLs are entirely unstable. Therefore if I link to documentation, I have learned to always link to github that has a stable api as long as we do not resturcture our documentation. This worked fine over the years while or website links changed several times a year.

hohwille (changes requested):
Thanks for providing a constructive change as PR.
However, look at my review comment. I am most likely to close this PR.
Instead I would add #505 more directly:
IMHO the commandlet should check the further args before passing them to `doMavenArchetype` and also before asking about the workspace.
It would even be possible to ask interactive for the package, artifactId, etc. if omitted and just `devon java create` is called.

hohwille:
replaced by PR #509

Issue #507, opened 2021-01-05
switch eclipse edition


tpontz:
As a user, I want to configure eclipse edition so that I can choose between standard edition an enterprise edition.

Affected version:

OS: Windows 10
devon IDE: 2020.12.001

PR #508, opened 2021-01-05
#507 eclipse edition can be switched by variable

tpontz:
Implementation for #507


hohwille (approved):
@tpontz thanks for this great PR 👍 
Your code change looks fine, variable name makes sense and documentation is also changed properly. Wow! I will directly merge this.

Issue #522, opened 2021-01-22
update devon4j to 2020.12.002 and create migration to this version


hohwille:
We need to update devon4j version to 2020.12.002 (in submodule and migrations).
PR for last release can be found here: https://github.com/devonfw/ide/pull/503/files

Updates of libs all from this PR: https://github.com/devonfw/devon4j/pull/310

hohwille:
Anybody available to implement this story?

PR #272, opened 2019-10-09
revision metadata changes for devon4j

vapadwal:
This is PR for https://github.com/devonfw/ide/issues/237

PR #525, opened 2021-01-28
#522 :update devon4j to 2020.12.002

suvmanda:
Implements #522

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -1,6 +1,9 @@
 = Changelog of devonfw-ide
 
 This file documents all notable changes to https://github.com/devonfw/ide[devonfw-ide].
+== 2020.12.002
+
+* https://github.com/devonfw/ide/issues/522[#522] :update devon4j to 2020.12.002 and create migration to this version
 
 == 2020.12.001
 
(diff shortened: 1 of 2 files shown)


hohwille (changes requested):
@suvmanda Excellent. This is looking very good. 👍 
I just merged your PR in `ide-settings`. Can you include the submodule update into this PR to fix the travis build?
Also there are two other version updates missing from the PR I linked:
```
   <guava.version>30.0-jre</guava.version>
    <junit.version>5.7.0</junit.version>
```
Can you please add them as well?

hohwille (approved):
@suvmanda Thanks for your updates. Now travis is green and all is perfect. 👍 

FYI: I removed the line with the `junit.version` update as it was my mistake to suggest it:
In #272 we wanted to migrate projects automatically from `juni4` to `junit5` and we figured out that this migration is too complex and we will not be able to provide this in a reliable way. So we decided not to force projects to uprade to `junit5` as the added value is limited while the effort may be high for the project.
As `junit` is a test-tool, CVEs and security updates do not have the highest priority and projects can do that themselves as needed. So for simplicity lets avoid having advanced logic that only updates to the latest `junit` version `5.7.0` if the current `junit` version is starting with `5.`.

PR #526, opened 2021-02-02
VS Code and node.js LTS latest updates

sjimenez77:
As usual here it is the update for these packages:

- VS Code updated to 1.52.1
- node.js LTS updated to v14.15.4 _(previous LTS version was a node 12 one!)_


maybeec (approved):

hohwille (approved):
@sjimenez77 awesome. Thanks again for maintenance and keeping this up-to-date.

PR #527, opened 2021-02-11
devonfw/ide#523: check snapshot versions from dependencies on release

### LINKED ISSUE #523: improve release to check for SNAPSHOT versions
`devon release` massively boosts automation of releases and makes my life easier.
However, I expect a release to have no `*-SNAPSHOT` verions in `pom.xml` files of maven projects.
Current behaviour is that `devon release` does not warn or complain in such case.

Therefore I would suggest to improve `devon release` such that it also checks for `*-SNAPSHOT` verions when it analyzes the `pom.xml` files. If a `*-SNAPSHOT` verion is present it should render a _strong_ warning and ask the user for confirmation before proceeding (and in `batch` mode fail immediately if not `force` is also active).
This feature would improve `devon release` and prevent users from creating "broken" releases saving time and despair.

tobka777:
fixes #523

### DIFF
--- a/documentation/advanced-tooling-generic.asciidoc
+++ b/documentation/advanced-tooling-generic.asciidoc
@@ -10,7 +10,7 @@ If you are looking for a git client that works cross-platform we recommend to us
 To draw diagrams for your project or for blueprints in devonfw, we recommend the following cross-plattform tools:
 
 * https://draw.io/[draw.io] is a powerful generic vector painting program (similar to visio). You can get a free open-source edition for your desktop from https://github.com/jgraph/drawio-desktop/releases[here].
-* https://www.objectaid.com/[ObjectAid] is a nice and easy to use eclipse plugin that you can use to quickly create UML diagrams. While class-diagrams are supported for free, you need to buy a license if you want to use the other diagram types.
+* https://www.objectaid.com/[ObjectAid] is a nice and easy to use eclipse plugin that you can use to quickly create UML diagrams from existing code. While class-diagrams are supported for free, you need to buy a license if you want to use the other diagram types.
 * https://plantuml.com/[PlantUML] is a great tool that can render UML diagrams from simple markup that can be easily managed in `git` or other version-control systems together with your code. Its simplicity allows branching and merging unlike other greedy binary UML data-formats.
 
 == Browser Plugins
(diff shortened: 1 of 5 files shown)


hohwille:
Please fix these shellcheck issues:

```

In scripts/command/mvn line 242:

  if [ ! -z "${snapshot_list}" ]

       ^-- SC2236: Use -n instead of ! -z.

In scripts/command/mvn line 244:

    if  [ ! -z "${batch}" ] && [ -z "${force}" ]

          ^-- SC2236: Use -n instead of ! -z.
```

hohwille (approved):
@tobka777 thanks for your PR. This is really great work 👍 
You solved this tricky problem nicely and also updated documentation consistently.

I also tested your change with a dummy project. The only problem I found is that if SNAPSHOT dependencies are found they are printed multiple times:
```
devon release
 Parent with non-existing relativePath ../pom.xml
 You are inside a top-level maven project
WARNING: It is not recommended to create releases with SNAPSHOT dependencies!
The following dependencies are SNAPSHOT versions:
    io.github.m-m-m:mmm-base:jar:1.0.0-SNAPSHOT:compile
    io.github.m-m-m:mmm-base:jar:1.0.0-SNAPSHOT:compile
    io.github.m-m-m:mmm-base:jar:1.0.0-SNAPSHOT:compile
    io.github.m-m-m:mmm-base:jar:1.0.0-SNAPSHOT:compile
    io.github.m-m-m:mmm-base:jar:1.0.0-SNAPSHOT:compile
We strongly recommend to abort and use release versions.

Do you want to continue?
(yes/no):
```
I added this SNAPSHOT dependency only once (as single dependency in the parent POM). As it is inherited to all child modules it is listed multiple times.
As a nice to have improvement we could pipe the output to `sort -u` that would eliminiate duplicates.

Anyhow, I already approve as your feature is working nicely. It takes some noticable additional time (couple of seconds) but IMHO all fine and reasonable.

hohwille (approved):
Excellent. Tested. Works like a charm now:
```
WARNING: It is not recommended to create releases with SNAPSHOT dependencies!
The following dependencies are SNAPSHOT versions:
    io.github.m-m-m:mmm-base:jar:1.0.0-SNAPSHOT:compile
We strongly recommend to abort and use release versions.
```

PR #529, opened 2021-02-17
Github actions for CI/CD

sujith-mn:


.github/workflows/sync-docs.yml
```diff
@@ -0,0 +1,71 @@
+env:
+    REPO_CONSOLIDATE: "devonfw-guide"
+    BUILD_USER: ${{ secrets.BUILD_USER }}
+    BUILD_USER_PASSWD: ${{ secrets.BUILD_USER_PASSWD }}
+    BUILD_USER_EMAIL:  ${{ secrets.BUILD_USER_EMAIL }}
+    EMAIL: "icsddevonfwsupport.apps2@capgemini.com"
```
maybeec, line 6: Please don't put the email here. Take it from the secrets like in other repositories

maybeec (changes requested):

.github/workflows/sync-docs.yml
sujith-mn: Sure.. the BUILD_USER_EMAIL will be used.

hohwille (changes requested):
Seems there has been some confusion in the sync of the different approaches how to split and name the github actions. Can you please revert the last commit.

hohwille (changes requested):
@sujith-mn Looks goot to me now.
However, you are mixing indendation with 4 spaces and with 2 spaces. Can you use 2 spaces indendation everywhere as YAML really relies on indendation and readability is getting bad when indedation witdth is inconsistent. Thanks. Appart from that ready to be merged.

hohwille (approved):
I fixed the indent so we can merge and make progress...

maybeec (approved):

Issue #531, opened 2021-02-22
migration from travis CI to github actions


hohwille:
Due to limitations and problems with travis-ci we want to migrate to github actions for all our devonfw repositories.

hohwille:
Still not working:
```
 Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.971 s <<< FAILURE! - in com.devonfw.tools.ide.migrator.MigrationsTest
Error:  testMigrations(com.devonfw.tools.ide.migrator.MigrationsTest)  Time elapsed: 0.81 s  <<< ERROR!
java.nio.file.NoSuchFileException: ../settings/src/main/settings/eclipse/archetype-catalog.xml
	at com.devonfw.tools.ide.migrator.MigrationsTest.testMigrations(MigrationsTest.java:52)
```

Seems as if cloning is not happening recursive and therefore submodules are missing.

hohwille:
https://github.com/marketplace/actions/checkout-submodules

hohwille:
This is still not done. I just noticed in PR that we now do not have any CI in place anymore:
https://github.com/devonfw/ide/pull/540

maybeec:
@hohwille, what exactly do you mean by that?
 
![grafik](https://user-images.githubusercontent.com/1427255/113339258-85585400-932a-11eb-8e67-9a0fec537331.png)

hohwille:
OSSRH nexus deployment now also works.
I can close this issue as all is working now as planned.

PR #533, opened 2021-02-26
#531: deleted .travis.yml

hohwille:
To complete #531 I deleted `.travis.yml` config.
We will see if now PR is verified by GH-action and all hooks pass.


hohwille:
Travis check is still required:
```
Travis CI - Pull Request Expected — Waiting for status to be reported 
```
And github actions are not showing up as PR check
So this definetly needs some more tweaks!

maybeec (approved):

hohwille:
So we now have github actions and removed travis but now we do not have any working CI as GH actions are broken and the status for PR is never reported. If we do not get urgent assistence, I have to go back travis to be fully operable again...

PR #534, opened 2021-03-04
Update VS Code to 1.54.1 and node.js to v14.16.0

sjimenez77:
As usual, my tiny contribution upgrading this components:

- VS Code 1.54.1
- node.js 14.16.0 LTS


hohwille (approved):
@sjimenez77 Thanks 👍

PR #540, opened 2021-03-17
devonfw/ide#287: command autocompletion

### LINKED ISSUE #287: Command autocompletion
As a develop, could be useful to have some hints while I'm writting a devon-ide command.

For example: while I'm writting the command `devon java` could be useful to have some help if you forget some command, so if you press TAB it will show you the commands available for `devon java`. 

It's a nice to have. I don't know if it is possible to do for all consoles (cmd, powershell, bash...)

tobka777:
fixes #287


scripts/src/main/resources/scripts/devon
```diff
@@ -72,24 +72,50 @@ then
...
   then
     touch ~/.zshrc
   fi
   if ! grep -q "alias devon=" ~/.zshrc
   then
     echo "Installing devon alias to your .zshrc"
-    echo -e '\nalias devon="source ~/.devon/devon"\ndevon' >> ~/.zshrc
+    {
+      echo -e '\nalias devon="source ~/.devon/devon"\ndevon'
+      echo -e "autoload -Uz compinit\ncompinit\nautoload bashcompinit\nbashcompinit\n${AUTOCOMPLETION}"
```
hohwille, line 99: Very nice but I have some questions:
* what if the user already has some manual autocompletion in his shell config (`.zshrc`)? Will duplicate invocation of `compinit` cause any harm? Is that always available on all platforms as it is a build-in command of both `bash` and `zsh`?
* what if the user already has devonfw-ide installed? Should we offer a central command to install or uninstall the autocompletion in `ide` commandlet instead and invoke that from the existing places?

scripts/src/main/resources/scripts/devon-autocomplete
```diff
@@ -0,0 +1,34 @@
+#/usr/bin/env bash
+_devon_completions()
+{
+  local cur commandlist options
+  COMPREPLY=()
+  source ~/.devon/home
```
hohwille, line 6: here we also have the case as above but this time it is not easy to solve.
Can you try to research in which context this `_devon_completions` function is invoked? Is it possible to determine the current IDE installation based on `$PWD` like it is done in `environment-project` script?

hohwille (changes requested):
@tobka777 Thanks for this PR. You again managed to surprise me with great quality. You reverse-engineered all commands to get the proper list of args. You even found some existing missing options in the help output that you fixed alongside. 👍 

However, I added very few review comments that you should have a look and see if you can rework and resolve your PR. Thanks in advance.

scripts/src/main/resources/scripts/command/help
```diff
@@ -1,4 +1,17 @@
 #!/usr/bin/env bash
+
+# autocompletion list
+if [ "${1}" = "shortlist" ]
+then
+  if [ -z "${2}" ]
+  then
+    # shellcheck disable=SC1090
+    source ~/.devon/home
```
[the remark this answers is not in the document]
tobka777: When I remove `~/.devon/home` then I didn't get `DEVON_IDE_HOME` on MACOSX. Maybe it works on other OS, but not on MACOSX.
`ls: /scripts/command/: No such file or directory`
Yes it is called directly called from bash completion: `devon help<TAB><TAB>`
But now I tried with `$(dirname "${BASH_SOURCE:-$0}")` and it's works. (see next commit)

scripts/src/main/resources/scripts/devon
tobka777: - You are right, I have not tested this case. There seems to be problems when compinit already exists. I will check if compinit or bashcompinit exists before adding the config. I tested this on the Mac and didn't have to install anything to do it. I will test it on Linux later, but think it will work.
- If `devon-ide` is installed, `devon-ide` must be reinstalled. Alternatively I can extend the `devon ide update` command that it extends the configuration. Basically I think that it is not useful to have an extra command to install the autocomplete, because actually every user finds this useful and can use it. After all, there are no disadvantages to having autocompletion.

scripts/src/main/resources/scripts/devon-autocomplete
tobka777: It works very well with the same command `$(dirname "${BASH_SOURCE:-$0}")`. It returns me the path where the autocompletion script is located. (see next commit)

scripts/src/main/resources/scripts/devon
tobka777: I tested it with WSL Ubuntu and zsh with autocomplete works fine.

hohwille: > Basically I think that it is not useful to have an extra command to install the autocomplete, because actually every user finds this useful and can use it.

My intention was not to make it optional but just to have the ability to "reinstall" it e.g. if the initial installation failed, has been lost due to changes in `.bashrc` or `.zshrc`, etc.

scripts/src/main/resources/scripts/devon-autocomplete
hohwille: The autocomplete script is registered globally on the first setup of your devonfw-ide. The following things can happen:
* A newer version of devonfw-ide is installed in a separate folder that has additional commands. Those will then not be available for autocompletion
* After additional installation(s) of `devonfw-ide` have been added the first initial devonfw-ide installation was removed. The config written to `.bashrc` or `.zshrc` was pointing to that installation with an absolute path and therefore will be broken.

So to make it "perfect" we would need to do the same trick as we did with `devon[.bat]` script:
* create a minimal but stable and universal script as shim also for `devon-autocomplete`
* this can be installed to a central installation (`~/scripts` just like with `devon.bat` or `~/.devon`) on initial setup and the be registered to `.bashrc` and `.zshrc`.
* This `devon-autocomplete` script itself would work portable such that is uses the current `PWD` where the user is currently located to find `DEVON_IDE_HOME` (see `environment-project`).

hohwille:
Sorry, I was quite busy with other stuff. This PR is looking very promising now. 👍 
I still have some comments left for perfection.
However, we could still consider to merge this PR as is and then create a new story for these remaining aspects to make it perfect. Lets have a chat after easer celebration days to discuss if we proceed this way.
Thanks again for your great work.

tobka777:
Now the autocompletion is made dependent on the installation folder. For this I created a file `autocomplete`, which is copied to the folder `~/.devon/` during the installation. If you are not in the IDE folder, there is no autocomplete.

scripts/src/main/resources/scripts/devon
```diff
@@ -72,24 +72,61 @@ then
...
   then
     touch ~/.bashrc
   fi
   if ! grep -q "alias devon=" ~/.bashrc
   then
     echo "Installing devon alias to your .bashrc"
-    echo -e '\nalias devon="source ~/.devon/devon"\ndevon' >> ~/.bashrc
+    {
+      echo -e '\nalias devon="source ~/.devon/devon"\ndevon'
+      echo -e "${AUTOCOMPLETION}"
```
hohwille, line 87: This approach will not add the autocompletion feature to computers where devonfw-ide has already been "installed" before in an older version.

hohwille (approved):
@tobka777 Thanks for your rework 👍 
I am happy with the current solution. We need to test this by 1-2 more people and then can merge it.
I also added one last comment for perfection. However, this can also be addressed after the merge by a new PR...

scripts/src/main/resources/scripts/devon
tobka777: This is correct! Should we change this or we say to the people, please reinstall the ide.

hohwille:
I have tested this new feature locally and got this strange effect:
```
joerg@LAPTOP-65UUDRL1 MINGW64 /d/projects/test/workspaces
$ devon
build*     eclipse*   help*      ionic*     mvn*       npm*       sonar*
cicdgen*   gradle*    ide*       java*      ng*        project*   vscode*
cobigen*   help       intellij*  jenkins*   node*      release*   yarn*

joerg@LAPTOP-65UUDRL1 MINGW64 /d/projects/test/workspaces
$ devon build*
Unknown command build*
```
So for some reason most commands have an asterisk appended and completion adds it including the asterisk what causes an error.
@tobka777 Can you have a look. Did this not happen on your end as well?

tobka777:
> I have tested this new feature locally and got this strange effect:
> 
> ```
> joerg@LAPTOP-65UUDRL1 MINGW64 /d/projects/test/workspaces
> $ devon
> build*     eclipse*   help*      ionic*     mvn*       npm*       sonar*
> cicdgen*   gradle*    ide*       java*      ng*        project*   vscode*
> cobigen*   help       intellij*  jenkins*   node*      release*   yarn*
> 
> joerg@LAPTOP-65UUDRL1 MINGW64 /d/projects/test/workspaces
> $ devon build*
> Unknown command build*
> ```
> 
> So for some reason most commands have an asterisk appended and completion adds it including the asterisk what causes an error.
> @tobka777 Can you have a look. Did this not happen on your end as well?

tobka777:
I did not have this effect. I don't know why you get it. Have you any idea?

grafik

hohwille:
> I did not have this effect. I don't know why you get it. Have you any idea?

I assume this is a platform specific issue (ocurs on windows and not on MacOS).
My assumption is that the `*` is the executable flag that MINGW adds to `ls` command by default:
https://github.com/devonfw/ide/blob/1919f706a4d1065f31ab27ef09102b2fcf3368dd/scripts/src/main/resources/scripts/devon-autocomplete#L20

So you need to explicitly supply options to `ls` supressing this.

hohwille:
https://superuser.com/questions/178786/what-does-the-asterisk-mean-after-a-filename-when-you-type-ls-l

hohwille:
Fixed asterisk issue and retested on windows. IMHO now ready to be merged.
@tobka777 I am happy if you can retest the feature on your end once more. However, I assume that it should now work on all platforms and will merge in optimistic expectation. We can still create a PR to resolve any potential issues as well as the already discussed improvements to activate autocompletions on update on machines with existing IDE installation.
Thanks again for your great work 👍

tobka777:
Oh yes. I also thought that this could be a platform dependent error. I think we would have to do the trick in one more place. Or works `devon help <tab><tab>` without stars?

https://github.com/devonfw/ide/blob/1919f706a4d1065f31ab27ef09102b2fcf3368dd/scripts/src/main/resources/scripts/command/help#L8

Anyway, I'm glad we have this feature now.

PR #543, opened 2021-04-01
Make conf/vscode - directory also managed/updated by devon

jkrainz-cg:
The configuration of vscode is distributed across multiple directories.

The directory "conf/vscode" is currently not managed by devon.
Unfortunately, that is where the setting to disable vs code telemetry is located.

Thus this PR: to also manage that directory with devon.

### DIFF
--- a/scripts/src/main/resources/scripts/command/vscode
+++ b/scripts/src/main/resources/scripts/command/vscode
@@ -4,6 +4,9 @@ if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
 source "$(dirname "${0}")"/../functions
 cd "${DEVON_IDE_HOME}" || exit 255
 
+VSCODE_CONF="${DEVON_IDE_HOME}/conf/vscode"
+VSCODE_EXT="${DEVON_IDE_HOME}/software/vscode-extensions"
+
 function doSetup() {
   if [ -n "${1}" ] || [ ! -d "${VSCODE_HOME}" ]
   then
@@ -30,7 +33,7 @@ function doSetup() {
         chmod a+x "${VSCODE_HOME}/bin/code"
       fi
     fi
-	
+
     cleanupPlugins
     doAddPlugins
   fi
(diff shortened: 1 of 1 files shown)


scripts/src/main/resources/scripts/command/vscode
```diff
@@ -94,13 +97,12 @@ function doConfigureVsCode() {
   if [ -n "${mode}" ]
   then
     doConfigureWorkspace "${SETTINGS_PATH}/vscode/workspace" "${WORKSPACE_PATH}" "${mode}"
+    doConfigureWorkspace "${SETTINGS_PATH}/vscode/conf/vscode" "${VSCODE_CONF}" "${mode}"
```
hohwille, line 100: This does not make any sense to me. There is no such folder:
https://github.com/devonfw/ide-settings/tree/master/vscode
IMHO this is all a missunderstanding of the way devonfw-ide works and is designed.
Can you provide more details about your use-case and what you actually want to archive?

However, there is this folder that acts as a template for configuration during initial setup:
https://github.com/devonfw/ide-settings/tree/master/devon/

If you just want to have some initial config that preconfigures something for the end-user, just put it inside this structure. This is documented here:
https://github.com/devonfw/ide/blob/master/documentation/settings.asciidoc

hohwille (changes requested):
@jkrainz-cg thanks for your contribution and for trying to improve devonfw-ide.
I really appreciate this 👍 
However, I think there are some things unlcear. Maybe my review comment will help to clarify. Otherwise do not hesitate to get in touch.

scripts/src/main/resources/scripts/command/vscode
jkrainz-cg: Maybe looking at the other PR i have will clear things up a little: https://github.com/devonfw/ide-settings/pull/15/files

So for background: VS Code has some settings with unacceptable defaults (telemetry, but there will probably be more...). So we need to configure those settings.

It is probably not acceptable to require our developers to re-install devon completely when we notice a missing setting, so using the template for configuration during initial setup is probably out of the question.

The settings this concerns reside exactly in the folder "${DEVON_IDE_HOME}/conf/vscode", so this PR simply makes that folder managed like the other configuration folder.

hohwille:
I have created issue #553 and willl create a PR for it.

hohwille:
replaced by PR #554

hohwille:
@jkrainz-cg thanks again for your contribution and the tricks how to improve vscode settings and disable telemetry. 👍 We have added your valuable input to devonfw-ide even though it did not happen directly via this PR. Feel free to test and give feedback if all is working as expected. The next official release 2021.04.001 will appear in ~2 weeks including your nice feature!

Issue #542, opened 2021-03-26
Support placeholders in settings.xml template


hohwille:
As a user of devonfw-ide, I want to be able to put placeholders for logins and passwords into the `settings.xml` template inside my settings git repo so that during setup of the ide the user will be prompted for these placeholders and in case of credentials these will automatically be encrypted via `maven -ep`.
Further, it should be possible to repeat only this step to "instantiate" the `settings.xml` again overwriting the existing one (suggestion: `devon mvn update-settings`) e.g. when my password has changed (yes, some systems still force their users to change their password in a specific frequency even though it has been proven that this is an anti-pattern reducing security instead of increasing it - https://www.sans.org/security-awareness-training/blog/time-password-expiration-die).

This is a consequent step to automate the setup, simplify the standard processes of users joining a project.

hohwille:
Obvious syntax would be something like `${...}` however, this is already used by maven itself and can be used to put variable expressions in your `settings.xml` that should not be resolved.
So I would suggest to use `$[...]` instead.
Further I see the following aspects:
* which variable value has to be encrypted and how can we distinuish? KISS would be to encrypt all variables.
* what will be prompted to the end-user when he is asked for the value? Is the name of the variable itself already sufficient? Will the IDE admin choose an appropriate name that is self explanatory? For KISS, I would leave it with the variable name. A more complex alternative would be to have a syntax like `$[«variable»:«prompt text»]`. As a colon is typically not used in variable names, such feature can still be added later if really needed.
* if a variable with the same name is used multiple times in the template, the user should be asked only once and the value has to be filled into all occurrences.

hohwille:
In general this is easy. However, with `bash` easy things can turn out to be tricky.
The thing is that the variable to replace as well as the replacement will come from a variable itself. However, passing this to tools like `sed` will cause problems as they do not support easy literal replacements so some escaping magic would be needed. On the other hand it would be even easier to use bash internal functions for replacement (see #358) by reading the entire `settings.xml` template into a variable and doing the replacements in memory within bash but the problem then is that bash itself already interpolates variables such as `${USER}` to its actual value or `$[...]` as `0` what is entirely unintendet.
I am really thinking of implementing this with Java as there it is trivial to solve but launching a JVM for such a simple replacement seems some kind of overkill. Still searching if there is an easy bash solution...

hohwille:
Was a little tricky to figure out as the bash documentation is lacking how to solve this but I found the solution here:
[...]

PR #544, opened 2021-04-06
#542: support secrets from variables in settings.xml

hohwille:
Implements issue #542


hohwille:
Should we change our `settings.xml` template now to be ready for production-line or should we leave this up to the projects setting up their own settings git (to avoid annoying users who just want to jumpstart the IDE)?

hohwille:
Shellcheck broke the build. Sorry, I should have checked in advance. `SC2155` is already fixed (trivial) but suggested pattern to fix `SC2013` does not work.
I changed this line
```
          for var in $( grep -o '\$\[.*\]' "${SETTINGS_PATH}/devon/conf/.m2/settings.xml" | sort -u)
```
to this line:
```
          grep -o '\$\[.*\]' "${SETTINGS_PATH}/devon/conf/.m2/settings.xml" | sort -u | while IFS= read -r var
```
As suggested here:
https://github.com/koalaman/shellcheck/wiki/SC2013

But testing reveals that then the entire thing is not working anymore and the variable `var` is set to the empty value.
I will rather add an exclusion of this shellcheck warning.
If some expert like @markusschuh may have a hinter later how to properly solve this `SC2013` issue, we can sill go for it but I wont let this issue get blocked by some shellcheck suggestion that does not seem to work.

documentation/conf.asciidoc
```diff
@@ -23,7 +23,7 @@ This folder contains configurations for your IDE:
 The `.m2` folder is used for configurations of link:mvn.asciidoc[maven]. It contains the local `repository` folder used as cache for artifacts downloaded and installed by maven (see also https://maven.apache.org/guides/introduction/introduction-to-repositories.html[maven repositories]).
 Further, there are two configuration files for maven:
 
-* https://maven.apache.org/settings.html[settings.xml] initialized from a template from your `devonfw-ide` xref:settings[]. You may customize this to your needs (configuring HTTP proxies, credentials, or other user-specific settings).
+* https://maven.apache.org/settings.html[settings.xml] initialized from a template from your `devonfw-ide` xref:settings[]. You may customize this to your needs (configuring HTTP proxies, credentials, or other user-specific settings). Secrets can be specified as `$[«variable.name»]` and will be prompted, encrypted and replaced automatically during the setup (unless in batch mode).
```
maybeec, line 26: Could you provide a link here to the section which explains what batch mode is?

scripts/src/main/resources/scripts/command/mvn
```diff
@@ -30,6 +30,28 @@ function doSetup() {
...
+        local secret
+        local crypted
+        local settings
+        settings=$(cat "${SETTINGS_PATH}/devon/conf/.m2/settings.xml")
+        if ! doIsBatch
+        then
+          # shellcheck disable=SC2013
+          for var in $( grep -o '\$\[.*\]' "${SETTINGS_PATH}/devon/conf/.m2/settings.xml" | sort -u)
+          do
+            read -r -p "Please enter secret value for variable ${var}: " secret
```
maybeec, line 45: Does this allow empty values? What if you don't have passwords at the moment? E.g. A person did not setup sonatype login yet?

maybeec (changes requested):

maybeec:
> Should we change our `settings.xml` template now to be ready for production-line or should we leave this up to the projects setting up their own settings git (to avoid annoying users who just want to jumpstart the IDE)?
> 
> https://github.com/devonfw/ide-settings/blob/9503a2dd6945a0cb38eac1ee5e775032b65ede60/devon/conf/.m2/settings.xml#L23

A valid point. Another suggestion would be to allow configuring mandatory credentials and optional ones by naming conventions on the variable names in the templates. Thus people get prompted on installation with only the mandatory set and we provide devon ide commands to reset the mandatory passwords (in case of password change) and one command to set all 'not yet set credentials.
Wdyt?

scripts/src/main/resources/scripts/command/mvn
hohwille: sure. You can type any password, login, or nonsense here including the empty string. I tested with empty string and `mvn -ep` will give you `{}` that then ends up as the replacement. I think this is OK.

documentation/conf.asciidoc
hohwille: OK, "batch mode" was not documented in asciidoc before either. Added it and linked here.

hohwille:
> A valid point. Another suggestion would be to allow configuring mandatory credentials and optional ones by naming conventions on the variable names in the templates. Thus people get prompted on installation with only the mandatory set and we provide devon ide commands to reset the mandatory passwords (in case of password change) and one command to set all 'not yet set credentials.
Wdyt?

mhm, I was rather thinking that when you do not provide a settings URL on setup (by just hitting return) then the variables are not resolved (like in batch mode). I do not want to overengineer this. If your password has changed or your recieve a password after the setup you can simply delete your `settings.xml` file and recreate it via `devon mvn setup` that will prompt again for the variables/passwords.

maybeec (approved):

PR #547, opened 2021-04-08
#536: safe downloads

### LINKED ISSUE #536: Improve handling of aborted downloads
During IDE updates DevonFW IDE downloads a lot of larger files.
If you interrupt a download a partially downloaded files is kept.
If you then restart the update process the script detect this files as already downloaded and tries to unzip/install it. This fails since the file isn't complete.
To resolve this you have to manually delete the file before restarting the update.
This is not clear to endusers.

We should find a better approach here.

A solution could be to download files to a temp file and rename them after download is completed.

hohwille:
Fixes #536.
Besides stopped downgrading Java to 1.8 in integration tests as tools such as Eclipse do not support this version anymore.

### DIFF
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -372,7 +372,7 @@ function doDownload() {
   then
     curl_opt="--silent"
   fi
-  tmp_file=$(mktemp -ut "${filename}")
+  tmp_file=$(mktemp -ut "${filename}.XXXXXX")
   curl ${curl_opt} -fL "${1}" -o "${tmp_file}"
   result=${?}
   if [ "${result}" != 0 ]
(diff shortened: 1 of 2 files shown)


hohwille:
Nice: github actions/CI reveald that my approach is not working on linux:
```
mktemp: too few X's in template ‘java-11.0.9.1_1-linux.tar.gz’
```

I will imporve and update this PR later.

maihacke:
Code looks code. @hohwille has done testing on MacOS. Linux is covered by CI. Somebody should tests this on Windows. I cannot do this, since I'm also on Mac.

maihacke (approved):

hohwille:
@maihacke Thanks 👍 
We will do various tests on windows before the release.

PR #551, opened 2021-04-08
#542: #544: do not prompt for secrets/variables for default settings

hohwille:
do not prompt for secrets/variables for default settings for issue #542 as discussed in PR #544.

### DIFF
--- a/scripts/src/main/resources/scripts/command/mvn
+++ b/scripts/src/main/resources/scripts/command/mvn
@@ -38,9 +38,7 @@ function doSetup() {
         local settings
         local settingsUrl
         settings=$(cat "${SETTINGS_PATH}/devon/conf/.m2/settings.xml")
-        cd "${SETTINGS_PATH}"
-        settingsUrl="$(git remote get-url origin)"
-        cd -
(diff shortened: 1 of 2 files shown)


hohwille:
I am loosing my faith... Why did shellcheck find issues and therefore failed build for commit https://github.com/devonfw/ide/pull/551/commits/edaedff2caee68cdbb2b299153a72ffcc553f498 but then succeeded the build for commit https://github.com/devonfw/ide/pull/551/commits/950963c1826f803da9831f667ba3f8ca3eb28dbd that contains the same script with the same bug pattern?

hohwille:
Here is the first commit with the shellcheck issue:
https://github.com/devonfw/ide/blob/edaedff2caee68cdbb2b299153a72ffcc553f498/scripts/src/main/resources/scripts/command/mvn#L41
And this is the second commit that shellcheck did not compain about:
https://github.com/devonfw/ide/blob/950963c1826f803da9831f667ba3f8ca3eb28dbd/scripts/src/main/resources/scripts/command/mvn#L41

Does anybody have an explanation for this? That seems entirely nuts...

PR #559, opened 2021-04-15
#287: improvements of command autocompletion

tobka777:
Some improvements of command autocompletion #287:
- `ls` fix for windows on `devon help`
- install autocompletion if not already installed on `devon ide setup`

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -173,9 +173,34 @@ function doUpdateCliScript() {
   fi
 }
 
+function doInstallAutocompletion() {
+  # Install autocompletion
+  devon_autocomplete="${DEVON_HOME_DIR}/.devon/autocomplete"
+  if [ ! -f "${devon_autocomplete}" ]
+  then
+    cp "${DEVON_IDE_HOME}/scripts/autocomplete" "${devon_autocomplete}"
+    AUTOCOMPLETION="source ~/.devon/autocomplete"
+    echo -e "${AUTOCOMPLETION}" >> ~/.bashrc
+    if ! grep -q "compinit" ~/.zshrc
+    then
+      echo -e 'autoload -Uz compinit\ncompinit' >> ~/.zshrc
+    fi
+    if ! grep -q "bashcompinit" ~/.zshrc
+    then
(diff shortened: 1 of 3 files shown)


tobka777:
I have now implemented both improvements. Now the autocompletion should also be set up during setup. For this I wrote a function in the `ide` command.

scripts/src/main/resources/scripts/command/ide
```diff
@@ -173,9 +173,34 @@ function doUpdateCliScript() {
...
+    fi
+    if ! grep -q "bashcompinit" ~/.zshrc
+    then
+      echo -e 'autoload bashcompinit\nbashcompinit' >> ~/.zshrc
+    fi
+    echo -e "${AUTOCOMPLETION}" >> ~/.zshrc
+    if doIsMacOs
+    then
+      echo -e "${AUTOCOMPLETION}" >> ~/.bash_profile
+    fi
```
hohwille, line 196: why do we need this on MacOS? It do not see we need this at all or why it is OS specific...

hohwille (approved):
@tobka777 thanks for this PR. 👍 
Can we now remove the autocompletion installation from the devon script to avoid duplicated code?

tobka777:
> @tobka777 thanks for this PR. 👍
> Can we now remove the autocompletion installation from the devon script to avoid duplicated code?

That's right, I didn't think of that anymore. I think that is no longer necessary because `ide setup` is run with devon install. I will remove it today or tomorrow and check if everything still works as usual.

scripts/src/main/resources/scripts/command/ide
tobka777: This is OS specific. Mac uses `.bash_profile` and Linux use `.bashrc`. I added for this autocompletion `.bash_profile`to the code that it also runs with MacOS. We can also add a condition for Linux, because it is also not useful for mac, but maybe for windows. This is the reason why only add the condition for  `.bash_profile`, because it is only for MacOS.

hohwille (approved):
@tobka777 excellent. Now it is perfect 👍 
I also like that now all the logic to install and uninstall is in one place (ide commandlet) what increases maintainability. I will merge now.

Issue #557, opened 2021-04-13
minimize setup by reducing DEVON_IDE_TOOLS


hohwille:
The setup of devonfw-ide should be fast for better UX.
Therefore we should remove tools that need longer downloads and installation for the default tools:
https://github.com/devonfw/ide/blob/f03a5885301047abec244112f9fe5203189ea288/scripts/src/main/resources/scripts/devon.properties#L7

All tools are always available via lazy initialization if not explicitly configured in `DEVON_IDE_TOOLS`.
The only benefit of having a tool in `DEVON_IDE_TOOLS` is that the tool is immediately there and you could even get offline.

However, the drawback is that it takes very long to setup a vanilla devonfw-ide and new users might think "WTF is going on?"
As people tend to become more and more impatient we should consider reducing this to the minimum.
I would at least kick out:
* Eclipse
* VSCode
* maybe even `ng`

We could even be radical and make the list empty by default.

Please note that any project is still free to customize this `DEVON_IDE_TOOLS` to their own needs. However, with the default settings, we do not know what is the use-case of the user. Will he need java or node.js tooling at all?
Preinstalling this therefore seems annoying and waste.

hohwille:
One thing to consider is that the launch scripts like `eclipse-main.bat` are currently only created if the according IDE (in this example `eclipse`) is listed as tool. I never use them as users can simply run `devon eclipse` to lauch eclipse but newbe users might not be familiar with command shells and will be lost. We should consider still having such launch scripts generated on setup but downloading and installing the IDE on the first launch.

maybeec:
Actually, I would like to get a clean vanilla devonfw ide not installing anything besides extracting the scripts by default.

To anyhow provide valuable tools to handle bulk installing without the need to actually do installation step by step again based on any project tutorial, I would like to propose a slide adaption to the current possibilities of handling software configuration.

I loved the idea of oomph having a basic eclipse only, possibly even vanilla and let the project decide what is needed for development.
E.g., imagine we introduce something like software dependencies for project. Meaning that in the project configuration, you could think of another property `software: (eclipse, maven, java)`

Then `devon project setup projectname` (not 100% sure about the command) would also lead to automatic installation of the software needed for development.

Thus (as we are just talking about the devonfw space having this problems), you could start with vanilla and simply saying which project you want to work on, you would get the necessary tools installed without the need to think about it.

hohwille:
[...]

PR #561, opened 2021-04-19
#557: minimize tools on setup

hohwille:
implements #557

### DIFF
--- a/scripts/src/main/resources/scripts/command/eclipse
+++ b/scripts/src/main/resources/scripts/command/eclipse
@@ -279,7 +279,6 @@ do
   elif [ "${1}" = "setup" ]
   then
     doSetup setup
-    doCreateIdeScript eclipse
   elif [ "${1}" = "create-script" ]
   then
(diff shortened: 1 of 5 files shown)


hohwille:
I have tested this locally  on my mac. Impacts:
* scripts `eclipse-main[.bat]` and `vscode-main[.bat]` are now always created on setup (no matter what tools are configred (and therefore `intellij-main[.bat]` is not created on setup and needs an explicit call of `devon intellij create-script`).
* I dropped angular CLI that always takes some extra time on setup. That means after setup now by default `ng` command is not available until the first call of `devon ng` or `devon ng setup`. This can be changed by projects by adding `ng` to `DEVON_IDE_TOOLS`
* Currently integration-tests do not setup eclipse or vscode anymore during CI. Pro: faster, Contra: potential bugs not detected. We can also configure `DEVON_IDE_TOOLS` for integration-tests
* Potentially also some impact for tutorials that I can not forsee

PR #570, opened 2021-04-27
Added commandlets for docker and kubernetes setup

isandesh1986:
implements issue #584


hohwille (commented):
Thanks @isandesh1986 - nice progress. Will we in the end add a commandlet that we can do something like:
```
devon docker setup
devon kubernetes setup
```

Those commandlets could detect OS and then do the according steps required or delegate to *.bat file one windows?

scripts/src/main/resources/scripts/setup-kubernetes
```diff
@@ -0,0 +1,14 @@
...
+    echo Installing Kubectl...
+    curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg ^| sudo apt-key add -
+    sudo touch /etc/apt/sources.list.d/kubernetes.list
+    echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" ^| sudo tee -a /etc/apt/sources.list.d/kubernetes.list
+    sudo apt-get update
+    sudo apt-get install -y kubectl
+    echo Kubernetes installed successfully.
+else 
+    echo Kubernetes already installed. 
+fi
```
maybeec, line 13: please also echo the kubernetes version information after fi

scripts/src/main/resources/scripts/setup-docker
```diff
@@ -0,0 +1,17 @@
...
+    sudo apt-get update -y
+    sudo apt-get install -y apt-transport-https ca-certificates curl software-properties-common
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+    sudo apt-key fingerprint 0EBFCD88
+    sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
+    sudo apt-get update -y
+    sudo apt-get install -y docker-ce
+    sudo usermod -aG docker $USER
+    echo Docker installed successfully.
+    docker --version
```
maybeec, line 14: please do docker --verion after fi so it will in any case be outputted

maybeec (changes requested):

scripts/src/main/resources/scripts/setup-docker
isandesh1986: Done

scripts/src/main/resources/scripts/setup-kubernetes
isandesh1986: Done

isandesh1986:
> Thanks @isandesh1986 - nice progress. Will we in the end add a commandlet that we can do something like:
> 
> ```
> devon docker setup
> devon kubernetes setup
> ```
> 
> Those commandlets could detect OS and then do the according steps required or delegate to *.bat file one windows?

Yes Jorg. Taken care now.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+wsl_linux_version=$(wsl -l)
```
hohwille, line 2: even though we do not support other OS than windows yet, could you please alread add OS checks as suggested in last review?
```
if doIsWindows
then
   # ... add your code here
else
  doFail "Sorry, docker-installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
fi
```

Further, I suggested to design these commandlets following the philosophy of other commandlets.
So what you are doing here should only happen for `devon docker setup`.
For any other command (excluding `help`) you should delegate to an existing docker installation and pass the arguments to the docker command.
Finally, we always do a setup in a lazy way. So currently I can run
```
devon mvn clean install
```
And if `mvn` is not yet installed, it will be installed first.
At the end (both if mvn was already present or just has been installed) `mvn clean install` is called.

You should be able to check if `docker` command is present in an OS independent way.
Then it can already be used on all OS but only the automatic installation does not work on OS other than windows.

Please have a look at existing commandlets like `mvn`, `npm`, `java`, etc. and follow our existing philosophy.

scripts/src/main/resources/scripts/command/kubernetes
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+linux_version=$(wsl -l)
```
hohwille, line 2: same as with docker commandlet.

scripts/src/main/resources/scripts/setup-docker
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+docker_check=$(docker --version)
+if [[ $docker_check != *"Docker version"* ]];then
+	echo "Installing Docker..."
+	sudo apt-get update -y
```
hohwille, line 5: will this sudo require a password to be entered by the user? Do we need to prompt something before to explain what kind of password is actually required here to avoid confusion?

hohwille (changes requested):
@isandesh1986 thanks for this great PR 👍 The installation on windows is looking good to me.
However, we need some improvements before this can be merged as if used on Linux or Mac this would give the user a very strange technical error.
Can you please look at my review comments and rework accordingly.
Thanks in advance!

scripts/src/main/resources/scripts/command/docker
maybeec: I think as docker or kubernetes are global services, it does not really make sense to proxy the commands by devon command.
I even did not understood why we are providing this anyhow, I am not using it at all and it makes me feel even more to be working with a framework if I am told to use `devon mvn *` instead of `mvn *`.

hohwille:
BTW: when we add new commandlets we also add documentation as asciidoc, and also integrate that into the existing documentaiton.
* add `docker.asciidoc` and `kubernetes.asciidoc` like this example: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/mvn.asciidoc
* add `include` here: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/devonfw-ide-usage.asciidoc
* add `link` here: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/cli.asciidoc

scripts/src/main/resources/scripts/setup-docker
isandesh1986: Added a hint.

scripts/src/main/resources/scripts/command/docker
isandesh1986: > even though we do not support other OS than windows yet, could you please alread add OS checks as suggested in last review?
> 
> ```
> if doIsWindows
> then
>    # ... add your code here
> else
>   doFail "Sorry, docker-installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
> fi
> ```
> 
> Further, I suggested to design these commandlets following the philosophy of other commandlets.
> So what you are doing here should only happen for `devon docker setup`.
> For any other command (excluding `help`) you should delegate to an existing docker installation and pass the arguments to the docker command.
> Finally, we always do a setup in a lazy way. So currently I can run
> 
> ```
> devon mvn clean install
> ```
> 
> And if `mvn` is not yet installed, it will be installed first.
> At the end (both if mvn was already present or just has been installed) `mvn clean install` is called.
> 
> You should be able to check if `docker` command is present in an OS independent way.
> Then it can already be used on all OS but only the automatic installation does not work on OS other than windows.
> 
> Please have a look at existing commandlets like `mvn`, `npm`, `java`, etc. and follow our existing philosophy.

Have incorporated the OS checks now.

isandesh1986:
> BTW: when we add new commandlets we also add documentation as asciidoc, and also integrate that into the existing documentaiton.
> 
> * add `docker.asciidoc` and `kubernetes.asciidoc` like this example: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/mvn.asciidoc
> * add `include` here: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/devonfw-ide-usage.asciidoc
> * add `link` here: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/cli.asciidoc

sure, will add.

isandesh1986:
> BTW: when we add new commandlets we also add documentation as asciidoc, and also integrate that into the existing documentaiton.
> 
> * add `docker.asciidoc` and `kubernetes.asciidoc` like this example: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/mvn.asciidoc
> * add `include` here: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/devonfw-ide-usage.asciidoc
> * add `link` here: https://github.com/devonfw/ide/blob/0a7d8a6af55829efb6a2a99fa37ca964aef07c48/documentation/cli.asciidoc

Documentation has been pushed for these commandlets.

scripts/src/main/resources/scripts/command/docker
hohwille: Nobody is forced to do `devon mvn *` instead of `mvn *`. However, this allows scripts to be portable and work seamlessly. Otherwise, you would have to write something like:
```
if ! command -v mvn &> /dev/null
then
  devon mvn setup
fi
mvn my-goal
```

This might still not be enough if `mvn` is installed but the devon environment variables have not been set accordingly.
Instead a script can also do
```
devon mvn my-goal
```
and it will always work and even install maven on the fly if not yet available.

Aint the entire goal of our docker and kubernetes support that projects could also provide scripts to provide e.g. a database running via container as a most simple script in a portal and automated way?

If you want we can have a fundamental discussion about the philopsophy of devonfw-ide and if you think that I designed all odd and wrong. However, we IMHO here want to discuss about adding these two commandlets.
I am repeating myself, but I do not accept that all existing commandlets follow a common philosophy for IMHO good reasons and that these two new commandelts should ignore them what would also is required for specific features to work. E.g. you currently can not add `docker` or `kubernetes` to `DEVON_IDE_TOOLS` variable because of that.

maybeec (approved):

hohwille:
just in case something is unclear:
This PR needs rework such that the `docker` and `kubernetes` commandlets follow the style of all the other existing commandlets:
* parse arguments
* arg `setup` triggers the current functionality
* arg `help` prints the help
* in any other case call "silent setup" (will do nothing if the executable native command `docker` or `kubernetes` is available (`command -v`). Then delegate args to that command.

Thanks

hohwille:
FYI: I created an official issue for this story:  #584
There I also wrote some hints for MacOS support. I could also provide bash snipplets or if this PR is merged, I can create another PR that would add some basic MacOS support (semi-automated).

maybeec:
@isandesh1986, @ssarmokadam would you be able to provide a fix for that at a short term or do we need to estimate a task for the next sprint to keep track? Please organize on your own.

isandesh1986:
> @isandesh1986, @ssarmokadam would you be able to provide a fix for that at a short term or do we need to estimate a task for the next sprint to keep track? Please organize on your own.

Hi @maybeec , We have created 2 separate tasks for Docker and Kubernetes and placed it in backlog. Based on the capacity available we will pull it into next sprint mostly. Thanks

isandesh1986:
> just in case something is unclear:
> This PR needs rework such that the `docker` and `kubernetes` commandlets follow the style of all the other existing commandlets:
> 
> * parse arguments
> * arg `setup` triggers the current functionality
> * arg `help` prints the help
> * in any other case call "silent setup" (will do nothing if the executable native command `docker` or `kubernetes` is available (`command -v`) and if not present trigger the setup - that will currently fail if not on windows). Then delegate args to that command.
> 
> Thanks

Hi @hohwille , pushed the changes for the commandlet philosophy for Docker-. please review the same. Thanks

scripts/src/main/resources/scripts/command/docker
```diff
@@ -0,0 +1,72 @@
...
+  exit
+fi
+
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+# Initiates docker setup
+function doSetup() {
+  if doIsBatch
```
hohwille, line 19: Here we should first check if docker command is already available in OS independent way (see other commandlets). If already installed, we are already done.

scripts/src/main/resources/scripts/command/kubernetes
```diff
@@ -0,0 +1,72 @@
...
+  exit
+fi
+
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+# $1: optional setup
+function doSetup() {
+  if doIsBatch
```
hohwille, line 19: same here, if `kubectrl` is available as command we can assume that nothing is left to do.

scripts/src/main/resources/scripts/command/kubernetes
```diff
@@ -0,0 +1,72 @@
...
+  else
+    echo "Sorry, kubernetes-installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
+    exit 0
+  fi
+}
+
+
+# Call Kubernetes with specified arguments
+function doRun() {
+  doEcho "Running: kubectl ${*}"
```
hohwille, line 52: first we need to call `doSetup` to ensure that it is already installed.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -0,0 +1,72 @@
...
+    wsl bash ./scripts/setup-docker
+  else
+    doEcho "Sorry, docker-installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
+    exit 0
+  fi
+}
+
+# Call Docker with specified arguments
+function doRun() {
+  doEcho "Running: docker ${*}"
```
hohwille, line 52: here we first should call `doSetup` so that if docker is not already installed, it should first be installed.

documentation/kubernetes.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+* As part of the setup, K3D will create a cluster with a single node with a default name as "devonfw-cluster"
+
+The arguments (`devon kubernetes «args»`) are explained by the following table:
+
+.Usage of `devon kubernetes`
+[options="header"]
+|=======================
+|*Argument(s)*             |*Meaning*
+|`setup`                   |setup Kubernetes (install and verify) as per above flow.
+|`«args»`                  |call kubernetes(kubectl) with the specified arguments. Call `kubectl help` for details or use kubectl directly as preferred." (`«args»`)
```
hohwille, line 29: is the last part left from some C&P?
```
" (`«args»`)
```
Looks missplaced to me.

scripts/src/main/resources/scripts/command/kubernetes
```diff
@@ -0,0 +1,72 @@
...
+  kubectl "${@}"
```
hohwille, line 53: I could never test on windows, but can this run after installation directly or does it have to run via `wsl bash`? If that runs directly this would be perfect and work same for every OS.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -0,0 +1,72 @@
...
+  docker "${@}"
```
hohwille, line 53: can this run directly (without `wsl bash`) on windows?

hohwille (commented):
@isandesh1986 thanks for your improvements. This is now perfectly following the philosophy of devonfw-ide and all the other existing commandlets.
Still I have left some last review comments. Also in some places indent is by 4 spaces and could be aligned. Will you be available to update PR once more? Otherwise let me know when I need to adapt the PR directly...

scripts/src/main/resources/scripts/command/kubernetes
hohwille: https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/mvn#L309

scripts/src/main/resources/scripts/command/docker
hohwille: Example:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/ionic#L19

scripts/src/main/resources/scripts/command/docker
```diff
@@ -0,0 +1,72 @@
...
+  then
+    doFail "Interactive installation is required, cannot proceed in batch mode. Please rerun without batch option."
+  fi
+  if doIsWindows
+  then 
+    wsl_distro_list=$(wsl -l 2>/dev/null)
+    if [[ $wsl_distro_list == *"no installed distributions"* ]];then
+      doEcho "No Linux distributions found in WSL or WSL is disabled. Please enable WSL and install any linux distribution to proceed with docker setup."
+      exit 0
+    fi
```
hohwille, line 29: 1. should we also echo a link to the devonfw-ide documentation with further instructions how to install WSL2 that we can even improve after the release?
2. we could move such "WSL" code to `functions` if we want to share it between `docker` and `kubernetes` commandlets.

scripts/src/main/resources/scripts/command/docker
isandesh1986: > @isandesh1986 thanks for your improvements. This is now perfectly following the philosophy of devonfw-ide and all the other existing commandlets.
> Still I have left some last review comments. Also in some places indent is by 4 spaces and could be aligned. Will you be available to update PR once more? Otherwise let me know when I need to adapt the PR directly...

Hi @hohwille  - I will update the PR and revert. Thanks

scripts/src/main/resources/scripts/command/docker
isandesh1986: Fixed it by prefixing wsl.

scripts/src/main/resources/scripts/command/kubernetes
isandesh1986: Fixed it by prefixing wsl

documentation/kubernetes.asciidoc
isandesh1986: Done. fixed it.

scripts/src/main/resources/scripts/command/kubernetes
isandesh1986: Hi @hohwille, For mvn it does make sense to install mvn when the user calls `mvn install`, but for docker and kubernetes I believe we should either redirect them to call `devon docker/kubernetes setup` rather than directly installing. WDYT?

Option1:- 
`Docker is not installed neither on Windows nor on WSL. Please call devon docker setup to install on WSL.`

Option2:-
`Docker is not installed neither on Windows nor on WSL.  Proceeding to install. Do you want to continue`

scripts/src/main/resources/scripts/command/kubernetes
```diff
@@ -0,0 +1,120 @@
...
+      echo "Sorry, kubernetes-installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
+      exit 0
+    fi
+  fi  
+}
+
+
+# Call Kubernetes with specified arguments
+function doRun() {
+  if doIsWindows
```
hohwille, line 55: why should we limit this to windows?
If `kubectl` is available on linux or MacOS we can make `devon kubernetes «args»` delegate to `kubectl «args»` without errors.
Why not simply reusing `doSetup` as we do in other commandlets? (we specify an argument to `doSetup` so we can distinguish if the user wants to setup explicitly or implicitly).

hohwille (approved):
@isandesh1986 thanks for your rework and completing this PR.
Now everything is perfect and all review comments are reworked.
I will merge and then add MacOS support.

PR #580, opened 2021-05-19
devonfw/devon4j#336: execute separate mvn command for batch

tobka777:
related to devonfw/devon4j#336

if -Dbatch exists execute archetype for batch


hohwille:
With PR https://github.com/devonfw/devon4j/pull/394 we replaced PR https://github.com/devonfw/devon4j/pull/392 and therefore do not need any change here in IDE.
This is great as otherwise we would introduce compatibility issues when ide and devon4j versions do not fit togehter.
I will therefore close this PR.

PR #582, opened 2021-05-21
#581: fixed redundant update-sites

### LINKED ISSUE #581: Redundant update sites not updated
We had to update two eclipse update-sites here:
https://github.com/devonfw/ide-settings/tree/master/eclipse/plugins

However, if no projects are configured in the settings git there is legacy support via redundant configs here:
https://github.com/devonfw/ide/blob/b092a6475cc7924501661214e89475d4997e9819/scripts/src/main/resources/scripts/command/eclipse#L171

As we can see for CobiGen we still have the old bintray URL that is not working anymore. See #567 
Also there was this fix:
https://github.com/devonfw/ide-settings/commit/8dbadc20bc1da1340e76c8af9ae5b25e84e8bff8#diff-8a11a5d42991fc938bd2470234d1f27698f7d114945e2bdd1ac5bee73ea534c2

These changes have to be reflected in the eclipse commandlet as well.

hohwille:
fixes #581

### DIFF
--- a/scripts/src/main/resources/scripts/command/eclipse
+++ b/scripts/src/main/resources/scripts/command/eclipse
@@ -165,10 +165,10 @@ function doAddPlugin() {
   then
     if [ "${1}" = "startexplorer" ]
     then
-      doInstallEclipsePlugin "de.bastiankrol.startexplorer.feature.feature.group" "http://basti1302.github.com/startexplorer/update/"
+      doInstallEclipsePlugin "de.bastiankrol.startexplorer.feature.feature.group" "http://basti1302.github.io/startexplorer/update/"
     elif [ "${1}" = "cobigen" ]
     then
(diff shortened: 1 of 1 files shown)


scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -165,10 +165,10 @@ function doAddPlugin() {
   then
     if [ "${1}" = "startexplorer" ]
     then
-      doInstallEclipsePlugin "de.bastiankrol.startexplorer.feature.feature.group" "http://basti1302.github.com/startexplorer/update/"
```
maybeec, line 168: couldn't we reuse here the properties instead of respecifying the hole stuff?
I am not getting why this code is here again whereas we have the same configuration in the plugin properties

hohwille: You need to understand the history of `devonfw-ide`. We need to care about downward compatibility.
The `plugins` folder with the properties files was later introduced. However, we still have users that never had and still do not have this folder in their settings git.
Removing this code will break their IDE installation.

PR #583, opened 2021-05-27
#573: improve completion

### LINKED ISSUE #573: improve autocompletion
With #287 we now have autocompletion for devonfw-ide what is really great and awesome.
However, I see the following room for improvements:
1. add missing options: Currently we have the following levels but only support completion for 0-2:
   * 0: devon
   * 1: commandlet
   * 2: commandlet-main function
   * 3: optional additional args or flags
2. optimize completion: Currently we have the following chain for autocompletion:
    * `source ~/.devon/autocomplete` from `.bashrc` or `.zshrc` (only once when shell is created)
    * `source ${DEVON_IDE_HOME}/scripts/devon-autocomplete` (on every completion when tab is hit)
    * get the completion either via `ls` or by calling commandlet with `shortlist` option (on every completion when tab is hit)

So this issue aims to improve 1. that I can also get level 3 things like:
`devon java migrate [TAB] from`

Furhter, my idea is that the second and third aspects from `2.` are generated at build or at least setup time:
So instead of calling chains for scripts, we could create an optimized completion function with all the options from our scripts. In other words the `shortlist` would be called to get the options but instead we directly put them into the autocompletion script that gets generated as we assume that once a release is installed, the autocompletion does not change anymore so no dynamic `ls` or invocation of `shortlist` will ever be needed.

If we go fully for KISS and ignore maintainability aspects, we could even manually hardcode the entire completion.
However, that would imply that whenever we add a new commandlet or option we redundantly have to change that completion script. This last point is still in discussion and yet to decide.

tobka777:
fixes #573

### DIFF
--- a/scripts/src/main/resources/scripts/command/eclipse
+++ b/scripts/src/main/resources/scripts/command/eclipse
@@ -6,6 +6,13 @@ then
   if [ -z "${2}" ]
   then
     echo "--all setup add-plugin run start ws-update ws-reverse ws-reverse-add create-script import help"
+  elif [ "${2}" = "--all" ]
+  then
+    shortlist3="setup add-plugin run start ws-update ws-reverse ws-reverse-add create-script import help"
+    if [ -z "${3}" ] || [[ ! " ${shortlist3} " == *" ${3} "* ]]
+    then
+      echo "${shortlist3}"
+    fi
   fi
   exit
 fi
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -9,15 +9,13 @@ then
   elif [ "${2}" = "update" ]
   then
     shortlist3="scripts settings software projects all"
-    shortlist3arr=("${shortlist3}")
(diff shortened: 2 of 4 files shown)


hohwille (approved):
@tobka777 thanks for this improvement. Looks good to me and can be merged. 👍 
However, just as a side-notice there are several other similar cases that also need to be addressed for autocompletion (`devon ide update scripts`, `devon ide update settings`, `devon ide update software`, `devon ide update scripts to 2021.04.001`, etc.).
Why is this PR still in draft state? If you want me to merge, simply take it out of draft...

tobka777:
I have now extended the code so that deeper autocompletion levels are possible and extended/adapted the support for java and ide.

hohwille (approved):
@tobka777 thanks for the PR. Looks good to me. 👍 
I had to read twice to understand how you are checking "contains" in the "whitspace-separate-lists" and that this also works for the first and last entry, but all looks good to me.
Thanks for the solution. I will merge.

Issue #584, opened 2021-06-01
add commandlets for docker and kubernetes


hohwille:
As a user of devonfw-ide, I want to have docker and kubernetes support so that I can automate this for my team and can easily write scripts that work cross-platform using docker and kubernetes.

## Acceptance

* `devon docker setup` will install `docker`. If docker is already installed on my machine, it will notice and honor the existing installation.
* `devon kubernetes setup` will install `kubernetes`. If kubernetes is already installed on my machine, it will notice and honor the existing installation.

hohwille:
Notice for MacOS:
* Installation: https://docs.docker.com/docker-for-mac/install/
* We could add support for MacOS that:
  * detects architecture (intel vs. apple CPU)
  * downloads the latest release for the detected architecture
  * opens the downloaded release (on MacOS that is a `dmg` file that is actually a virtual disc image that then gets mounted)
  * instruct the user with the required manual steps (go to the mounted disc drive and drag&drop the docker app to the applications folder, then launch docker from launchpad)

hohwille:
Easy to detect architecture on macos:
```
uname -m
x86_64
```

hohwille:
Was autoclosed by PR #570. I will reopen to add support for MacOS.

Any linux user of devonfw-ide available who could assist for linux support?
https://docs.docker.com/engine/install/ubuntu/

hohwille:
For windows network issues with VPN see:
https://github.com/sakai135/wsl-vpnkit

hohwille:
was autoclosed by PR #593 hence reopening as not yet done.

hohwille:
Docker Desktop is not available for Linux. However, I also found this one that supports all 3 OS:
https://dockstation.io/

Anybody experience with this tool?
I personally like Docker Desktop on MacOS and can recommend it. I have not yet used DockStation at all.

hohwille:
> Minimal system requirement:
Docker for Windows v1.12.0+
Docker Compose v1.6.0+ 
Windows 10 Pro (64bit)
Microsoft Hyper-V

Does not really seem to be state-of-the-art and what we are looking for. Maybe an optional aspect to close the Linux UI gap.

hohwille:
https://www.upnxtblog.com/index.php/2018/01/17/top-6-gui-tools-for-managing-docker-environments/

hohwille:
From PR #593 for Linux:

[...]

PR #593, opened 2021-07-08
#584: docker + k8s: improved windows and added mac support

hohwille:
For #584 (docker & kubernetes support) I improved Windows support and added MacOS support.


hohwille:
When I test this on windows it is still not working:
```
$ devon docker setup
Installing Docker on WSL...
DEVON_IDE_HOME variable set to /mnt/d/projects/test
Hint:You may be required to enter sudo user password of your Windows Subsystem for Linux(WSL) when prompted.
Installing Docker...

We trust you have received the usual lecture from the local System Administrator. It usually boils down to these three things:

   #1) Repsect the privacy of others.
   #2) Think before you type.
   #3) With great power comes great responsibility.


```

Then nothing happens. The process is stalled forever.
So IMHO all is still not working and we are not where we want to be.

hohwille:
> BTW: I have choosen Debian as Linux Distribution in case that matters.

LOL. Indeed this was the problem. Nowhere in the devonfw-ide documentation is stated which linux distribution is required.
In the provided URL from microsoft it says:
> Open the Microsoft Store and select your favorite Linux distribution.

However, I now digged in the code again and found this:
https://github.com/devonfw/ide/blob/0d89a96b5c1a268998e8932656363cb1a3b122b2/scripts/src/main/resources/scripts/setup-docker#L10

So actually the implementation only works for ubuntu.

I thought I had been very pickly already in my reviews. Now, I regret that we have already merged this incomplete feature onto `master`.
Remaining questions:

* Should we detect the linux distribution and support all major linux distros?
* Or if we only support ubuntu, we need to document this properly and also we need to verify this *AND* in case that a WSL2 user already has other linux installed alongside, we would need to find a way to ensure that we run `wsl` in ubuntu and not any of the other distros. So in the end this option does not seem any easier to me.

scripts/src/main/resources/scripts/functions
```diff
@@ -1155,32 +1162,19 @@ function doIsPackageJsonContainingScript() {
...
+  elif wsl -l > /dev/null
   then
     return
-  else   
-    return 255  
-  fi   
+  else
+    error="WSL 2 is disabled or no linux distributions found in WSL.\nPlease ensure WSL is enabled and a proper linux distribution is installed in WSL."
+  fi
+  doFail "${error}\nTo fix this follow these instructions:\nhttps://docs.microsoft.com/en-us/windows/wsl/install-win10"
```
SchettlerKoehler, line 1177: I think it would be better to use the short link here. The content will be displayed in the user's language and the link does not beak when MS restructures the docs. https://aka.ms/wsl2-install

scripts/src/main/resources/scripts/command/docker
```diff
@@ -17,84 +17,107 @@ source "$(dirname "${0}")"/../functions
...
+      fi
+    elif doIsMacOs
+    then
+      doRequireNotBatch
+      local downloadUrl
+      local arch
+      arch="$(uname -m)"
+      if [ "${arch}" = "x86_64" ]
+      then
+        downloadUrl="https://desktop.docker.com/mac/stable/amd64/Docker.dmg?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-mac-amd64"
```
SchettlerKoehler, line 54: I think you can just us https://desktop.docker.com/mac/stable/amd64/Docker.dmg
The parameters are just google analytics tracking parameters

scripts/src/main/resources/scripts/command/docker
```diff
@@ -17,84 +17,107 @@ source "$(dirname "${0}")"/../functions
...
-          fi
-          if [[ $docker_win_version != *"Server:"* ]];then
-            doEcho Please start Docker engine on WSL.
-          fi
-        else 
-          doEcho Docker not installed on WSL.
-          # Install Docker on WSL
-          wsl bash ../scripts/setup-docker
-        fi        
+        downloadUrl="https://desktop.docker.com/mac/stable/arm64/Docker.dmg?utm_source=docker&utm_medium=webreferral&utm_campaign=docs-driven-download-mac-arm64"
```
SchettlerKoehler, line 57: Same here

maybeec:
I feel we should support Debian and Ubuntu, nothing more. These are the most often used ones by users, especially windows users, who most often just know these distributions. For sure we can wait for contributions to support more if there are people who want to support more.
Nevertheless, I fully agree that we have to make it transparent. I would propose that we implement kind of a wsl adapter which checks for the current distribution installed. If a distribution is not supported, it should give a proper error message if trying to execute any command on wsl

isandesh1986:
> > BTW: I have choosen Debian as Linux Distribution in case that matters.
> 
> LOL. Indeed this was the problem. Nowhere in the devonfw-ide documentation is stated which linux distribution is required.
> In the provided URL from microsoft it says:
> 
> > Open the Microsoft Store and select your favorite Linux distribution.
> 
> However, I now digged in the code again and found this:
> https://github.com/devonfw/ide/blob/0d89a96b5c1a268998e8932656363cb1a3b122b2/scripts/src/main/resources/scripts/setup-docker#L10
> 
> So actually the implementation only works for ubuntu.
> 
> I thought I had been very pickly already in my reviews. Now, I regret that we have already merged this incomplete feature onto `master`.
> Remaining questions:
> 
> * Should we detect the linux distribution and support all major linux distros?
> * Or if we only support ubuntu, we need to document this properly and also we need to verify this _AND_ in case that a WSL2 user already has other linux installed alongside, we would need to find a way to ensure that we run `wsl` in ubuntu and not any of the other distros. So in the end this option does not seem any easier to me.

I think we should go with the 2nd one - mentioning in our documentation that we are supporting only these distros rather than not mentioning anything which gives the impression that it should work in any distro.

isandesh1986:
> > > BTW: I have choosen Debian as Linux Distribution in case that matters.
> > 
> > 
> > LOL. Indeed this was the problem. Nowhere in the devonfw-ide documentation is stated which linux distribution is required.
> > In the provided URL from microsoft it says:
> > > Open the Microsoft Store and select your favorite Linux distribution.
> > 
> > 
> > However, I now digged in the code again and found this:
> > https://github.com/devonfw/ide/blob/0d89a96b5c1a268998e8932656363cb1a3b122b2/scripts/src/main/resources/scripts/setup-docker#L10
> > 
> > So actually the implementation only works for ubuntu.
> > I thought I had been very pickly already in my reviews. Now, I regret that we have already merged this incomplete feature onto `master`.
> > Remaining questions:
> > 
> > * Should we detect the linux distribution and support all major linux distros?
> > * Or if we only support ubuntu, we need to document this properly and also we need to verify this _AND_ in case that a WSL2 user already has other linux installed alongside, we would need to find a way to ensure that we run `wsl` in ubuntu and not any of the other distros. So in the end this option does not seem any easier to me.
> 
> I think we should go with the 2nd one - mentioning in our documentation that we are supporting only these distros rather than not mentioning anything which gives the impression that it should work in any distro.

We used the instructions from one of the 1000kit guides
https://1000kit.gitlab.io/guides/docs/dev-environment/wsl2-pure/#docker

scripts/src/main/resources/scripts/command/docker
hohwille: Excellent feedback. 👍 GET parameters removed.

scripts/src/main/resources/scripts/command/docker
hohwille: done

hohwille:
I will evaluate Docker Desktop for Windows as I feel we are pointlessly reinventing the wheel here. Feedback will follow.

isandesh1986:
> I will evaluate Docker Desktop for Windows as I feel we are pointlessly reinventing the wheel here. Feedback will follow.

If DockerDesktop is installed on Windows - its anyways accessible from the WSL.

hohwille:
I tested the Docker Desktop installation on Windows:
* It works like a char
* It comes with a GUI installer that leads you through the required steps
* At the end you have WSL2 properly setup
* Also you get native docker and kubectl Commands that are directly available in your windows path and can be used in CMD and git-bash without prefixing via WSL.
* For Kubernetes support to work properly you once have to do an additional step: inside Docker Desktop go to the settings the choose Kubernetes and check "Enable Kubernetes". That is all.
* Docker Desktop gives the end user the same UX on all major OS: Linux, MacOS, and Windows.

Is there any rationale why we should not simply use Docker Desktop and reinvent the wheel?
My doubts are that with our handmade scripts it will take us > 1/2 year to get this production ready and work on all OS without users getting lost in space what is all already solved with Docker Desktop.

hohwille:
FYI: When I run `wsl -l` I now get 2 additional distros:
* docker-desktop-data
* docker-desktop

This does not interfere with any other WSL distro already installed and always uses the correct WSL distro.
All this is currently not functional with our current approach...

hohwille:
Also for the record: My test laptop has only 8GB of memory. With kubernetes enabled and all running it shows that 6 GB are used. That is significant but IMHO not much overhead from Docker Desktop at all.

scripts/src/main/resources/scripts/functions
hohwille: @SchettlerKoehler thanks for this feedback.
However, this PR is about the MacOS support so I will not include further windows changes here and better get this merged now.
For windows I will create another PR immediately after merge but there I will entirely remove this and go to Docker Desktop instead what makes that link to MS obsolete. In case we should decide against this PR and Docker Desktop for Windows we should do what you suggested. You can also edit the URL in the browser on github and create a PR from that with 2 clicks if you fear that this change might be forgotten or lost.

PR #599, opened 2021-09-03
spell check workflow added.

sujith-mn:


.spellcheck.yml
```diff
@@ -0,0 +1,24 @@
...
+- name: asciidoc source py spell check configuration
+  aspell:
+    lang: en
+    d: en_US
+  sources:
+  - documentation/*.asciidoc 
+  dictionary:
+    wordlists:
+    - en-devonfw.dict
+    output: solicitor.dic
```
hohwille, line 11: why is it called `solicitor.dic`? Is that a C&P?

hohwille (approved):
@sujith-mn thanks for your PR and adding this workflow.
Has this been tested in other repos?
As it does not run on the PR itself I have to merge rather blindly.
Anyway I am happy to do so and get the feature ready.
I also added a small review comment for you to check.

.spellcheck.yml
sujith-mn: yes ,, it is a C&P issue

PR #608, opened 2021-10-01
spellcheck & sync-docs on change of docs only

sujith-mn:


SchettlerKoehler (approved):

.github/workflows/spellcheck.yml
```diff
@@ -1,6 +1,9 @@
 name: spellcheck 
 on:
   push:
+    paths:
+       - '**.asciidoc'
```
SchettlerKoehler, line 5: Should we also add adoc?

.github/workflows/sync-docs.yml
```diff
@@ -7,6 +7,9 @@ on:
   push:
     branches:
       - master
+    paths:
+      - '**.asciidoc'
```
SchettlerKoehler, line 11: Should we also add adoc?

SchettlerKoehler (changes requested):

.github/workflows/spellcheck.yml
maybeec: I think there is a convention in docgen to use asciidoc only.

maybeec (approved):

PR #612, opened 2021-10-07
Update/spellcheck

prathibhapadma:


hohwille:
What is the purpose of this PR? I do not see any relation to the subject and no description is provided.
I will therefore close it. Please only reopen after a reasonable description is given.

PR #625, opened 2021-11-16
There is no doIsQuiet yet for CobiGen

maybeec:
as it is done with --verbose it does not make any sense and it's not working as expected


hohwille (approved):
@maybeec thanks for this PR 👍 
If `--verbose` is not supported or does not make sense (to have by default) this change is perfectly fine. Writing output to `/dev/null` for quite also does not make sense.
So I approve and merge.

Issue #641, opened 2021-12-02
Improve tools download to be stable


hohwille:
So here is our plan how we want to fix issues like #626 on the long run:

* we create a new github repo `ide-mirrors`
* in that repo we create a folder for each supported tool (e.g. `eclipse`, `java`, `mvn`)
* inside such folder we create a file `default` that contains download mirror base URLs (one URL per line)
* for each version of the tool that can not be downloaded from the `default` mirror list, we create a separate file named exactly like the version of the tool (e.g. `2021-03` or `11.0.9.1_1`) that contains the alternative mirror URLs
* when we setup devonfw-ide, we clone this new `ide-mirrors` in addition to the settings. However here we always clone the original repo maintained by us from github.
* we change the `doDownload` function to also take additional parameters
  * the name of the tool
  * the version of the tool
  * the URL path of the download within the mirror (e.g. `openjdk${major_version}-binaries/releases/download/${jdk_folder}/OpenJDK${major_version}U-jdk_x64_${os}_hotspot_${software_version}${extension}`) or alternatively (maybe even better) we use variables in the URLs on the `ide-mirrors` that are later resolved (but beside standard variables like `os` and `version` we would then need to provide variables from the commandlet to `doDownload` that then needs to resolve those variables)
* the `doDownload` function is changed such that it
  * performs a git pull on the `ide-mirrors` local clone
  * it looks if the specified tool exists as folder in the cloned `ide-mirrors`
  * if that folder exists it looks if the version of the tool exists as file in that folder, otherwise it takes `default` as file
  * if it has such a file of URLs that exists and is not empty, the download will be perfomed using this URL (appending the path or resolving variables).
  * we read the file line-by-line into a list/array
  * as a nice-to-have we can bring that list/array into a random order (to distribute load accross mirrors)
  * we then download the tool starting with the first URL from the list/array, if that failes for some reason, we try the next from the list
  * if all fails or no URLs file was ever present, etc. we fall back to the old approach to download directly from the URL specified by the commandlet (this also allows commandlets for tools that are not mainted by us in `ide-mirrors` - e.g. if a project adds its own commandlet and maybe even uses internal URLs for download that shall not be published on github).

[...]

PR #643, opened 2021-12-11
#641: use mirrors for managed tools

hohwille:
Implemented #641

That was a long day/night...

### DIFF
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -346,47 +346,140 @@ function doDevonCommandAndReturn() {
 }
 
 # $1: the URL to download
-# $2: the optional, explicit filename to save to
-# $3: the optional, target directory to save to
+# $2: the optional, target directory to save to
+# $3: the software to download (e.g. 'java', 'mvn', 'node')
+# $4: the version of the software to download
+# $5: operating system name
+# $6: the optional architecture (e.g. x64 or x86_64)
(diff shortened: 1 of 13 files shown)


scripts/src/main/resources/scripts/functions
```diff
@@ -346,47 +346,140 @@ function doDevonCommandAndReturn() {
...
+    fi
+    local mirrors_file="${mirrors_dir}/${software}/${version}"
+    if [ ! -e "${mirrors_file}" ]
+    then
+      doDebug "Mirror file not found for version at ${mirrors_file}"
+      mirrors_file="${mirrors_dir}/${software}/default"
+    fi
+    if [ ! -e "${mirrors_file}" ]
+    then
+      doFail "No mirror is configured at ${mirrors_file}"
```
maybeec, line 406: This is a breaking change as people would have to migrate their settings towards the new design, which is not intuitive at all.
So enable this change (which I really appreciate), we need to provide automated migration capabilities and notify the user to properly sync his / her settings after automatic migration.
Do you also updated the ide settings accordingly such that we can merge both PRs together only?

maybeec (changes requested):

scripts/src/main/resources/scripts/functions
maybeec: Clarified on my own. Looks great. Thanks for the effort.

maybeec (approved):

PR #647, opened 2021-12-16
#641: improve mirrors

hohwille:
implements #641 and addes the suggested improvements after PR #643 was already merged.
I also found several bugs that I fixed alongside.
For the record: This PR MUST be merged before creating a release as I had to also merge https://github.com/devonfw/ide-mirrors/pull/1 so this PR can build green.


hohwille:
For some strange reasons I got conflicted with myself (maybe due to odd merges).
Resolving conflicts in the browser is not much fun.
I will double check by compaing my original local FB with this one after merge.

hohwille:
> For some strange reasons I got conflicted with myself (maybe due to odd merges). Resolving conflicts in the browser is not much fun. I will double check by compaing my original local FB with this one after merge.

OK. Verified. Only two new lines have been added due to conflict resolution but otherwise `functions` file is identical. Other changes are also fine but small and easy to review.

maybeec (approved):

hohwille:
Build failed with error:
```
*** Setting up mvn ***

******** ATTENTION ********
ERROR: Download of java in version 11.0.9.1_1 failed after trying 1 mirror(s).
```

Ouch. There are `doEcho` statements that should print some messages before this error actually occurrs.
I have to find a way to reproduce this as it did not happen on my machine. May the conflict resolution broke something that I missed.

hohwille:
I will still address the `quiet` vs. `verbose` issue with dependent setups.
FYI: The tests have not been fixed by my last commit (code style) but by this commit https://github.com/devonfw/ide-mirrors/commit/b8059d09f8b8ceee30900c3f31852304e0f8162a

hohwille:
I have tested this with arbitrary variants and combinations (and found and fixed bugs I  made temporary). It is quite astounding how many combinations of tools, versions, operating-systems, etc. we do support.
I will merge this now and would be very happy if some people could then get a latest SNAPSHOT from OSSRH nexus (https://github.com/devonfw/ide/blob/master/documentation/setup.asciidoc#testing-snapshot-releases).

Issue #631, opened 2021-11-19
terminal control codes aren't processed when using Cobigen-cli with devon prefix


LarsReinken:
### Expected behavior

As a cobigen user, I want to use cobigen-cli.

### Actual behavior
![image](https://user-images.githubusercontent.com/9500956/142628584-cd9fd2f7-5ec5-4a50-ab8a-bcbf667d6fd6.png)

### Steps to reproduce (bug) / Use Case of feature request (enhancement)
1. install devonfw ide
2. Open a devon cmd using your context menu inside your project
2. Run: "devon cobigen" to install CobiGen CLI
3. Run "devon cobigen generate"

### Related/Dependent Issues
https://github.com/devonfw/cobigen/issues/1404

### Comments/Hints:
after updating the path by running "devon". Using just cobigen doesn't produce these characters, neither is the output coloured.
using git bash instead of devon-cmd works fine.

### Affected version:
release 2021.04.002

hohwille:
Thanks for the feedback. However, I am quite clueless how to solve this:
We are piping the output of cobigen-cli in bash that itself (in case of the error) is processed by a windows CMD terminal that is unable to understand the color control output produced by cobigen.
Maybe we could try to detect if we are inside a CMD terminal and in such case try to pass a parameter to cobigen disabling colors. However, if so then is devonfw-ide in charge at all here or is this a bug of cobigen?
Does it work when cobigen is invoked from CMD without being prefixed by `devon` script?

hohwille:
I just tested this without `devon` as wrapper command and get the same output.
So if you consider this as a bug it is a bug of CobiGen or MSDos/CMD but not devonfw-ide and cannot fix it here.

PR #655, opened 2021-12-21
added colored output

maybeec:
warnings are outputted yellow
steps cyan
sucess messages green
error messages red


scripts/src/main/resources/scripts/command/ide
```diff
@@ -318,8 +318,9 @@ function doUpdateSoftware() {
   local tool
   for tool in "${DEVON_IDE_TOOLS[@]}"
   do
-    doEcho "\n*** Setting up ${tool} ***"
+    doEcho -e "\e[96m\n*** Setting up ${tool} ***\e[39m"
```
hohwille, line 321: Did you do search&replace of `echo` case insensitive?
`-e` is wrong here for `doEcho`.
See here: 
https://github.com/devonfw/ide/blob/d462bc6ae93075f613081d710cd88f1490fd1ed8/scripts/src/main/resources/scripts/functions#L149

scripts/src/main/resources/scripts/command/ide
```diff
@@ -318,8 +318,9 @@ function doUpdateSoftware() {
...
     doDevonCommand "${tool}" setup
+    doEcho -e "\e[92mSuccess\e[39m"
```
hohwille, line 323: same here

scripts/src/main/resources/scripts/command/ide
```diff
@@ -331,7 +332,7 @@ function doUpdateSoftware() {
       software_version="${tool%%:*}"
       if [ -n "${software_version}" ]
       then
-        doEcho "\n*** Setting up ${software} ***"
+        doEcho -e "\e[96m\n*** Setting up ${software} ***\e[39m"
```
hohwille, line 335: same here

scripts/src/main/resources/scripts/command/ide
hohwille: Please note there is also `doWarning` and `doDebug`, etc.
We could add the color automatically for `doWarning` what would make much more sense then hardcoding this strange escape syntax redundantly everywhere...

scripts/src/main/resources/scripts/devon
```diff
@@ -57,7 +57,7 @@ do
   fi
   if [ "${PWD}" = "/" ]
   then
-    echo "You are not inside a devonfw-ide installation: ${L_PWD}"
+    echo -e "\e[93mYou are not inside a devonfw-ide installation: ${L_PWD}\e[39m"
```
hohwille, line 60: maybe use `doWarning`?

scripts/src/main/resources/scripts/devon
```diff
@@ -142,7 +142,7 @@ then
   fi
   if ! [ -d "${DEVON_IDE_HOME}" ]
   then
-    echo "ERROR: DEVON_IDE_HOME does not exist!"
+    echo -e "\e[91mERROR: DEVON_IDE_HOME does not exist!\e[39m"
```
hohwille, line 145: we should use `doWarning` or add `doError`

scripts/src/main/resources/scripts/devon
```diff
@@ -168,7 +168,7 @@ then
       result=${?}
     fi
   else
-    echo "Unknown command ${command}"
+    echo -e "\e[93mUnknown command ${command}\e[39m"
```
hohwille, line 171: use `doWarning` instead.

scripts/src/main/resources/scripts/functions
```diff
@@ -19,27 +19,27 @@ DEVON_DOWNLOAD_DIR="${DEVON_HOME_DIR}/Downloads/devonfw-ide"
 # $2: optional exit code
 function doFail() {
   doEchoAttention ""
-  echo -e "ERROR: ${1}"
-  echo "We are sorry for the inconvenience. Please check the above errors, resolve them and try again."
+  echo -e "\e[91mERROR: ${1}\e[39m"
+  echo -e "\e[91mWe are sorry for the inconvenience. Please check the above errors, resolve them and try again.\e[39m"
```
hohwille, line 23: we could simplify by joining both to a single echo with `\n` separator.

scripts/src/main/resources/scripts/functions
```diff
@@ -334,7 +334,7 @@ function doDevonCommandAndReturn() {
   else
     if [ ! -x "${command}" ]
     then
-      doEcho "Command ${1} is not executable. Trying to repair..."
+      doEcho -e "\e[93mCommand ${1} is not executable. Trying to repair...\e[39m"
```
hohwille, line 337: use `doWarning`

scripts/src/main/resources/scripts/functions
```diff
@@ -343,7 +343,7 @@ function doDevonCommandAndReturn() {
     result=${?}
     if [ ${result} != 0 ]
     then
-      doEcho "ERROR: command '${command_name} ${*}' failed with exit code ${result}"
+      doEcho -e "\e[91mERROR: command '${command_name} ${*}' failed with exit code ${result}\e[39m"
```
hohwille, line 346: same here

scripts/src/main/resources/scripts/functions
```diff
@@ -576,11 +576,11 @@ function doDownloadInternal() {
   local result=${?}
   if [ "${result}" != 0 ]
   then
-    doEcho "Failed to download ${url} with exit code ${result}"
+    doEcho -e "\e[91mFailed to download ${url} with exit code ${result}\e[39m"
```
hohwille, line 586: same here

scripts/src/main/resources/scripts/functions
```diff
@@ -576,11 +576,11 @@ function doDownloadInternal() {
...
     return ${result}
   else
     mv "${tmp_file}" "${target}"
-    doDebug "Download of ${DOWNLOAD_FILENAME} from ${url} succeeded."
+    doDebug -e "\e[32mDownload of ${DOWNLOAD_FILENAME} from ${url} succeeded.\e[39m"
```
hohwille, line 590: same here

hohwille (changes requested):
@maybeec thanks for your PR.
It is a great improvement if we get the coloring.
However as already discussed in #631 the colors lead to strange output and confusion in CMD that also needs to be supported by devonfw-ide on windows.
IMHO we should first get this solved or find a way to determine that case and prevent color escape codes in CMD.
Besides you created several bugs that I commented.
As I am short on time, I would pick up this issue next year and create a new release with this great feature in January 2022 unless you have to time to get everything solves yourself immediately...

maybeec:
@hohwille at least in my tests I could not see any strange output.
I will have a look at the ticket you mentioned but I think it's an issue of the Java generated console output.

@jan-vcapgemini possibly you can give a try and check whether you get strange output on cmd

maybeec:
Any news on that? I feel it's ready to be tested. I couldnt made some more Bugfixing but would like to get other opinions as well. Testing welcome

maybeec:
@hohwille merge conflicts start to raise as the PRs are quite long open. Is there anything I can do in addition you would like to get changed?

hohwille:
I tested `devon mvn install` in CMD shell and it gives color output just as with calling `mvn install` directly.
So it is definitely not a problem or bug of a `bash` being in between CMD and the outputting process.
I can take over the conflict resolution here and retest the latest code.
If all works out, I can merge this.
Thanks for all your work and great that I can be back after being M.I.A.

hohwille:
@maybeec thanks for your great work. 👍 
It works fine beside some small bugs and things to improve that I will take over.
Replacing this PR with PR #778

Issue #667, opened 2022-01-05
devon npm setup ignores a given version


fdcg:
### Expected behavior

As a user, I want to define the variable `NPM_VERSION` as described in the documentation so that this defined version is installed with `devon npm setup`.

### Actual behavior
The variable `NPM_VERSION` is ignored and there is only the NPM version installed with node available. 

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1.  Install node with `devon node setup`
2.  Call `npm -version` and notice the version
3.  Set the variable `NPM_VERSION` to a different version.
4.  Call `devon npm setup`
5.  Call  `npm -version`

The output of step 5 is the same as of step 2.

### Affected version:
* Devon scripts: 2020.08.002
* OS: Windows
* Browser: independent

hohwille:
npm command itself just delegates to node:
https://github.com/devonfw/ide/blob/1b535ea2c19881a4795e93569e3452c861742df9/scripts/src/main/resources/scripts/command/npm#L21

Seems that someone thought an implicit requirement is to get latest npm insode node command:
https://github.com/devonfw/ide/blob/1b535ea2c19881a4795e93569e3452c861742df9/scripts/src/main/resources/scripts/command/node#L39

IMHO I did not design this part.
Can someone still shed a light into this?

IMHO the expecation of @fdcg is correct and building on this I would say:
1. node command should not install or update npm at all
2. only npm command should do that
3. npm command should honor `${NPM_VERSION}` variable to be consistent with all other commandlets. Only if that variable is undefined it should fallback to `latest`.

hohwille:
@fdcg It seems that npm version is somehow bundled with node version. So the safest and easiest approach would be to maintain `NODE_VERSION` in your project that in the end will both manage `node` and `npm` for your project as these two bundled together are guaranteed to work in together fine.

Do you happen to need a specific `npm` version AND a speciifc `node` version that do not come together?
We could easily add some logic to `npm` commandlet that checks for `NPM_VERSION` and if present does the exact installation of that `npm` version regardless of the `node` version used. However, this approach could easily end in a setup that is broken.

hohwille:
I created a simple fix with PR #672
@fdcg Is this solving your needs?

PR #668, opened 2022-01-05
Bugfix for issue #667 - devon npm setup ignores a given version

fdcg:


scripts/src/main/resources/scripts/command/npm
```diff
@@ -16,13 +16,47 @@ source "$(dirname "${0}")"/../functions
...
+  if [ -z "$software_version" ] || [ "$installed_version" == "$software_version" ] 
+  then
+    if [ "${1}" != "silent" ]
+    then
+      doEcho "Version ${installed_version} of npm is already installed at ${NPM_CMD}"
+    fi
+  else
+    local download_url="https://registry.npmjs.org/npm/-/npm-${software_version}.tgz"
+    local software_dir="${NODE_HOME}/node_modules/npm"
+    doInstall "${software_dir}" "${download_url}" "npm" "${software_version}"
```
hohwille: this will not work as you have developed this PR for an older version of devonfw-ide and the function and download process has been improved and refactored. See: #641 
So we should to put the URL into https://github.com/devonfw/ide-mirrors/ and change the arguments of `doInstall` (see other commandlets that have been refactored already).

hohwille: As a quickfix, I changed the order of the parameters so it should already work, but we want to keep the download URLs out of our release as explained in #641

scripts/src/main/resources/scripts/command/npm
```diff
@@ -16,13 +16,47 @@ source "$(dirname "${0}")"/../functions
...
+      rm "${NODE_HOME}/npx"
+    fi    
+    if [ -f "${NODE_HOME}/npx.cmd" ]
+    then 
+      rm "${NODE_HOME}/npx.cmd"
+    fi
+    cp "${software_dir}/bin/npm" "${NODE_HOME}"
+    cp "${software_dir}/bin/npm.cmd" "${NODE_HOME}"
+    cp "${software_dir}/bin/npx" "${NODE_HOME}"
+    cp "${software_dir}/bin/npx.cmd" "${NODE_HOME}"
```
hohwille, line 58: Instead of removing and copying, I would prefer to create symlinks but due to windows limitations this does not seem to be an option. So I think there is no better solution than yours here that should always work (KISS).

hohwille (commented):
@fdcg thanks for this great PR 👍 
I did some minimal changes directly as constructive review.
For the download we should follow the new approach as commented in the review below.
Should I take over or would have have the change to rework?

scripts/src/main/resources/scripts/command/npm
hohwille: I have already added the according config to the ide-mirrors repo:
https://github.com/devonfw/ide-mirrors/tree/master/npm

hohwille: so we could stip out the download_url (line 36) and change to:
```
doInstall "-" "${software_dir}" "npm" "${software_version}"
```

scripts/src/main/resources/scripts/command/npm
hohwille: For the record: the current approach could fail on Windows if a `npm.cmd` is currently running in parallel due to file locking.

hohwille:
@fdcg awesome. Thanks for updating this PR and testing on win.
Ready to be merged.

hohwille (approved):

Issue #638, opened 2021-11-24
GraalVM integration


SchettlerKoehler:
As a devon IDE user, I want to have GraalVM available in the devon IDE so that I and my team can easily ramp up our development environment and we are using the same and correct versions of the tool.

Acceptance criteria:

- [x] There is a new commandlet that allows to install and launch GraalVM from the CLI
- [x] The new commandlet is a wrapper for GraalVM CLI
- [x] GraalVM can be installed automatically (during setup via settings) as well as via the commandlet call
- [x] During the installation process I get information about the current status of the installation
- [x] The new commandlet is documented on the website
- [x] Other tools are able to find GraalVM (GRAALVM_HOME is set)

Sprint 2 ACs:
- [x] GraalVM integration is wokring with Windows.
- [x]  GraalVM can be installed and used together with any of the devon IDE standard tools without negative interference - also tested
- [x] GraalVM is integrated in a way that complies with the multi-tenancy of the devon IDE(independently of the used system (mac, linux, windows).

Sprint 3 ACs:
- [x] Add licenec documentation

hohwille:
A problem with GraalVM is that it contains commands like `java` in the bin folder. Adding both Java and GraalVM to your path will cause chaos.
It is fine to just install it to `software/extra/graalvm` so it is not automatically added to the path.
However, then we can only do the following:
* export some variable `GRAALVM_HOME` - actually best if the user does this via his `devon.properties` manually?
* provide `graalvm` commandlet as wrapper script delegating to GraalVM (e.g. `devon graalvm java ...`).

Bene90:
@tobka777 Please keep in mind that the lincence documentation must be mentioned.
@hohwille What is the right location for it? (Also relevant for Rancher Desktop, HELM, Terraform)

tobka777:
@Bene90 I did this already! 
See Commit https://github.com/devonfw/ide/pull/698/commits/f892a28190d8c58310f189985661866b2326363d and https://github.com/devonfw/ide/pull/698/commits/6e90e9664f3360508c94af2d8c8294ee20c813da

PR #682, opened 2022-01-11
#678: prevent to add extra to path

hohwille:
Improvement for #678: do not add `extra` to path.

Side-note: On MacOS for Java a workaround is needed:
* java follows MacOS app conventions but not standards of other OS
* that is has a folder `Contents` that contains subfolders `Home` and `MacOS`
* the expected layout of a JDK is within `Contents/Home` that also contains the `bin` folder that contains `java`, etc.
* However, the entire structure is required for the JDK to work
* as a workaround devonfw-ide detects this and then renames the `java` folder to `jdk` and creates a symlink `java` pointing to `jdk/Contents/Home` so everthing else works as expected.
* as a result also the `jdk` folder is put on the `PATH`. This does not cause real harm but it should also be prevented.
* Further we have the same issue also with other apps on MacOS. For Eclipse and IntelliJ we simply create launch scripts ourselves at the expected locations but this only works for "single command tools". For things like GraalVM (see #638) we will have the same problems as with Java.
* Therefore I am thinking about moving this workaround from the `java` commandlet to the `doInstall` function and make it generic. Maybe instead of renaming the folder inside `software` it could be moved to a generic subfolder (something like the new `extra` folder in `software` but to avoid conflicts it needs to be a different one - maybe just `macos` or if we expect similar workarounds also for other OS something like `internal` also excluded from the path.

### DIFF
--- a/scripts/src/main/resources/scripts/environment-project
+++ b/scripts/src/main/resources/scripts/environment-project
@@ -129,17 +129,20 @@ fi
 SOFTWARE_PATH=${DEVON_IDE_HOME}/software
 for SOFTWARE_FOLDER in "${SOFTWARE_PATH}"/*
 do
-  if [ -d "${SOFTWARE_FOLDER}/bin" ]
+  if [ "${SOFTWARE_FOLDER}" != "${SOFTWARE_PATH}/extra" ]
   then
-    PATH="${SOFTWARE_FOLDER}/bin:${PATH}"
-  else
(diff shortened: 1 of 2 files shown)


hohwille:
QA note: I will do proper test on windows after merging this PR when the SNAPSHOT is build.

hohwille:
Manually tested on windows - works like a charm.

Issue #693, opened 2022-01-24
Docker isQuiet: command not found


ewap85:
Expected behavior
As a devonfw-ide user,  run devon docker setup without errors.

Actual behavior
scripts/command/docker: line 79: isQuiet: command not found
Steps to reproduce (bug) / Use Case of feature request (enhancement)

The same bug like:
#664 / 662: scripts/command/eclipse: line 79: isQuiet: command not found

 if [ "${1}" != "silent" ] && ! isQuiet 
has to be doIsQuiet instead.

PR #694, opened 2022-01-25
#693 bugfix in docker file isQuiet

ewasikora:


hohwille (approved):
OMG - another stupid bug like this ;)
@ewasikora excellent finding. Thanks for the fix 👍

Issue #696, opened 2022-02-01
Move CobiGen home folder from user home to ide per default


jan-vcapgemini:
As a CobiGen user, I want to keep my CobiGen home folder inside the ide so that it does not conflict with my home folder.

Solution:
Add COBIGEN_HOME to devon.properties

PR #697, opened 2022-02-01
#696 added new COBIGEN_HOME environment variable to devon.properties

jan-vcapgemini:
Adresses: https://github.com/devonfw/cobigen/issues/1430 and implements #696


hohwille (approved):
@jan-vcapgemini thanks for this PR and fix 👍 
Does this work when the folder does not yet exists and will cobigen create it?
It slighly conflicts with requirements from issue #5 but as we have a bug issue fixed this way, So I am happy to approve and merge. Users who do not use cobigen can still unset the property in the own `devon.properties` if they have problems with this while it is more important that things work easily OOTB.

Side-node: Typically env vars like `<Tool>_HOME` point to the software installation folder of that tool rather than a config folder. However, this is more a question of cobigen than devonfw-ide and most likely a choice already made that can not be changed easily.

Issue #683, opened 2022-01-11
Generalize MacOS app workaround


hohwille:
From PR #682:

On MacOS for Java a workaround is needed:

* java follows MacOS app conventions but not standards of other OS
* that is has a folder Contents that contains subfolders Home and MacOS
* the expected layout of a JDK is within Contents/Home that also contains the bin folder that contains java, etc.
* However, the entire structure is required for the JDK to work
* as a workaround devonfw-ide detects this and then renames the java folder to jdk and creates a symlink java pointing to jdk/Contents/Home so everthing else works as expected.
* as a result also the jdk folder is put on the PATH. This does not cause real harm but it should also be prevented.

We have the same issue also with other apps on MacOS. For Eclipse and IntelliJ we simply create launch scripts ourselves at the expected locations but this only works for "single command tools". For things like GraalVM (see GraalVM integration #638) we will have the same problems as with Java.

Therefore I am thinking about moving this workaround from the java commandlet to the doInstall function and make it generic. Maybe instead of renaming the folder inside software it could be moved to a generic subfolder: something like the new extra folder in software but to avoid conflicts it needs to be a different one - maybe just `macos` or if we expect similar workarounds also for other OS something like `internal` also excluded from the path.

When we implement this we have to consider existing installations. Ideally the old `software/jdk` folder will be deleted/archived if `devon java setup` is executed.

tobka777:
I implemented a general workaround for java and graalvm.
For Eclipse, Intellij, VSCode we have to look for another solution. Now they use a script to pass the arguments.

PR #698, opened 2022-02-01
GraalVM integration

### LINKED ISSUE #638: GraalVM integration

tobka777:
fixes #638

### DIFF
--- a/scripts/src/main/resources/scripts/command/graalvm
+++ b/scripts/src/main/resources/scripts/command/graalvm
@@ -44,6 +44,7 @@ function doSetup() {
     fi
     doInstall "-" "${installDir}" "graalvm" "${version}"
     echo -e "export GRAALVM_HOME=${GRAALVM_HOME}" >> "${DEVON_IDE_HOME}/conf/devon.properties"
+    doExtendPath ${GRAALVM_HOME}
   fi
   if [ "${1}" != "silent" ] && ! doIsQuiet
   then
@@ -54,7 +55,7 @@ function doSetup() {
     fi
     doRunCommand "'${java_binary}' -version" "verify installation of GraalVM"
   fi
-  export GRAALVM_HOME="$GRAALVM_HOME"
+  export GRAALVM_HOME
 }
 
 # Call GraalVM with specified arguments
--- a/scripts/src/main/resources/scripts/command/graalvm
+++ b/scripts/src/main/resources/scripts/command/graalvm
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+
+# autocompletion list
+if [ "${1}" = "shortlist" ]
+then
+  if [ -z "${2}" ]
+  then
+    echo "setup help"
+  fi
+  exit
+fi
+
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+# $1: optional setup
+function doSetup() {
+  GRAALVM_HOME="${DEVON_IDE_HOME}/software/extra/graalvm"
+  doDevonCommand java setup silent
+  if [ ! -d "${GRAALVM_HOME}" ]
+  then
+    local version="${GRAAL_VERSION:-22.0.0.2}"
+    local javaVersion="${EXTRA_JAVA_VERSION}"
+    if [ -z "${javaVersion}" ]
+    then
+      javaVersion="${JAVA_VERSION}"
+    fi
+    if [ -n "${javaVersion}" ]
+    then
+      doVersionCompare "${javaVersion}" "11u0"
+      if [ "${?}" == 2 ]
+      then
+        # Java version is lower than 11
+        doConfirmWarning "GraalVM requires Java 11 or newer.\nHowever, [EXTRA_]JAVA_VERSION is ${javaVersion} - in this setup GraalVM can not work.\nPlease update JAVA_VERSION, set EXTRA_JAVA_VERSION or downgrade ECLIPSE_VERSION in your settings/devon.properties!"
+      fi
+    fi
+    doInstall "-" "${GRAALVM_HOME}" "graalvm" "${version}"
+  fi
+  if doIsMacOs
+  then
+    GRAALVM_HOME="${DEVON_IDE_HOME}/software/extra/graalvm/Contents/Home"
+  fi
+  if [ "${1}" != "silent" ] && ! doIsQuiet
+  then
+    doRunCommand "'${GRAALVM_HOME}/bin/java' -version" "verify installation of GraalVM"
+  fi
+}
+
+# Call GraalVM with specified arguments
+function doRun() {
+  doSetup silent
+  ${GRAALVM_HOME}/bin/${@}
+}
+
+# CLI
+case ${1} in 
+"help" | "-h")
+  echo "Setup GraalVM."
+  echo
+  echo "Arguments:"
+  echo " setup                    setup GraalVM (install and verify)"
+  echo " «args»                   call GraalVM with the specified arguments"
+  echo  
+;;
+"setup" | "s" | "")
+  doSetup "${2}"
+;;
+*)
+  doRun "${@}"
+;;
+esac
--- a/scripts/src/main/resources/scripts/command/graalvm
+++ b/scripts/src/main/resources/scripts/command/graalvm
@@ -65,26 +65,21 @@ function doSetup() {
 # Call GraalVM with specified arguments
 function doRun() {
   doSetup silent
-  command="$GRAALVM_HOME/bin/$*"
-  local binary=${command%% *}
-  local subcommand=${command/"$binary"/}
-  if [ -f "${binary}" ]
+  local binary="${GRAALVM_HOME}/bin/${1}"
+  shift
+  if [ ! -x "${binary}" ]
   then
-    binary=${binary}
-  elif [ -f "${binary}.exe" ]
-  then
-    binary=${binary}".exe"
-  else
-    binary=${binary}".cmd"
+    if [ -f "${binary}.exe" ]
+    then
+      binary=${binary}".exe"
+    fi
+    if [ -f "${binary}.cmd" ]
+    then
+      binary=${binary}".cmd"
+    fi
   fi
 
-  if [ -f "${binary}" ]
-  then
-    # shellcheck disable=SC2086
-    "${binary}" ${subcommand}
-  else
-    doFail "Command not found"
-  fi
+  doRunCommand "${binary} ${*}"
 }
 
 # CLI
--- a/scripts/src/main/resources/scripts/command/graalvm
+++ b/scripts/src/main/resources/scripts/command/graalvm
@@ -51,7 +51,13 @@ function doSetup() {
 # Call GraalVM with specified arguments
 function doRun() {
   doSetup silent
-  "${GRAALVM_HOME}"/bin/"${@}"
+  command="${GRAALVM_HOME}/bin/${@}"
+  if [ -f $( echo "$command" | cut -d ' ' -f1 ) ]
+  then
+    ${command}
+  else
+    doFail "Command not found"
+  fi
 }
 
 # CLI
--- a/scripts/src/main/resources/scripts/command/graalvm
+++ b/scripts/src/main/resources/scripts/command/graalvm
@@ -32,7 +32,7 @@ function doSetup() {
       if [ "${?}" == 2 ]
       then
         # Java version is lower than 11
-        doConfirmWarning "GraalVM requires Java 11 or newer.\nHowever, [EXTRA_]JAVA_VERSION is ${javaVersion} - in this setup GraalVM can not work.\nPlease update JAVA_VERSION, set EXTRA_JAVA_VERSION or downgrade ECLIPSE_VERSION in your settings/devon.properties!"
+        doConfirmWarning "GraalVM requires Java 11 or newer.\nHowever, [EXTRA_]JAVA_VERSION is ${javaVersion} - in this setup GraalVM can not work.\nPlease update JAVA_VERSION or set EXTRA_JAVA_VERSION in your settings/devon.properties!"
       fi
     fi
     doInstall "-" "${GRAALVM_HOME}" "graalvm" "${version}"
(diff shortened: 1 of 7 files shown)


tobka777:
Ready version for macOS done. Tests for Windows and Linux are needed.

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,79 @@
...
+source "$(dirname "${0}")"/../functions
+
+# $1: optional setup
+function doSetup() {
+  GRAALVM_HOME="${DEVON_IDE_HOME}/software/extra/graalvm"
+  doDevonCommand java setup silent
+  if [ ! -d "${GRAALVM_HOME}" ]
+  then
+    local version="${GRAAL_VERSION:-22.0.0.2}"
+    local javaVersion="${EXTRA_JAVA_VERSION}"
```
hohwille, line 24: Great that you look into other commandlets and apply existing solutions...
So does GraalVM really require a JDK (with a version >= 11)? I was not fully aware of this. Do you have a link for this fact? 
I only found this what seems to cause other issues we might have to cover:
https://www.graalvm.org/22.0/reference-manual/native-image/#prerequisites
If it requires JDK >= 11 it is prefect that you are checking this here.
However, I would suggest that we use `EXTRA_JAVA_VERSION` only for IDEs and for all build related stuff we should ignore it and just honor `JAVA_VERSION` and the default Java (which will be on the PATH, in JAVA_HOME and physically in `software/java`).

hohwille (changes requested):
@tobka777 excellent, you did a great job and covered all the documentation and code changes for this story nicely. 👍 
I have one last review comment before we should merge this.
From that we should also test this new feature and see if we can build native image using GraalVM or need to address additional prerequisites. For this I would recommend that you clone https://github.com/devonfw-sample/devon4quarkus-reference.git and try to build a native image from that without using docker. If that works everything is perfect and our final goal is reached.

scripts/src/main/resources/scripts/command/graalvm
tobka777: I understood that because for the download of GraalVM there are two versions: for 11 and 17, so I thought that meant that you had to have version 11 or 17. I chose 11 so as not to require 17. Maybe I'am wrong.

Base JDKs (GraalVM Community) = OpenJDK 11.0.13 and 17.0.1

Source:
https://www.graalvm.org/downloads/
https://github.com/graalvm/graalvm-ce-builds/releases

tobka777:
I have tested with the sample project. For this, I still had to export the GRAALVM_HOME. After that it ran on macos. 
```
devon graalvm gu install native-image
#devon graalvm native-image
devon mvn package -Dnative -DskipTests
```

I tested Graalvm for Windows and Linux. For this, I still had to use the `.exe` and `.cmd` files. On Windows, it does not find the `cl.exe` because it is not in the path. I have extended the path with the function doExtendPath, but it still does not work. 
PowerShell has the problem of not executing the cmd files, but this works fine with Commandline.

I am a bit confused why the environment variables worked and sometimes not. Why can't you just do `export` in the command files, instead you have to put them in files that are loaded with `source`.

tobka777:
I found out, that the error cl.exe is not related to my graalvm commandlet. 
cl.exe is missing because I did not have Visual Studio installed with MSVC 2017 15.5.5 or later.

"To build a native GraalVM image on Windows, you'll need Microsoft Visual C++ (MSVC)."

Source:
- https://stackoverflow.com/a/64285183/10457835
- https://www.graalvm.org/22.0/reference-manual/native-image/#prerequisites

tobka777:
Now Image Native only works for windows in "x64 Native Tools Command Prompt".
Source: https://medium.com/graalvm/using-graalvm-and-native-image-on-windows-10-9954dc071311

Also tested for negative inference and multi-tenancy and there is no effect. `GRAALVM_HOME` is correct for each installation. Outside the devon installation, graalvm is called from `DEVON_IDE_HOME`.

So PR is ready for merge!

tobka777:
I have just looked at whether for Graalvm can take the latest version. On Github there are only release notes that contain the download links. But these are always with an exact version number. However, it also looks like there is only a new release about every half year.
https://github.com/graalvm/graalvm-ce-builds/releases

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,100 @@
...
+    then
+      doVersionCompare "${javaVersion}" "11u0"
+      if [ "${?}" == 2 ]
+      then
+        # Java version is lower than 11
+        doConfirmWarning "GraalVM requires Java 11 or newer.\nHowever, [EXTRA_]JAVA_VERSION is ${javaVersion} - in this setup GraalVM can not work.\nPlease update JAVA_VERSION or set EXTRA_JAVA_VERSION in your settings/devon.properties!"
+      fi
+    fi
+    doInstall "-" "${installDir}" "graalvm" "${version}"
+    echo -e "export GRAALVM_HOME=${GRAALVM_HOME}" >> "${DEVON_IDE_HOME}/conf/devon.properties"
```
hohwille, line 46: This is not really robust:
* what if `GRAAL[VM]_VERSION` is changed and the user reruns `devon graalvm setup`? This should upgrade graalvm - currently it IMHO does not. If you would change the `if` statement to make it work, the user would get multiple entries of `export GRAALVM_HOME=...` in his devon.properties. The same already happens now if you delete `software/graalvm` after installation and rerun `devon graalvm setup`.
* The user is in control of the file `conf/devon.properties`. He can edit this file. In case he adds a line like
`#end` to the end of the file that is not terminated with a newline, you would IMHO append it to `#endexport GRAALVM_HOME=....` and it would not work. Simply prefix with `\n` should solve this as you are already use `-e` option.

Please also add `GRAALVM_HOME` to documentation (`variables.asciidoc`).

hohwille (changes requested):
@tobka777 Great that you solved the windows problems. 👍 
I have added some last comments for rework and improvement.
Should not take long to resolve...

hohwille:
> Now, Image Native only works for windows in "x64 Native Tools Command Prompt".

Maybe we can have a small demo tomorrow. I still see some last obstracles - e.g. if you write `GRAALVM_HOME` to the `devon.properties` during the setup, it wont be exported without rerunning `devon` command alias to change env vars.
Not a real issue but we should see this from and end-user perspective:
* setup a fresh devonfw-ide with graalvm preconfigured
* clone quarkus project (what might also happen automatically via settings)
* CD to quarkus project and run maven build (`mvn package -Pnative` for our quarkus-reference)

Will this already work out of the box without errors?

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,100 @@
...
+  then
+    GRAALVM_HOME="${DEVON_IDE_HOME}/software/extra/graalvm/Contents/Home"
+  else
+    GRAALVM_HOME="${installDir}"
+  fi
+
+  doDevonCommand java setup silent
+  if [ ! -d "${installDir}" ]
+  then
+    local version="${GRAAL_VERSION:-22.0.0.2}"
```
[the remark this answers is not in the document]
tobka777: Ok I changed `GRAAL_VERSION` to `GRAALVM_VERSION`  as well in the documentation.

scripts/src/main/resources/scripts/command/graalvm
tobka777: Ok, I found a solution. GraalVM will also install other versions that are installed via `GRAALVM_VERSION` specified. Export Statement will only added to `devon.properties` if not added before. `\n` is added to the export statement.
`GRAALVM_HOME` is also added to the documentation.

tobka777:
So I had to add support for the extra folder in software to automatically install the tools.
There was also an error with maven which I solved.

tobka777:
> > Now, Image Native only works for windows in "x64 Native Tools Command Prompt".
> 
> Maybe we can have a small demo tomorrow. I still see some last obstracles - e.g. if you write `GRAALVM_HOME` to the `devon.properties` during the setup, it wont be exported without rerunning `devon` command alias to change env vars. Not a real issue but we should see this from and end-user perspective:
> 
>     * setup a fresh devonfw-ide with graalvm preconfigured
> 
>     * clone quarkus project (what might also happen automatically via settings)
> 
>     * CD to quarkus project and run maven build (`mvn package -Pnative` for our quarkus-reference)
> 
> 
> Will this already work out of the box without errors?

I tested this with my settings: https://github.com/tobka777/ide-settings
All is installed correctly, but `devon graalvm gu install native-image` had to be installed first.
After that the build process ran through without problems in the same commandline.

scripts/src/main/resources/scripts/command/ide
```diff
@@ -320,7 +320,7 @@ function doUpdateSoftware() {
   local tool_path
   if [ -d "${DEVON_IDE_HOME}/software" ]
   then
-    for tool_path in "${DEVON_IDE_HOME}/software"/*
+    for tool_path in "${DEVON_IDE_HOME}/software"/* "${DEVON_IDE_HOME}/software/extra"/*
```
hohwille, line 323: Mhm, I see your point here but this seems dangerous and might contradict with what we actually plan with `extra` folder (see e.g. #683).

IMHO you want to see if graalvm is installed already and therefore on an update it should also be considered so that a changed `GRAALVM_VERSION` would apply.

Should we maybe just do an extra case here for GraalVM until this is further clarified instead of scanning entire `extra` folder what will IMHO have undesired implications?

scripts/src/main/resources/scripts/command/ide
```diff
@@ -377,14 +377,14 @@ function doUpdateSoftware() {
   for tool in "${DEVON_IDE_TOOLS[@]}"
   do
     tool_path="${DEVON_IDE_HOME}/software/${tool}"
-    if [ "${tool}" == "maven" ]
+    if [ "${tool}" == "mvn" ]
```
hohwille, line 380: OMG, this was indeed swapped.
You found a stupid bug here and fixed it. Thanks!

hohwille (approved):
@tobka777 thanks. This is pretty cool now.
I would consider this PR to be ready but I have added one last comment for a final improvement.
If possible you can address this before we merge...

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,106 @@
...
+      java_binary="${GRAALVM_HOME}/bin/java.exe"
+    fi
+    doRunCommand "'${java_binary}' -version" "verify installation of GraalVM"
+  fi
+}
+
+# Call GraalVM with specified arguments
+function doRun() {
+  doSetup silent
+  command="$GRAALVM_HOME/bin/$*"
```
hohwille, line 68: IMHO this should be:
```
local binary="${GRAALVM_HOME}/bin/${1}"
```

So only the first arg is put in here whereas `$*` will fill in all args.
Example: ` devon graalvm gu install nodejs`
Here `$*` will be `gu install nodejs` and `$1` will only be `gu`.

The two other variables `command` and `subcommand` can then simply be removed.

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,106 @@
...
+  local binary=${command%% *}
```
hohwille, line 69: here we are removing a potential asterisk?
Maybe I am not getting what is really going on here...

scripts/src/main/resources/scripts/command/graalvm
tobka777: ok that's a better solution

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,106 @@
...
+  then
+    binary=${binary}".exe"
+  else
+    binary=${binary}".cmd"
+  fi
+
+  if [ -f "${binary}" ]
+  then
+    # shellcheck disable=SC2086
+    "${binary}" ${subcommand}
```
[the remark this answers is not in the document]
tobka777: You are right! The `shift` command is better.
But I have to pass `${*}`. With `${@}` I get only `install` and not `install nodejs`.

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,106 @@
...
+
+# Call GraalVM with specified arguments
+function doRun() {
+  doSetup silent
+  command="$GRAALVM_HOME/bin/$*"
+  local binary=${command%% *}
+  local subcommand=${command/"$binary"/}
+  if [ -f "${binary}" ]
+  then
+    binary=${binary}
```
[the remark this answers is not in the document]
tobka777: `-x` is better. I do not yet know all the operators ;).

scripts/src/main/resources/scripts/command/graalvm
```diff
@@ -0,0 +1,106 @@
...
+  else
+    binary=${binary}".cmd"
+  fi
+
+  if [ -f "${binary}" ]
+  then
+    # shellcheck disable=SC2086
+    "${binary}" ${subcommand}
+  else
+    doFail "Command not found"
```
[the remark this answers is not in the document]
tobka777: Ok I use `doRunCommand`. But I add a condition `[ -f "${binary}.cmd" ]` that the error is correct also for mac/linux.

hohwille (approved):
@tobka777 excellent. All perfect and ready for merge.

Issue #637, opened 2021-11-24
Terraform integration


SchettlerKoehler:
As a devon-ide user, I want to have Terraform available in the devon-ide so that I and my team can easily ramp up our development environment and we are using the same and correct versions of the tool.

Bene90:
Hi @CREITZ25, could you update that story? What ACs have been achieved so far?

hohwille:
done with PR #709

PR #709, opened 2022-02-22
#637 Terraform integration

CREITZ25:
Files to integrate terraform installation into devonfw IDE

### DIFF
--- a/scripts/src/main/resources/scripts/command/terraform
+++ b/scripts/src/main/resources/scripts/command/terraform
@@ -15,7 +15,6 @@ if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
 source "$(dirname "${0}")"/../functions
 
 # Check if terraform is installed.
-#
 function doIsTerraformInstalled() {
   if command -v terraform &> /dev/null
   then 
@@ -27,7 +26,7 @@ function doIsTerraformInstalled() {
 
 # Call Terraform with specified arguments.
 function doRun() {
-  doEcho "Running: terraform ${*}"
+  doSetup silent
   terraform "${@}"
 }
 
@@ -35,26 +34,27 @@ function doSetup() {
 
   if doIsTerraformInstalled 
   then
-    doEcho "terraform is already installed at $(command -v terraform)"
-    terraform version
-    echo
+    if [ "${1}" != "silent" ] && ! doIsQuiet
+    then 
+      terraform version
+      echo
+    fi 
   else
-      # Check if TERRAFORM_HOME is already set
-    #
-    if [ -z "$TERRAFORM_HOME" ]
+    # Check if TERRAFORM_HOME is already set
+    if [ -z "${TERRAFORM_HOME}" ]
     then
       doEcho "set TERRAFORM_HOME..."
-      export TERRAFORM_HOME="${DEVON_IDE_HOME}/software/terraform"
+      TERRAFORM_HOME="${DEVON_IDE_HOME}/software/terraform"
     fi
 
-    # get leatest release
-    if [ -z "$TERRAFORM_VERSION" ]
+    # Get leatest release
+    if [ -z "${TERRAFORM_VERSION}" ]
     then
       doEcho "Getting latest release..."
-      TERRAFORM_VERSION=$(curl -s https://checkpoint-api.hashicorp.com/v1/check/terraform | awk -F "," '/version/ {print $2}' | awk -F ":" '{print $2}'|sed 's/"//g')
+      TERRAFORM_VERSION=$(curl -s https://api.github.com/repos/hashicorp/terraform/releases/latest | awk -F ":" '/tag_name/ {print $2}'| awk -F "\"" '{print $2}' | sed 's/v//')
     fi
 
-    doInstall "-" "$TERRAFORM_HOME" "terraform" "${TERRAFORM_VERSION}"
+    doInstall "-" "${TERRAFORM_HOME}" "terraform" "${TERRAFORM_VERSION}"
     terraform -install-autocomplete
   fi
 }
@@ -71,16 +71,10 @@ case ${1} in
   echo
 ;;
 "setup" | "s" | "")
+  doEcho "Installating Terraform..."
   doSetup
 ;;
 *)
-  if doIsTerraformInstalled 
-  then
-    doRun "${@}"
-  else
-    doEcho "terraform is not installed, installing..."
-    doSetup
-    doRun "${@}"
-  fi
+  doRun "${@}"
 ;;
 esac
--- a/scripts/src/main/resources/scripts/command/terraform
+++ b/scripts/src/main/resources/scripts/command/terraform
@@ -45,13 +45,11 @@ function doSetup() {
     fi
   fi
 
-  if command -v terraform &> /dev/null 
+  if [ "${1}" != "silent" ] && ! doIsQuiet
   then
-    if [ "${1}" != "silent" ] && ! doIsQuiet
-    then
-      doRunCommand "terraform version"
-    fi 
-  fi
+    doRunCommand "terraform version"
+  fi 
+
 }
 
 
--- a/scripts/src/main/resources/scripts/command/terraform
+++ b/scripts/src/main/resources/scripts/command/terraform
@@ -21,12 +21,6 @@ function doRun() {
 }
 
 function doSetup() {
-  # Check if TERRAFORM_HOME is already set
-  if [ -z "${TERRAFORM_HOME}" ]
-  then
-    TERRAFORM_HOME="${DEVON_IDE_HOME}/software/terraform"
-  fi
-
   if [ ! -d ${TERRAFORM_HOME} ] || [ "${1}" != "silent" ]
   then
     # Get leatest release
@@ -38,8 +32,7 @@ function doSetup() {
 
     doInstall "-" "${TERRAFORM_HOME}" "terraform" "${TERRAFORM_VERSION}"
 
-    ERG=$(grep terraform ~/.bashrc >/dev/null; echo $?)
-    if [ ${ERG} != 0 ]
+    if ! grep terraform ~/.bashrc >/dev/null
     then
       terraform -install-autocomplete
     fi
--- a/scripts/src/main/resources/scripts/command/terraform
+++ b/scripts/src/main/resources/scripts/command/terraform
@@ -51,7 +51,7 @@ function doSetup() {
     if [ -z "$TERRAFORM_VERSION" ]
     then
       doEcho "Getting latest release..."
-      TERRAFORM_VERSION=`curl -s https://checkpoint-api.hashicorp.com/v1/check/terraform | awk -F "," '/version/ {print $2}' | awk -F ":" '{print $2}'|sed 's/"//g'`
+      TERRAFORM_VERSION=$(curl -s https://checkpoint-api.hashicorp.com/v1/check/terraform | awk -F "," '/version/ {print $2}' | awk -F ":" '{print $2}'|sed 's/"//g')
     fi
 
     doInstall "-" "$TERRAFORM_HOME" "terraform" "${TERRAFORM_VERSION}"
--- a/scripts/src/main/resources/scripts/command/terraform
+++ b/scripts/src/main/resources/scripts/command/terraform
@@ -17,24 +17,18 @@ source "$(dirname "${0}")"/../functions
 # Call Terraform with specified arguments.
 function doRun() {
   doSetup silent
-  terraform "${@}"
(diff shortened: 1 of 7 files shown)


scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+;;
+"setup" | "s" | "")
+  doSetup
+;;
+*)
+  if doIsTerraformInstalled 
+  then
+    doRun "${@}"
+  else
+    doEcho "terraform is not installed, installing..."
```
tobka777, line 81: Add echo statement to `doSetup`

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+  terraform "${@}"
+}
+
+function doSetup() {
+
+  if doIsTerraformInstalled 
+  then
+    doEcho "terraform is already installed at $(command -v terraform)"
+    terraform version
+    echo
```
tobka777, line 40: Remove wrong `echo` statement. It does nothing.

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+  else
+      # Check if TERRAFORM_HOME is already set
+    #
+    if [ -z "$TERRAFORM_HOME" ]
+    then
+      doEcho "set TERRAFORM_HOME..."
+      export TERRAFORM_HOME="${DEVON_IDE_HOME}/software/terraform"
```
tobka777, line 47: Should we export it system-wide like I did for `GRAALVM_HOME`?

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+  then 
+    return
+  else
+    return 255
+  fi
+}
+
+# Call Terraform with specified arguments.
+function doRun() {
+  doEcho "Running: terraform ${*}"
```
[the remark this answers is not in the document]
CREITZ25: command deleted

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+  if doIsTerraformInstalled 
+  then
+    doRun "${@}"
+  else
+    doEcho "terraform is not installed, installing..."
+    doSetup
+    doRun "${@}"
+  fi
+;;
+esac
```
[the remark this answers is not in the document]
CREITZ25: ok, you are right, I correct it

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+    fi
+
+    # get leatest release
+    if [ -z "$TERRAFORM_VERSION" ]
+    then
+      doEcho "Getting latest release..."
+      TERRAFORM_VERSION=$(curl -s https://checkpoint-api.hashicorp.com/v1/check/terraform | awk -F "," '/version/ {print $2}' | awk -F ":" '{print $2}'|sed 's/"//g')
+    fi
+
+    doInstall "-" "$TERRAFORM_HOME" "terraform" "${TERRAFORM_VERSION}"
```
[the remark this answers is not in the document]
CREITZ25: Of course

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+
+  if doIsTerraformInstalled 
+  then
+    doEcho "terraform is already installed at $(command -v terraform)"
+    terraform version
+    echo
+  else
+      # Check if TERRAFORM_HOME is already set
+    #
+    if [ -z "$TERRAFORM_HOME" ]
```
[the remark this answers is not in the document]
CREITZ25: Of course

scripts/src/main/resources/scripts/command/terraform
CREITZ25: I was not sure if there will somewhere started a sub shell. For that we need an export. I correct it

scripts/src/main/resources/scripts/command/terraform
CREITZ25: deleted

scripts/src/main/resources/scripts/command/terraform
CREITZ25: done

scripts/src/main/resources/scripts/command/terraform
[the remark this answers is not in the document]
CREITZ25: ok, done

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+
+# Call Terraform with specified arguments.
+function doRun() {
+  doEcho "Running: terraform ${*}"
+  terraform "${@}"
+}
+
+function doSetup() {
+
+  if doIsTerraformInstalled
```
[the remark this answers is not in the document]
CREITZ25: Ok, done

scripts/src/main/resources/scripts/command/terraform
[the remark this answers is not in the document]
hohwille: this change to add curly braces IMHO did not yet arrive in your PR.

scripts/src/main/resources/scripts/command/terraform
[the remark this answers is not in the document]
hohwille: typically always call `doSetup silent` before calling `doRun`. If terraform is already installed, this should do nothing otherwise it would install it (also also echo something already - so currently you might even get the "terraform is not installed, installing..." twice what would be confusing).

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function doSetup() {
+
+  if doIsTerraformInstalled 
+  then
+    if [ "${1}" != "silent" ] && ! doIsQuiet
+    then 
+      terraform version
+      echo
```
tobka777, line 40: Why we need this empty echo?

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,80 @@
...
+    fi 
+  else
+    # Check if TERRAFORM_HOME is already set
+    if [ -z "${TERRAFORM_HOME}" ]
+    then
+      doEcho "set TERRAFORM_HOME..."
```
tobka777, line 46: I would remove this `doEcho "set TERRAFORM_HOME..."`

CREITZ25: deleted

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,86 @@
...
+      # Check if TERRAFORM_HOME is already set
+    #
+    if [ -z "$TERRAFORM_HOME" ]
+    then
+      doEcho "set TERRAFORM_HOME..."
+      export TERRAFORM_HOME="${DEVON_IDE_HOME}/software/terraform"
+    fi
+
+    # get leatest release
+    if [ -z "$TERRAFORM_VERSION" ]
```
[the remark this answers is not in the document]
CREITZ25: ok, done

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,78 @@
...
+
+# Call Terraform with specified arguments.
+function doRun() {
+  doSetup silent
+  terraform "${@}"
+}
+
+function doSetup() {
+
+  if doIsTerraformInstalled
```
hohwille, line 35: Is there are real point in creating this function instead of just calling `command -v terraform &> /dev/null` directly in this `if` statement?
The function `doIsTerraformInstalled` is not called from any where else...

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,78 @@
...
+}
+
+function doSetup() {
+
+  if doIsTerraformInstalled 
+  then
+    if [ "${1}" != "silent" ] && ! doIsQuiet
+    then 
+      terraform version
+    fi
```
hohwille, line 40: This block should come at the end.
E.g. when you do `devon terraform setup` and meanwhile `TERRAFORM_VERSION` has changed, it will first print the (old) version (if it was instealld before) and then install the new version.
The result will be confusing for end users.
See other commandlets how the implement `doSetup` if unclear...

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,80 @@
...
+    then
+      doEcho "set TERRAFORM_HOME..."
+      TERRAFORM_HOME="${DEVON_IDE_HOME}/software/terraform"
+    fi
+
+    # Get leatest release
+    if [ -z "${TERRAFORM_VERSION}" ]
+    then
+      doEcho "Getting latest release..."
+      TERRAFORM_VERSION=$(curl -s https://api.github.com/repos/hashicorp/terraform/releases/latest | awk -F ":" '/tag_name/ {print $2}'| awk -F "\"" '{print $2}' | sed 's/v//')
```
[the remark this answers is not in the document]
hohwille: As we already discussed, different github projects may use completely different conventions so the `awk` pipe would look different. Lets just keep it and if we need this for other commandlets again, we can see if we can extract some function to avoid redundancies...

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,78 @@
...
+    fi
+
+    doInstall "-" "${TERRAFORM_HOME}" "terraform" "${TERRAFORM_VERSION}"
+    terraform -install-autocomplete
```
hohwille, line 56: nice 👍

hohwille (changes requested):
@CREITZ25 excellent work 👍 

I have two last tiny review comments added. Please have a look and update. Then I will merge. Thanks.

scripts/src/main/resources/scripts/command/terraform
CREITZ25: ok

scripts/src/main/resources/scripts/command/terraform
CREITZ25: No, you are right. It makes no sence. The same can be found in Commandlet docker with doIsDockerInstalled. It is really difficult, if I implement something in different way I was asked why do I implement it not in the same way as usually. For example: I prefer to add an single empty echo after an output to make sure it is visible that outputs generated from different commands. But thats not usual in our environment so I killed that echo again. 

I will change it, but please, tell me where are the developer guides, so I don't fall into the trap again.

By the way: since the last push the commandlet was deleted from my workspace and I don't understand why

scripts/src/main/resources/scripts/command/terraform
CREITZ25: Sorry, I don't understand.

First I asked if terraform is installed.
if yes I display the version
If not, I install it

doInstall only starting if terraform is not installed

hohwille:
The build failed due to shellcheck:
```
In scripts/command/terraform line 25:
  if $(command -v terraform &> /dev/null) 
     ^-- SC2091: Remove surrounding $() to avoid executing output.
```

scripts/src/main/resources/scripts/command/terraform
hohwille: 1. set TERRAFORM_VERSION=1.0.11
2. install terraform (`devon terraform setup`)
3. set TERRAFORM_VERSION=1.1.7
4. install/upgrade terraform (`devon terraform setup`)

Actual behaviour:
2. will not output any version and 4. will output version 1.0.11

Expected behaviour:
The version is outputted after the installation so 2. will give 1.0.11 and 4. will give 1.1.7

CREITZ25: I think it is now implemented in the right way. Please have a look

scripts/src/main/resources/scripts/command/terraform
```diff
@@ -0,0 +1,68 @@
...
+
+# Call Terraform with specified arguments.
+function doRun() {
+  doSetup silent
+  terraform "${@}"
+}
+
+function doSetup() {
+
+  if $(command -v terraform &> /dev/null)
```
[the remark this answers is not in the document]
CREITZ25: done

CREITZ25 (commented):
Please check again

hohwille (approved):
@CREITZ25 all perfect, thanks 👍

hohwille:
With my last commit the build should be fixed now.

Issue #636, opened 2021-11-24
Helm integration


SchettlerKoehler:
As a devon IDE user, I want to have Helm available in the devon IDE so that I and my team can easily ramp up our development environment and we are using the same and correct versions of the tool.

Acceptance criteria:

- [x] There is a new commandlet that allows to install and launch Helm from the CLI
- [x] The new commandlet is a wrapper for Helm CLI
- [x] Helm can be installed automatically (during setup via settings) as well as via the commandlet call
- [x] During the installation process I get information about the current status of the installation
- [x] The new commandlet is documented on the website
- [x] Helm can be installed and used together with any of the devon IDE standard tools without negative interference
- [x] Helm is integrated in a way that complies with the multi-tenancy of the devon IDE
- [x] Add licence documentation

Bene90:
Hi @CREITZ25, could you update that story? What ACs have been achieved so far?

CREITZ25:
Hi @Bene90, many things are done, but at the moment the workflow doesn't work due to spellcheck errors. Still open is to check Mac and Linux, and License.asciidoc has to be enhanced. It is not done because the pull request with terraform stopped by spellcheck errors, too.

Bene90:
Hi @CREITZ25, ok - could you create seperate tasks for these open bugs / ToDos?

PR #714, opened 2022-03-07
#636 Helm Integration

CREITZ25:
#636 Helm Integration

### DIFF
--- a/scripts/src/main/resources/scripts/command/helm
+++ b/scripts/src/main/resources/scripts/command/helm
@@ -14,37 +14,21 @@ if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
 
-# Check if helm is installed.
-function doIsHelmInstalled() {
-  if command -v helm &> /dev/null
-  then
-    return
-  else
-    return 255
-  fi
-}
-
 # Call helm with specified arguments.
 function doRun() {
   doSetup silent
-  helm "${@}"
+  doRunCommand "${HELM_HOME}/helm ${@}"
 }
 
 function doSetup() {
-
-  if doIsHelmInstalled
+  # Check if HELM_HOME is already set
+  if [ -z "${HELM_HOME}" ]
   then
-    if [ "${1}" != "silent" ] && ! doIsQuiet
-    then
-      helm version
-    fi
- else
-    # Check if HELM_HOME is already set
-    if [ -z "${HELM_HOME}" ]
-    then
-      HELM_HOME="${DEVON_IDE_HOME}/software/helm"
-    fi
+    HELMFORM_HOME="${DEVON_IDE_HOME}/software/helm"
+  fi
 
+  if [ ! -d ${HELM_HOME} ] || [ "${1}" != "silent" ]
+  then
     # Get leatest release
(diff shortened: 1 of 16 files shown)


scripts/src/main/resources/scripts/command/helm
```diff
@@ -0,0 +1,77 @@
...
+  helm "${@}"
+}
+
+function doSetup() {
+
+  if doIsHelmInstalled
+  then
+    if [ "${1}" != "silent" ] && ! doIsQuiet
+    then
+      helm version
```
hohwille, line 39: same issue(s) as with terraform.
Can you move the version output to the end (any consider using `doRunCommand` that will also add automatic error handling)?

hohwille (changes requested):
@CREITZ25 analog to #709 
Also I am missing the change to [`documentation/LICENSE.asciidoc`](https://github.com/devonfw/ide/blob/master/documentation/LICENSE.asciidoc)

scripts/src/main/resources/scripts/command/helm
```diff
@@ -0,0 +1,60 @@
...
+function doRun() {
+  doSetup silent
+  doRunCommand "${HELM_HOME}/helm ${@}"
+}
+
+function doSetup() {
+  # Check if HELM_HOME is already set
+  if [ -z "${HELM_HOME}" ]
+  then
+    HELMFORM_HOME="${DEVON_IDE_HOME}/software/helm"
```
hohwille, line 27: hast to be `HELM_HOME` instead of `HELMFORM_HOME` :)

scripts/src/main/resources/scripts/command/helm
```diff
@@ -0,0 +1,60 @@
...
+  then
+    # Get leatest release
+    if [ -z "${HELM_VERSION}" ]
+    then
+      doEcho "Getting latest release..."
+      HELM_VERSION=$(curl -LsH "Accept: application/vnd.github.v3+json" https://api.github.com/repos/helm/helm/releases/latest | awk -F':' '/tag_name/ {print $2}' | awk -F'"' '{print $2}')
+    fi
+
+    doInstall "-" "${HELM_HOME}" "helm" "${HELM_VERSION}"
+  fi
```
hohwille, line 40: now the version check output has been removed entirely, it should happen here if not silent

hohwille (changes requested):
setup now working 👍 
only a typo left and the version check output should still be present

scripts/src/main/resources/scripts/command/helm
```diff
@@ -0,0 +1,69 @@
...
+
+  if command -v helm &> /dev/null
```
hohwille, line 42: just for curiosity - what happened without this if?
Actually at this point helm must be present or the above code did not do its job.
Du you fear any PATH variable issues?
In other commandlets we do not have such check...

hohwille (approved):
excellent, perfect now.

scripts/src/main/resources/scripts/command/helm
CREITZ25: I removed the wrong if case. Same in Terraform

hohwille:
With my latest commit the build should go green and I can merge.

hohwille:
@CREITZ25 
license/cla Expected — Waiting for status to be reported 

Did you forgot to sign the CLA? This is required once for every developer in devonfw.

PR #716, opened 2022-03-15
fix for correct .m2 folder

### LINKED ISSUE #706: Folder ${M2_REPO} gets created
### Expected behavior

All maven artifacts should go to the local maven repo configured via `devon.properties` (by default `$DEVON_IDE_HOME/conf/.m2/repository` but can also be changed to e.g. `~/.m2/repository` via `devon.properties`).

### Actual behavior

For some odd reason some tool or script causes a folder named `${M2_REPO}` being created in the top-level IDE installation `$DEVON_IDE_HOME`).

1. Create a devonfw-ide installation and run setup
2. Use this IDE to work with maven and eclipse in your project
3. at some point (exact steps for 2. still unclear) this magic `${M2_REPO}` folder appears and is filled with some maven artifacts.

### Comments/Hints:

Seems as if at some point `${M2_REPO}` is used as variable but does not properly get resolved leading to the fact that it is used as relative path for the maven repository created as folder.

### Affected version:

* OS: happened both on Windows and Mac

tobka777:
fixes #706


scripts/src/main/resources/scripts/command/mvn
```diff
@@ -306,7 +306,7 @@ function doCheckNoSnapshots() {
 
 function doRunBuild() {
   local maven_cmd="${MVN}"
-  local maven_arguments=""
+  local maven_arguments="--settings=${DEVON_IDE_HOME}/conf/.m2/settings.xml"
```
hohwille, line 309: dont we need to export this to `MAVEN_ARGS` via `devon.properties?
This only applies when `mvn` is called via `devon` as wrapper command.
However, I see this bug occurring all the time without `devon` command being involved as wrapper but only to setup env variables.
My assumption is that the actual bug #706 somehow originates from m2e eclipse plugin and happens since a newer version of that plugin so it did not happen before and suddenly came with some ide or eclipse update...

hohwille: The current "hack" to relocate the settings to this directory happens here:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/devon.properties#L19

So far this worked well for commandline (with and without `devon` wrapper) as well as for all supported IDEs.

hohwille (approved):
@tobka777 thanks for this PR.
I am happy to approve it.
However, I am quite sure that it is not really related to the bug #706 and will not fix it.

scripts/src/main/resources/scripts/command/mvn
```diff
@@ -319,17 +319,17 @@ function doRunBuild() {
     if [ "${?}" == 2 ]
     then
       # Maven version is lower than 4.0 (3.x)
-	  maven_arguments="${MAVEN_ARGS}"
-	fi
+	    maven_arguments="${MAVEN_ARGS}"
+	  fi
```
hohwille, line 323: did those changes happen by accident? Seems the indentation was messed by your editor somehow.

scripts/src/main/resources/scripts/command/mvn
hohwille: later down the `maven_arguments` are overridden (`maven_arguments="${MAVEN_ARGS}"`) so with maven 3.x your change will have no effect. As maven 4.x is still not released, this PR does not work or make a big difference.
Should we close this or should we change the later code in this PR to honor existing `maven_arguments` and append to it instead of overriding.

hohwille (changes requested):
I just stumbled again over this PR when trying to review all open PRs and found a problem.
Once upon a time, you could rework or simply close this PR.

tobka777:
@hohwille I think we can close the ticket for now. I think @alfeilex has identified the main problem with his fix.
https://github.com/devonfw/ide/pull/814

PR #718, opened 2022-03-18
Generalize Workaround for macos

### LINKED ISSUE #683: Generalize MacOS app workaround

tobka777:
fixes #683


tobka777:
I implemented a general workaround for java and graalvm.
For Eclipse, Intellij, VSCode we have to look for another solution. Now they use a script to pass the arguments.

Issue #690, opened 2022-01-19
Support for RancherDesktop


JokeFrels:
As a project that has no license for docker desktop, I want to have an alternative docker variant as part of devon IDE so that I can still use devon IDE to set up docker.

Acceptance criteria:

- If the new variable DOCKER_EDITION is set to DOCKERDESKTOP than the current behavior is preserved (docker desktop). Else the licence free variant is used.
- The new variable ist documented on the website
- The current docker commandlet can be used as well for Docker desktop as for the licence free variant depending on which variant is defined in the settings
- The description of the docker commandlet on the website covers Docker desktop as well as the licence free variant
- During the installation process of the licence free variant I get information about the current status of the installation
- The licence free variant works with Kubernetes
**- Add licence documentation**

19.4: 
- Check mirror file @cinnamon-coder-hub 
- Testing on MAC OS is open (integration has been done) @tobka777

PR #727, opened 2022-04-19
690 cmdlet rancher desktop

cinnamon-coder-hub:
fixing #690. 
Additionally an issue with the mirror has to be and will be fixed soon. There is a formating problem when uploading to github, similar as @CREITZ25 experienced once.

### DIFF
--- a/documentation/docker.asciidoc
+++ b/documentation/docker.asciidoc
@@ -5,6 +5,10 @@ toc::[]
 
 The Docker commandlet allows to install and use https://www.docker.com/[Docker].
 On Windows https://docs.microsoft.com/en-us/windows/wsl/install-win10[WSL 2](Windows Subsystem for Linux) has to be installed properly as a prerequisite.
+It also offers the installation of https://rancherdesktop.io/[Rancher Desktop], which works on Windows, Mac and Linux.
+To run the tools installed and managed by Rancher Desktop it is necessary to have Rancher Desktop up and running.  
+To assure that you are able to use these tools directly after startup, you can add Rancher Desktop to the start up programs
+of your Operating System. For Windows 10 you can use the following https://www.dell.com/support/kbdoc/de-de/000124550/how-to-add-app-to-startup-in-windows-10[Instructions on how to set start up programs].
 
 *ATTENTION:*
 Currently this feature is new and therefore experimental.
@@ -25,9 +29,20 @@ The arguments (`devon docker «args»`) are explained by the following table:
 |=======================
 
 == setup
-The Docker commandlet will install Docker automatically.
+The Docker commandlet will install Docker Desktop or Rancher Desktop automatically.
+This depends on how the variable DOCKER_EDITION was set. We chose to set Rancher Desktop as standard
+for our installation, due to it being license free compared to Docker Desktop while offering a similar tool set (please read the document docker-desktop-alternative.asciidoc, in the same folder as this document, motivating why we added Rancher Desktop to the docker commandlet).
+Docker Desktop can still be used as before, by setting the variable DOCKER_EDITION equal to "DOCKERDESKTOP". 
+This can be done by your projects ide-admin in the https://github.com/devonfw/ide/blob/master/documentation/usage.asciidoc#developer[settings file]. If you want to switch from Rancher Desktop to Docker Desktop, then add a line with the following content 
+"DOCKER_EDITION=DOCKERDESKTOP" 
+(without the quotes around the allocation) to the file devon.properties located at the path ${DEVON_IDE_HOME}/conf/devon.properties.
+
+After successfull installation please start Rancher Desktop and go to the menu point "Kubernetes Settings". Under "Container Runtime" choose dockerd(moby) to enable the docker-cli. If you use Windows, then make sure to go to the Settings named "WSL Integration".
+There you can choose the WSL Shells you want Rancher Desktop to work on. Besides that the tools like docker or kubectl should 
+directly work with CMD or Powershell.
+
 Please note that besides the sandbox concept of devonfw-ide this is a global installation on your system.
-When uninstalling devonfw-ide, you may have to manually uninstall Docker and Kubernetes if you do not need it anymore.
+When uninstalling devonfw-ide, you may have to manually uninstall Docker and Kubernetes, or Rancher Desktop if you do not need it anymore. At least uninstalling Rancher Desktop should leave way less remains if you did not install one of the related software  Rancher Desktop offers separately.
 
 === requirements
 Running Docker and especially link:kubectl.asciidoc[Kubernetes] on your machine in the background will require quite some resources.
@@ -36,7 +51,7 @@ This will allocate at least 2GB of additional RAM.
 You will need at least 8GB of total RAM while we recommend to use 16GB+.
 
 You may also tune and scale it to your needs.
-When using Docker Desktop (Windows or MacOS) simply go to the https://docs.docker.com/docker-for-windows/#resources[resources tab in the settings].
+When using Docker Desktop (Windows or MacOS) simply go to the https://docs.docker.com/docker-for-windows/#resources[resources tab in the settings]. 
 It will depend on your usage frequency if you want to have it running in the background all the time.
 This is a balance between resource utilization and convenience.
 If you use Docker and Kubernetes on your local machine on a daily basis this makes sense.
@@ -44,7 +59,7 @@ If you use Docker and Kubernetes on your local machine on a daily basis this mak
 In case you only use Docker rarely, you can save resources by stopping it when not needed after it has been installed.
 
 === Windows and macOS
-To enable or disable autostart, you can launch Docker Desktop on Windows or MacOS got to the `Preferences` (gear icon in the title bar). Then in the `General` tab you can check or uncheck the option `Start Docker Desktop when you login` (see also https://docs.docker.com/docker-for-windows/#general[here]). When autostart is disabled and you launch Docker Desktop it will notice and ask your to start the service or do this automatically for you.
+To enable or disable autostart, you can launch Docker Desktop on Windows or MacOS go to the `Preferences` (gear icon in the title bar). Then in the `General` tab you can check or uncheck the option `Start Docker Desktop when you login` (see also https://docs.docker.com/docker-for-windows/#general[here]). When autostart is disabled and you launch Docker Desktop it will notice and ask your to start the service or do this automatically for you.
 On Windows you can also manually tweak this:
 
 * Hit `[windows][r]`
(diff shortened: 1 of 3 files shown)


scripts/src/main/resources/scripts/command/docker
```diff
@@ -16,46 +16,124 @@ source "$(dirname "${0}")"/../functions
...
+      elif doIsMacOs
+      then
+          doEcho "installing Docker Desktop for MacOS..."
+        local mount_dir="${DEVON_IDE_HOME}/updates/volume"
+        mkdir -p "${mount_dir}"
+        doRunCommand "hdiutil attach -quiet -nobrowse -mountpoint '${mount_dir}' '${DEVON_DOWNLOAD_DIR}/Docker.dmg'"
+        doRunCommand "cp -a '${mount_dir}'/*.app /Applications/"
+        doRunCommand "hdiutil detach -force '${mount_dir}'"
+      fi
+      read -r -p "Docker Desktop has been installed. Please read https://github.com/devonfw/ide/blob/master/documentation/docker.asciidoc#setup for enabling kubernetes and further important details. Press enter to continue" answer
```
tobka777, line 130: Why there is not only an echo? I was surprised that I still had to press Enter.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -16,46 +16,124 @@ source "$(dirname "${0}")"/../functions
...
+          os="mac"
+          #options that work with mac OS
+          #arch="arm64"
+          #arch="x86_64"
+          arch="$(uname -m)"
+        else
+          arch="-"
+          ext="other"
+          os="linux"
+        fi
```
tobka777, line 75: Isn't there a more elegant solution, because mirrors should take care of everything. Maybe work with code like docker instead. So ideally it's just `doInstall "-" "${installDir}" "rancher" "${version}"` at the end.

tobka777 (changes requested):
So I tested rancher for mac and your solution works with `hdiutil`.
Instead of .dmg you could also use the zip (https://github.com/rancher-sandbox/rancher-desktop/releases/download/v1.2.1/Rancher.Desktop-1.2.1-mac.x86_64.zip)  file, but that's fine. Then you would not need to mount the dmg.
I still had to adjust the name of the .dmg.

scripts/src/main/resources/scripts/command/docker
cinnamon-coder-hub: I would also love that, but the structure of my code is a kind of forced reflection of the urls structure 😆

documentation/docker.asciidoc
```diff
@@ -25,9 +28,20 @@ The arguments (`devon docker «args»`) are explained by the following table:
 |=======================
 
 == setup
-The Docker commandlet will install Docker automatically.
+The Docker commandlet will install Docker Desktop or Rancher Desktop automatically.
+This depends on how the variable DOCKER_EDITION was set. We chose to set Rancher Desktop as standard
+for our installation, due to it being license free compared to Docker Desktop while offering a similar tool set (please read the document link:docker-desktop-alternative.asciidoc[docker-desktop-alternative.asciidoc] motivating why we added Rancher Desktop to the docker commandlet).
```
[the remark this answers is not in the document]
cinnamon-coder-hub: @hohwille I just added the include you mentioned to devonfw-ide-usage.asciidoc. If that is not what you meant, please let me know so I can correct this soon.

hohwille (approved):
@cinnamon-coder-hub Excellent. Thanks for your updates 👍 
We should surely do more testing and collect early user-feedback but now approved and ready to be merged.

scripts/src/main/resources/scripts/command/docker
hohwille: Maybe we can get the changes from @CREITZ25 for `doInstall` and improve this here as well before we merge.
Otherwise, we can also merge as is and create a new story for such improvement.

hohwille: So our conclusion is to wait for this change, merge it into your feature branch and then remove this code-block and use `doInstall` directly instead of `doDownload`?

hohwille (approved):
We create extra  issues for doc improvements (installation help, screenshot of install options) and improvement for #742 as well as MacOS testing - rest is fine so ready for merge

tobka777 (approved):

scripts/src/main/resources/scripts/command/docker
```diff
@@ -16,46 +16,125 @@ source "$(dirname "${0}")"/../functions
...
+          fi
+        fi
+      else
+        doRequireNotBatch
+        local latestversion
+        latestversion=$(curl -Ls -o /dev/null -w '%{url_effective}' https://github.com/rancher-sandbox/rancher-desktop/releases/latest | sed 's/.*\([0-9]\.[0-9]\.[0-9]\).*/\1/')
+        local previousversion="1.2.1"
+        if [[ "${latestversion}" =~ ^[0-9]\.[0-9].[0-9]$ ]] 
+        then
+          if [ "${latestversion}" \> "${previousversion}" ]
```
hohwille, line 48: is this comparison syntax working reliably?
We have implemented our own function `doVersionCompare` in `functions` that was quite complex.
If this works as well, we could strip out `doVersionCompare` and replace it with this simplified syntax.
Otherwise, if this does not work for version numbers containing letters, `alpha`, `beta`, or multiple dots we should also use `doVersionCompare` here for consistency and correctness.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -16,46 +16,125 @@ source "$(dirname "${0}")"/../functions
...
+          then
+            doEcho "setting RANCHER_VERSION"
+            RANCHER_VERSION=${latestversion}
+            doEcho "RANCHER_VERSION is ${RANCHER_VERSION}"
+          fi
+        fi
+        local version="${RANCHER_VERSION:-1.2.1}"
```
hohwille, line 55: Version `1.2.1` gets duplicated here with `previousversion` in line 45.
You could avoid this redundancy.
Further the code above is inefficient as the latest version is always determined causing `curl` and `http` remote invocations but if `${RANCER_VERSION}` is set the latestversion and the result of the version-comparison is always discarded.
So IMHO simply do this:
```
local version="${RANCHER_VERSION}"
if [ -z "${version}" ]
then
  version=1.2.1
  # ... determine latest version and update version if newer ...
fi
```

BTW: I still do not get what we need `1.2.1` for if we simply get the latest version from CURL. So if CURL fails we use `1.2.1` as fallback, right?

hohwille (changes requested):
@cinnamon-coder-hub I did some constructive review with 2 direct commits of small and formal improvements.
Further I just wanted to get this ready for merge but found two more small things that I added as review comments. Please have a look.

scripts/src/main/resources/scripts/command/docker
cinnamon-coder-hub: Looks as if `doVersionCompare` has more power that my code-snippet for this one special case does (also there was a typo, another backslash would be needed before the second dot, to only accept dots in between the numbers). 
Therefore I switched to using this function and keep its usage in mind for future use cases. 
Thanks for the detailed feedback.

cinnamon-coder-hub: sorry, I must have missed to publish the changes after commiting them last week. Now the changes are also published.

hohwille: Now, the commit arrived and I can resolve this.

scripts/src/main/resources/scripts/command/docker
hohwille: Also solved. Thanks.

scripts/src/main/resources/scripts/command/docker
hohwille: This is now done. Thanks.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -72,6 +126,65 @@ function doIsDockerInstalled() {
...
+  #get latest version of the vpnkit and download its tar.gz file
+  partialUrl=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/sakai135/wsl-vpnkit/releases/latest)
+  VPNversion=${partialUrl##*/}
+  downloadUrl="https://github.com/sakai135/wsl-vpnkit/releases/download/${VPNversion}/wsl-vpnkit.tar.gz"
+  curl -fL "${downloadUrl}" -o "wsl-vpnkit.tar.gz"
+  touch activate-vpnkit
+  touch vpnkit-setup.ps1
+
+  #at first write a bash-script
+  touch activate-vpnkit
```
hohwille, line 165: this command is duplicated with line 161 and you would not even need this if you use `>` instead of `>>` in line 166.
But if we want some nice-to-have perfection tidy up here, we can also do it in the next PR after merge.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -72,6 +126,65 @@ function doIsDockerInstalled() {
...
+        return 255
+      fi
+    fi
+}
+
+function doInstallVpnkit() {
+  #get latest version of the vpnkit and download its tar.gz file
+  partialUrl=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/sakai135/wsl-vpnkit/releases/latest)
+  VPNversion=${partialUrl##*/}
+  downloadUrl="https://github.com/sakai135/wsl-vpnkit/releases/download/${VPNversion}/wsl-vpnkit.tar.gz"
```
hohwille, line 159: variables should also be local. Will be done in next PR.

hohwille (approved):
@cinnamon-coder-hub Thanks for all the rework and great that you could already integrate vpnkit. I consider this as ready for merge. Great job 👍

hohwille:
Shellcheck still fails:
```
In scripts/command/docker line 157:
  partialUrl=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/sakai135/wsl-vpnkit/releases/latest)
                                         ^-- SC1083: This { is literal. Check expression (missing ;/\n?) or quote it.
                                                       ^-- SC1083: This } is literal. Check expression (missing ;/\n?) or quote it.


In scripts/command/docker line 173:
  echo 'mv wsl-vpnkit.tar.gz $env:USERPROFILE' >> vpnkit-setup.ps1
  ^-- SC2129: Consider using { cmd1; cmd2; } >> file instead of individual redirects.
       ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.


In scripts/command/docker line 175:
  echo 'pushd $env:USERPROFILE' >> vpnkit-setup.ps1
       ^----------------------^ SC2016: Expressions don't expand in single quotes, use double quotes for that.


In scripts/command/docker line 176:
  echo 'wsl --import wsl-vpnkit $env:USERPROFILE\wsl-vpnkit wsl-vpnkit.tar.gz --version 2' >> vpnkit-setup.ps1
       ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.
```

I will quickly fix this oneline to get this merged and do release today...

Issue #712, opened 2022-02-24
Azure CLI integration


albertdb:
As a devon IDE user, I want to have Azure CLI available in the devon IDE so that I and my team can easily ramp up our cloud environment and we are using the same and correct versions of the tool.
 
Acceptance criteria:

- [x] Mirror is added.
- [x] There is a new commandlet that allows to install and launch Azure CLI from the CLI
- [x] The new commandlet is a wrapper for Azure CLI
- [x] Azure CLI (latest version) can be installed automatically (during setup via settings) as well as via the commandlet call - independently of the used OS plattform   (~~Linux~~, Windows, ~~MAC~~)
- [x] During the installation process I get information about the current status of the installation
- [x] The new commandlet is documented on the website
- [x] Azure CLI can be installed and used together with any of the devon IDE standard tools without negative interference

- [x] The new commandlet is documented on the website
- [x] Add licence documentation

19.04
- [x] Testing MAC OS / ~~Linux~~

3.5
- [x] Testing MAC OS
- [x] Adding more feedback/response during the installation
- [x] Azure CLI is integrated in a way that complies with the multi-tenancy of the devon IDE - this implies that configuration is stored per IDE instance and not globaly in users home directory.

albertdb:
Use case: https://github.com/devonfw/hangar/blob/master/documentation/azure-devops/setup-project.asciidoc

alfeilex:
How to install azure cli locally? I see in the [documentation](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest) only the possibility to install azure cli globally. 

The question refers to the following A.C: 

> _Azure CLI is integrated in a way that complies with the multi-tenancy of the devon IDE - this implies that configuration is stored per IDE instance and not globaly in users home directory._

alfeilex:
Azure CLI can be installed for Windows and MAC via command (devon az) 

The implementation for Linux is not done, because for different Linux distributions different package manager are needed to run the installation. (e.g. Ubuntu -> apt | Red Hat -> yum)

This is a topic for the next sprint meeting.

alfeilex:
@tobka777 Please test the command on your MAC. I tried the installation for MAC on my virtual machine. There was an error at first because of virtualenv (Python). I was able to fix the error by running following command.

`conda install virtualenv`

The following [requirements](https://github.com/alfeilex/ide/blob/master/documentation/azurecli.asciidoc#requirements) are given for installation on MAC via bash.

[...]

PR #733, opened 2022-04-28
devonfw/ide#712: azcli integration

alfeilex:
I tried the installation for MAC on my virtual machine. There was an error at first because of virtualenv (Python). I was able to fix the error by running following command.

`conda install virtualenv`

The following [requirements](https://github.com/alfeilex/ide/blob/master/documentation/azurecli.asciidoc#requirements) are given for installation on MAC via bash.


alfeilex:
Have to change #712

Issue #734, opened 2022-05-02
unzip: command not found (create doUnzip function)


Bene90:
```
unzip: command not found
```

On some systems the `unzip` command is not preinstalled (including linux distros like Ubuntu)

hohwille:
As a solution we plan to add `doUnzip` to `functions`.
In addition to calling `unzip` command with given params, it should first check if `unzip` command is present.
If not, the command should be installed according to current OS (or if not possible the user should be informed to install it manually giving an URL with instructions).

PR #737, opened 2022-05-02
#734 create doUnzip function

CREITZ25:
#734 create doUnzip function

The following command in functions has to be changed to activate the doUnzip:
#doRunCommand "unzip -qn '${1}' -d '${target_dir}'"
doUnzip ${1} ${target_dir}


hohwille (approved):
👍 well done. Excellent. Thanks @CREITZ25

PR #739, opened 2022-05-12
#717: fix intellij fails on update

### LINKED ISSUE #717: intellij makes devon ide update fail
### Expected behavior

As a user, I want to update my IDE according to the latest settings so that my tools are in the correct version

### Actual behavior

when I run `devon ide update`, I get this error:

```
******** ATTENTION ********
ERROR: Failed to verify installation of IntelliJ (command)
Exit code was 1
ERROR: command 'intellij setup' failed with exit code 1
```

As this terminates the entire update process, I am unable to proceed with my actual update.
Since I need to update a custom tool there seems to be now way to proceed (for a regular tool, I could run `devon <tool> setup` to trigger the update.

1. install devonfw-ide (`setup`)
2. install intellij (`devon intellij setup`)
3. run `devon ide update`

### Comments/Hints:

1. We should fix the actual bug with intellij update
2. We should consider to resume the update on the error of an individual tool (maybe with `doAskToConfirm`)

Workaround: delete `software/intellij` before running `devon ide update`.

### Affected version:

* 2021.12.005

tobka777:
fixes #717

### DIFF
--- a/scripts/src/main/resources/scripts/command/intellij
+++ b/scripts/src/main/resources/scripts/command/intellij
@@ -44,7 +44,6 @@ function doSetup() {
   fi
   if [ "${1}" != "silent" ] && ! doIsQuiet
   then
-    doRunCommand "command -v idea" "verify installation of IntelliJ"
     if [ -e "${IDEA_HOME}/product-info.json" ]
     then
       doRunCommand "cat '${IDEA_HOME}/product-info.json'"


hohwille (approved):
@tobka777 thanks for the quickfix. We should still implement #717 and prevent a single failure to stop the entire update of other tools. But great to have this as quickfix for the next release. And if idea does not further support this `-v` option we should strip it out anyways.

PR #741, opened 2022-05-12
#712: Azure CLI integration

alfeilex:
Azure CLI can be installed in Windows via Command `devon az`


scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,136 @@
...
+function modifyPathStructure(){
+  modifiedPath=${DEVON_DOWNLOAD_DIR}
+  partition=$(echo "${modifiedPath}" | cut -f2 -d'/'):'/' # -> <partition>:/
+  modifiedPath=${partition}$(echo "${modifiedPath}" | cut -f3- -d'/')
+  modifiedPath="${modifiedPath//'/'/'\'}"
+}
+
+# Install Azure CLI for Windows via Powershell Command
+function installOnWindows(){  
+  powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${modifiedPath}\az-${version}-windows.msi /QB-!'" &> /dev/null || exit_code=$?
```
tobka777, line 55: Is there not an other way for windows to use cmd.exe? If not then powershell should be add to the requirements and checked here if installed.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,136 @@
...
+    esac
+  fi
+  }
+
+# Set directory for azure cli configuration file
+function setConfigDir() {
+  azurecli_export="export AZURE_CONFIG_DIR=${DEVON_IDE_HOME}/conf/.azure"
+  if ! grep -q "${azurecli_export}" "${DEVON_IDE_HOME}/conf/devon.properties"
+    then
+      doRunCommand "export AZURE_CONFIG_DIR=${DEVON_IDE_HOME}/conf/.azure"
```
tobka777, line 80: Can you run `doRunCommand $(azurecli_export)` to avoid duplicate code to line 77?

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,136 @@
...
+      echo -e "\n${azurecli_export}" >> "${DEVON_IDE_HOME}/conf/devon.properties"
+      doEcho "Location of Azure's configuration file is set to ${DEVON_IDE_HOME}/conf/.azure"
+  fi
```
tobka777, line 83: Maybe it is a good solution to create a new variable for `${DEVON_IDE_HOME}/conf/.azure` and use it to avoid duplicate code.

documentation/az.asciidoc
```diff
@@ -0,0 +1,50 @@
...
+- https://docs.microsoft.com/de-de/cli/azure/install-azure-cli-windows?tabs=azure-cli#uninstall[Windows]
+
+- https://docs.microsoft.com/de-de/cli/azure/install-azure-cli-linux?pivots=script#uninstall[MacOS and Linux]
+
+== Requirements
+**MacOS and Linux users requires the following software:**
+
+- https://www.python.org/downloads/[Python 3.6.x, 3.7.x oder 3.8.x]
+- https://sourceware.org/libffi/[libffi]
+- https://www.openssl.org/source/[OpenSSL 1.0.2]
```
tobka777, line 41: If we don't support MacOS and Linux I think we can remove the requirements for Linux and MacOS.

tobka777, line 41: Please add in the documentation, that the commandlet `az` can be installed only on Windows.

tobka777 (changes requested):
Good work!
I have added some comments that you are welcome to implement or even discuss if you see something differently.
Please still agree with the Constribution Guildlines.

scripts/src/main/resources/scripts/command/az
alfeilex: Added doIsPwshInstalled function

scripts/src/main/resources/scripts/command/az
alfeilex: command runs

scripts/src/main/resources/scripts/command/az
alfeilex: done

documentation/az.asciidoc
alfeilex: Done

alfeilex: done. I add this information in the first section of the doc file

tobka777 (approved):

documentation/az.asciidoc
```diff
@@ -0,0 +1,46 @@
...
+= Azure CLI
+
+The Azure CLI commandlet allows to install and use https://github.com/Azure/azure-cli[Azure CLI]. Currently, the command works only for Windows.
+
+*ATTENTION:*
+Currently this feature is new and therefore experimental.
+It may change in incompatible ways in the next releases until we reach a stable state.
+We hope that all is working fine for you.
+However, do not expect everything to work out of the box.
+In case you are facing issues (e.g. network problems with Cisco AnyConnect, etc.) please give us feedback so we can improve.
```
hohwille, line 13: It seems as if this sentence always gets copied from docker to other commandlets even though it does not make any sense here. Only Docker/Rancher have this VPN issue.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+  exit
+fi
+
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+function doSetup() {
+  if doIsAzureInstalled
```
hohwille, line 18: if the command to test this is only as simple as `command -v az &> /dev/null` you could directly have it here instead of adding an extra function with 9 lines of code that does nothing else.
However, nothing wrong to do it this way.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+  fi
+}
+
+# modify a filepath for powershell path syntax e.g. /c/... -> c:/...
+function modifyPathStructure(){
+  modifiedPath=${DEVON_DOWNLOAD_DIR}
+  partition=$(echo "${modifiedPath}" | cut -f2 -d'/'):'/' # -> <partition>:/
+  modifiedPath=${partition}$(echo "${modifiedPath}" | cut -f3- -d'/')
+  modifiedPath="${modifiedPath////\\}"
+}
```
hohwille, line 51: here we should discuss as a team. Others had the same issue for git-bash and also we had solved this for cygwin years ago. IMHO we should have a function `doConvertPathToOs` or something that could solve this issue once and for all (different OS, shells).

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+  if command -v az &> /dev/null
+  then 
+    return
+  else
+    return 255
+  fi
+}
+
+# Check if Powershell is installed
+function doIsPwshInstalled() {
```
hohwille, line 103: official shortcut for PowerShell from Microsoft is just PS. However, if we want readability we should simply write `doIsPowershellInstalled` and stop saving 6 chars. You could even simply do the command in the if condition directly instead of wrapping it into a function as `command -v powershell` is at least for me already fully readable to understand what is happening...

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+  echo
+  echo "Arguments:"
+  echo " setup                    install Azure CLI on your machine (global installation)."
+  echo " <<args>>                 call Azure CLI with the specified arguments. Call az --help for details or use Azure CLI directly as preferred."
+  echo
+;;
+
+"setup" | "s")
+  doEcho "Installing Azure CLI..."
+  doSetup
```
hohwille, line 132: you should pass the arguments (e.g. "silent" option) like in other commandlets as you currently losing this here.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+;;
+
+"")
+if ! doIsAzureInstalled
+then
+  doEcho "Installing Azure CLI..."
+  doSetup
+else doRun "${@}"
+fi
```
hohwille, line 141: why not simply calling `doSetup silent`?
Why are you calling `doRun` only if already installed? Because installation will cause reboot anyway?
Why are you passing arguments here if the argument is empty (you are matching for "" in this case)?

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+;;
+
+*)
+if doIsAzureInstalled
+then
+  doRun "${@}"
+else doEcho "Unknown argument "
+fi
```
hohwille, line 149: similar to the block above:
why dont we simply first call `doSetup "silent"` and then directly follow with `doRun`?
The `else` case does not make sense to me: You do echo "Unknown argument " without any argument and then exit without error even though it seems this was an error situation...

hohwille (changes requested):
@alfeilex Thanks for this PR. You have done a very good job and it is more or less ready to be merged. Some of my comments are rather picky and are just nice-to-have hints for you. I would have merged right away but the last 2 comments seem to require a little improvement.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+  if doIsPwshInstalled
+  then
+    powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${modifiedPath}\az-${version}-windows.msi /QB-!'" &> /dev/null || exit_code=$?
+    if [[ -z "${exit_code}" ]]
+    then 
+      setConfigDir
+      doEcho "Azure CLI has been installed and is ready to use. \n\n******** IMPORTANT ******** \nRestart the console to use Azure CLI regular." 
+    else
+      read -r -p "You must run the installation with administration rights. Would you like to start the installation again? (Y/n)" answer 
+      case ${answer:0:1} in
```
hohwille, line 65: there already is a `doAskToContinue` function available from `functions`.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+  modifiedPath=${partition}$(echo "${modifiedPath}" | cut -f3- -d'/')
+  modifiedPath="${modifiedPath////\\}"
+}
+
+# Install Azure CLI for Windows via Powershell Command
+function installOnWindows(){ 
+  if doIsPwshInstalled
+  then
+    powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${modifiedPath}\az-${version}-windows.msi /QB-!'" &> /dev/null || exit_code=$?
+    if [[ -z "${exit_code}" ]]
```
hohwille: you do not need an extra variable. The common way to do this is just doing
```
<your command>
if [ "${?}" = 0 ]
```

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+    then 
+      setConfigDir
+      doEcho "Azure CLI has been installed and is ready to use. \n\n******** IMPORTANT ******** \nRestart the console to use Azure CLI regular." 
+    else
+      read -r -p "You must run the installation with administration rights. Would you like to start the installation again? (Y/n)" answer 
+      case ${answer:0:1} in 
+        y|Y )
+          doEcho "\nInstalling Azure CLI for Windows..."
+          exit_code=""
```
hohwille, line 68: See my comment above. This is causing confusion and pointless code. Variables should further be local where ever possible to avoid such nasty side-effects.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+          installOnWindows
```
hohwille, line 69: Assuming the user decides to rerun the installation after it failed due to the lack of admin rights.
This will not change anything so the same error will happen as it runs again without admin rights, correct?
If yes, this entire question and repeating does not make sense to me.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+        ;;
+        * )
+          doEcho "The installation has been canceled."
+        ;;
+      esac
+    fi
+  else doEcho "Powershell must be installed on your system to continue the installation."
```
hohwille, line 76: Doesn't windows 10 and 11 always come with PowerShell preinstalled?

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,151 @@
...
+# Set directory for azure cli configuration file
+function setConfigDir() {
+  azureconf_dir="${DEVON_IDE_HOME}/conf/.azure"
+  azurecli_export="export AZURE_CONFIG_DIR=${azureconf_dir}"
+  if ! grep -q "${azurecli_export}" "${DEVON_IDE_HOME}/conf/devon.properties"
+    then
+      doRunCommand "${azurecli_export}"
+      echo -e "\n${azurecli_export}" >> "${DEVON_IDE_HOME}/conf/devon.properties"
+      doEcho "Location of Azure's configuration file is set to ${azureconf_dir}"
+  fi
```
hohwille, line 90: nice solution.

scripts/src/main/resources/scripts/command/az
hohwille: as @CREITZ25 has figured out we can always use `cygpath` in bash on windows as both cygwin and git-bash do have this tool preinstalled. This will work more reliably than any custom replacement hacks.

documentation/az.asciidoc
alfeilex: I have now removed only the last sentence.

scripts/src/main/resources/scripts/command/az
alfeilex: > as @CREITZ25 has figured out we can always use `cygpath` in bash on windows as both cygwin and git-bash do have this tool preinstalled. This will work more reliably than any custom replacement hacks.

That's really cool! Thank you.

scripts/src/main/resources/scripts/command/az
alfeilex: That's right. I had added the query to be on the safe side. I have removed it now.

scripts/src/main/resources/scripts/command/az
alfeilex: If I call `doRun` without the query then the command `az` in the `doRun` function causes the installation process to restart. 

I removed the case statement with "" and modified the `doRun` function.

scripts/src/main/resources/scripts/command/az
alfeilex: The variable really makes no sense.

scripts/src/main/resources/scripts/command/az
alfeilex: Credits to @tobka777

scripts/src/main/resources/scripts/command/az
alfeilex:
```suggestion
function doRun() {
  doSetup silent
  if command -v az &> /dev/null
  then
    doEcho "Running: Azure CLI ${*}"
    setConfigDir
    az "${@}"
  fi
}
```

alfeilex:
@hohwille I have applied the changes. The code is now more compact. Thank you for the comments.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,104 @@
...
+    if [ "${1}" != "silent" ] && ! doIsQuiet
+    then
+      doEcho "Azure CLI is already installed at $(command -v az)"
+    fi
+  else
+    doRequireNotBatch
+    if doIsWindows
+    then
+      # Get leatest release
+      if [ -z "${AZURECLI_VERSION}" ]
```
hohwille, line 29: I did not notice this before but as the commandlet is called `az` and also the installed command from MS Azure is called `az` we should use `AZ_VERSION` instead of `AZURECLI_VERSION`. The documentation simply tells that the version is configured via `«TOOL»_VERSION` so the user has to guess what `«TOOL»` is. The natural expecation from `az` would be `AZ` rather than `AZURECLI`.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,104 @@
...
+      then
+        doEcho "Getting latest release..."
+        AZURECLI_VERSION=$(curl -s https://api.github.com/repos/Azure/azure-cli/releases/latest | awk -F ":" '/tag_name/ {print $2}'| awk -F "\"" '{print $2}' | awk -F "-" '{print $3}')
+      fi
+      local version="${AZURECLI_VERSION}"
+      doDownload "-" "" "az" "${version}"
```
hohwille, line 35: actually there is no additional use in the local variable `version`. You can simply pass `${AZURECLI_VERSION}` or better `${AZ_VERSION}` directly omitting this local variable.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,104 @@
...
+      installOnWindows
+    else
+      doFail "Sorry, Azure CLI installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
+    fi
+  fi
+}
+
+# Install Azure CLI for Windows via Powershell Command
+function installOnWindows() { 
+  winPath=$(cygpath -w "${DEVON_DOWNLOAD_DIR}")
```
hohwille, line 47: this is a global variable. Better do
```
local winPath
winPath=$(cygpath -w "${DEVON_DOWNLOAD_DIR}")
```

As the variable is only used in one place, you can even omit the variable.
However, from a documentation point of view the variable makes it more readable instead of putting `$(cygpath -w "${DEVON_DOWNLOAD_DIR}")` directly instead of `${winPath}` so it is fine to keep the variable.

scripts/src/main/resources/scripts/command/az
hohwille, line 35: Further, `doDownload` by default downloads and installs to the `software/«TOOL»` folder that is undesired here.
Analog to @CREITZ25 you should use a tool folder inside `update/install` instead.

scripts/src/main/resources/scripts/command/az
```diff
@@ -0,0 +1,104 @@
...
+}
+    
+# Set directory for azure cli configuration file
+function setConfigDir() {
+  azureconf_dir="${DEVON_IDE_HOME}/conf/.azure"
+  azurecli_export="export AZURE_CONFIG_DIR=${azureconf_dir}"
+  if ! grep -q "${azurecli_export}" "${DEVON_IDE_HOME}/conf/devon.properties"
+    then
+      doRunCommand "${azurecli_export}"
+      echo -e "\n${azurecli_export}" >> "${DEVON_IDE_HOME}/conf/devon.properties"
```
hohwille, line 71: very nice with leading newline to prevent errors if the file is not terminated with newline. :)

scripts/src/main/resources/scripts/command/az
hohwille: > Assuming the user decides to rerun the installation after it failed due to the lack of admin rights.
This will not change anything so the same error will happen as it runs again without admin rights, correct?

I still do not see this addressed. Do I miss something? What is the point in rerunning the installation without any change again if exit code was not 0 (succcess)? And the assumption that the admin permissions were causing the error might also be wrong. So if admin rights are required, the powershell process needs to be launched in a specific way granting admin rights and if that is done, there is IMHO no sense in retrying.

Or is this meant as following:
The PS is already launched with admin rights what renders a UAC popup dialog. If the user clicks on cancel you are telling that admin rights are required and rerun it so the user has to accept UAC. Is that what you intended?
Then it might be OK but shouldn't the user rerun the `setup` if he accidentally clicked cancel?
If he for whatever reason intended to cancel, will he be annoyed by devonfw-ide asking him forever, making it hard to abort?

scripts/src/main/resources/scripts/command/az
alfeilex: I assume that you mean `doInstall` instead of `doDownload`. I have changed the command and adressed myself to @CREITZ25 implementation of aws command.

hohwille (approved):
@alfeilex Excellent. 👍 Thanks for your changes. Now everything looks perfect to me.
Ready for merge.

PR #763, opened 2022-05-24
Updated devonfw-ide introduction

suprishi:


hohwille:
Actually looking at your change you seem to want to address that the devonfw.com website is giving you a better introduction about devonfw-ide on the fist click.
Did you ever read the documentation as PDF?
https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-doc/2022.04.002/devonfw-ide-doc-2022.04.002.pdf

IMHO the problem still is the devonfw.com website itself and I had long talks about this with Malte, Jan-Gerrit, Fabian, and many others that we need yet another redesign for the website to present the contents properly to the end-user.

hohwille:
> It is not another Eclipse.

For the record: If we need documentation to tell what something is not, we actually did something wrong earlier. We have to figure out, where we are giving a wrong expectation. Then we need to correct and question if we can resolve this at the root of the problem.

hohwille:
Finally closing this stale PR.

Issue #777, opened 2022-06-07
Support for colored output


hohwille:
As a devonfw-ide user, I want to have colored log output so that I can easier distinguish between success, error and warning message, etc.

* warnings are outputted yellow
* steps cyan
* sucess messages green
* error messages red

PR #778, opened 2022-06-07
#777: Support for colored output - merge of PR #655

hohwille:


hohwille:
So far an internal convention was that all functions in the `functions` file have the common naming convention starting with `do`.
Now, `doWarning` had to be changed to `logWarn`. In this case, I do not see why we had to do such change on the "API" as it affects all commandlets and I have to explain the team to adopt all their scripts under development while this PR is on its way to get merged.
I fully see the point that `doEcho` has to be changed for outputting success messages but IMHO this is the only required change and we could keep API changes minimal by simply adding a `doSuccess` function avoiding any other "API changes" and causing incompatibilities and bugs in merges.

hohwille:
I have tested the feature on Windows with the following constellations:
* CMD
* git-bash
* Powershell

All worked fine and gave me colored output. However, only where applicable. So e.g. in `devon help` you do not see any difference (no colors).

On observation I made was the following:
```
devon foo
Unknown command foo
```

The outputted message is an error and it is not printed in red.
The reason for that, is that `devon` is a command/script that gets globally installed once on your system and is not yet updated.
However, when we do a release it will be updated.
The problem is that due to the "API changes" discussed above, it will break other existing IDE installations if they are not updated as the function `logError` will be undefined in older versions of devonfw-ide.

Therefore, I will try some further reworking and rethinking:
* How can we keep the "API changes" as minimal as possible?
* How can we change `devon` command itself in a way that it still works with older versions of `functions` from previous IDE installations?
* Will the coloring also work properly on MacOS and Linux?
```

**Comment by hohwille** [comment]
Current build of this PR hangs in an infinity loop:
```
Please check the above errors, resolve them and try again.
Exit code was 1
1
WARNING: Project is already on version devon4j:2022.04.001. No migrations available to update.
Failed to run command: java -cp '/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/devonfw-ide-configurator-2022.04.003-SNAPSHOT.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-1.1.4.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-api-1.1.4.jar' com.devonfw.tools.ide.migrator.Migrator single
******** ATTENTION ********
Failed to run command java
Exit code was 1
1
WARNING: Project is already on version devon4j:2022.04.001. No migrations available to update.
Failed to run command: java -cp '/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/devonfw-ide-configurator-2022.04.003-SNAPSHOT.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-1.1.4.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-api-1.1.4.jar' com.devonfw.tools.ide.migrator.Migrator single
******** ATTENTION ********
Failed to run command java
Exit code was 1
1
WARNING: Project is already on version devon4j:2022.04.001. No migrations available to update.
Failed to run command: java -cp '/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/devonfw-ide-configurator-2022.04.003-SNAPSHOT.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-1.1.4.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-api-1.1.4.jar' com.devonfw.tools.ide.migrator.Migrator single
******** ATTENTION ********
Failed to run command java
Exit code was 1
1
WARNING: Project is already on version devon4j:2022.04.001. No migrations available to update.
Failed to run command: java -cp '/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/devonfw-ide-configurator-2022.04.003-SNAPSHOT.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-1.1.4.jar:/home/runner/work/ide/ide/scripts/target/integration-test/test-setup/scripts/lib/javax.json-api-1.1.4.jar' com.devonfw.tools.ide.migrator.Migrator single
******** ATTENTION ********
Failed to run command java
Exit code was 1
```

What change of this PR broke this feature?

**Comment by maybeec** [comment]
let's keep it KISS.
I would simply say if 
```
devon foo
Unknown command foo
```
is not properly colored - don't care... if at least the rest is, even better.

About the incompatibility issues. For sure, the naming with "doX" is quite non-speaking, therefore, I changed it without having in mind the need for a stable API. So as of this restriction, we clearly can change it again to be more stable, which is totally fine for me and should not be a show breaker.

File: scripts/src/main/resources/scripts/functions [hunk 1]
```diff
@@ -19,27 +19,26 @@ DEVON_DOWNLOAD_DIR="${DEVON_HOME_DIR}/Downloads/devonfw-ide"
...
   if [ -n "${2}" ]
   then
-    echo "Exit code was ${2}"
+    doError "Exit code was ${2}"
     if [ "${2}" = "0" ]
     then
       exit 1
     fi
-    exit "${2}"
+    doError "${2}"
```
hohwille, line 30: this is what broke the build. The exit was replaced with logging causing an infinity loop.

hohwille: fixed.

hohwille:
See #779 - but fix is also on its way and can be merged today.

Issue #781, opened 2022-06-07
Support for dotnet


hohwille:
As a devonfw-ide user, I want to install and use the DotNet SDK so that my project with .NET/C# technology can make use of devonfw-ide.

hohwille:
See also #788

PR #782, opened 2022-06-07
#781: DotNet support

hohwille:
I took over the stale PR #292 that was about to rotten to death.
This PR is therefore implementing #781 thanks to the work and basis created with PR #292.
I have tested this and it seems to work.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -13,4 +13,4 @@ jobs:
         with:
           java-version: '11'
       - name: Build project with Maven
-        run: mvn install
+        run: mvn -B install
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
         with:
           java-version: '11'
       - name: Build project with Maven
-        run: mvn install
+        run: mvn -B install
       - name: Deploy to OSSRH nexus
         env:
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,1 +1,1 @@
--Drevision=2022.04.002
+-Drevision=2022.04.003-SNAPSHOT
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,1 +1,1 @@
--Drevision=2022.04.002-SNAPSHOT
+-Drevision=2022.04.002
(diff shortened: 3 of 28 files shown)


hohwille:
The only room for improvement is the somewhat strange Template stuff in combination with `devon dotnet create` command.

Running `devon dotnet setup` a second time gives me this output:
```
Version 6.0.300 of dotnet is already installed.
Running: dotnet new -i devon4net.WebApi.Template
The following template packages will be installed:
   devon4net.WebApi.Template

Error: File C:\Users\hohwille\.templateengine\packages\Devon4Net.WebAPI.Template.6.0.3.nupkg already exists.
devon4net.WebApi.Template could not be installed, download failed.
6.0.300
Succeeded to verify installation of dotnet ('/d/projects/devonfw/software/dotnet/dotnet')
```

It continues even after the error so it does not really hurt but we could improve here to prevent this confusing error.

Further I go to `workspaces/test` and do `devon dotnet create` again in that folder and get:
```
Running: dotnet new Devon4NetAPI
Creating this template will make changes to existing files:
  Overwrite   ./assets/devonfw.png
  Overwrite   ./assets/LICENSE
...
...
  Overwrite   ./Templates/WebAPI/Devon4Net.Application.WebAPI.Implementation/Devon4Net.Application.WebAPI.Implementation

  Overwrite   ./Templates/WebAPI/Devon4Net.Application.WebAPI.Implementation/Devon4Net.Application.WebAPI.Implementation.csproj
  Overwrite   ./devon4net.WebAPI.sln

Rerun the command and pass --force to accept and create.
```

This is also not what I would expect as a user:
When I create a new project from a template, I want to have this created in my current workspace and give it a specific name for my project. This is how it already behaves for other stacks like Java or Angular.

hohwille:
So I meanwhile fixed the first issue that setup checks if templates are already installed.
The project creation remains odd and I have no clue how .NET/C# ecosystem is working.
I still would expect that I have to provide additional information to `devon dotnet create` command like a package or project name. Then I would expect that the specified project will get created inside my current workspace named as the specified project rather than in `$DEVON_IDE_HOME/Templates`. Without any feedback from devon4net community, I can not proceed any further here. So I might just merge this and let it as is hoping that some day the users will create an issue with more information about this.

hohwille:
> Hi @hohwille your are right. The previous email and commits are from my Capgemini account that cannot be used nowadays. All commits (@jmsanmu and [jose-manuel.sanchez-munoz@capgemini.com](mailto:jose-manuel.sanchez-munoz@capgemini.com)) are mine. Kind Regards.

Thanks for confirmation. I therefore merge as is since we cannot resolve the red X from @CLAassistant.

hohwille:
So I created #788 for the last remaining obstacle and can merge this now.

Issue #746, opened 2022-05-16
Fix MAC OS bug during Azure CLI integration


Bene90:
See: https://github.com/Azure/azure-cli/issues/22345

PR #783, opened 2022-06-08
#746 Azure CLI for Mac Integration

alfeilex:
This is a PR for the Azure CLI for Mac integration. Azure CLI for Mac can only be installed globally via homebrew, so this installation is local and is stored in the IDE in software/azure. 

The installation procedure is also possible for Windows, but this requires Python to be installed. On Mac, Python is already pre-installed. Either we do without the installation for Mac or we make this compromise. I am waiting for feedback.

PR refers to #746

### DIFF
--- a/scripts/src/main/resources/scripts/command/az
+++ b/scripts/src/main/resources/scripts/command/az
@@ -90,7 +90,7 @@ function doRun() {
     setConfigDir
     if doIsMacOs
     then
-      ${AZ_MAC_DIR}/bin/az "${@}"
+      "${AZ_MAC_DIR}"/bin/az "${@}"
     else
       az "${@}"
     fi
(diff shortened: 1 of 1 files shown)


alfeilex:
The only way to install Azure CLI locally on Windows is to install it using Python 3.X with venv and pip as on MacOS. Also, the provided [bash script](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=script) uses a Python script (https://azurecliprod.blob.core.windows.net/install.py) to install Azure CLI. Package managers like winget or chocolatey use the MSI installation URL for installation.

Further I could not find any other URLS with ZIP extensions and I could not anything with the unpacked data of the MSI file. For a local installation for all operation systems I only see the possibility to do it via Python venv. However, Python 3.X would be required for this.

alfeilex:
The PR is closed and a new issue #797 is created

Issue #751, opened 2022-05-17
Documentation of functions


hohwille:
As a new devonfw-ide developer, I want to get some documentation about the `functions` so that I can understand easier which functions are available, what they do and how to use them without digging and reverse-engineering the code.

The [functions](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/functions) are complex and long (>1000 lines). It is not properly documented and adding extensive documentation here might have some downsides as the script has to be parsed by bash all the time and extensive comments would massively increase the file length and size.
Therefore we instead want to add a `documentation/functions.asciidoc` file as documentation to functions.
Here we should have a heading for each function with a description of the function (e.g. `doInstall`) in a common format (e.g. using a table or numbered list points for parameters, etc.).
As this is quite some work, it would IMHO be a good idea to first start with a single function as example and then introduce the result to the entire team to get feedback before documenting all other functions in the same way.
This story is slightly related to #729

hohwille:
See #729

PR #784, opened 2022-06-09
#751 Documentation of functions

CREITZ25:
#751 Documentation of functions

### DIFF
--- a/documentation/functions.asciidoc
+++ b/documentation/functions.asciidoc
@@ -35,7 +35,7 @@ The function doIsIgnoredFolder checks whether the directory that is passed to it
 The doEcho function prints the arguments passed to it if the variable quiet is not set. Backslash-escaped characters, for example "\n" for new line, are also allowed in the message.
 
 = doInfo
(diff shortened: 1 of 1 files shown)


hohwille (commented):
@CREITZ25 thanks for this PR and the documentation. This is looking good already.
However, some remarks:
* We could group several functions together in a section making the functions themselves to subsections. E.g. all functions for output and input belong together (`doEcho`, `doInfo`, `doError`, `doFail`, `doAskToContinue`), all functions to test some flags could be grouped (`doIsDebug`, `doIsBatch`, etc.), all functions that check the Operating System like `doIsWindows` and `doIsMacOs` make a group, then all functions regarding download and installation of software make a group, etc.
* This looks like WIP (Work in Progress) as many functions are not yet documented but just listed. Shall I wait till this is complete before I do a detailed review? Then use the Draft-State for your Github-PR so I can notice this.
* I will do the detailed review when this is ready. There will be some small points (e.g. `doLogo` does output the logo via echo and not return it, etc.). However, I see this documentation is perfectly on its way and will be very helpful in the future. As always we can also merge it and continuously improve it in our agile way of work...

PR #785, opened 2022-06-09
bugfix terraform and helm

CREITZ25:
when terraform and helm are used as proxy commandlets in devonfw-ide the arguments are not properly passed and as a result it is not working:
```
# devon helm env
The Kubernetes package manager

Common actions for Helm:

- helm search:    search for charts
- helm pull:      download a chart to your local directory to view
- helm install:   upload the chart to Kubernetes
- helm list:      list releases of charts
...
```
so the `env` parameter is not passed to `helm`. Same problem exists for `terraform`.


hohwille (approved):
@CREITZ25 Excellent. Thanks for finding and fixing. 👍

Issue #692, opened 2022-01-24
Improve Windows Terminal with gitbash


JokeFrels:
As a devonfw-ide user on windows, I want to make use of microsoft Terminal app so that I get all my shell sessions as tabs in one window.
Therefore devonfw-ide should notice if Terminal app is installed on windows and if so instead of installing the regular context menu entries like "open devonfw CMD here" install something like this but with full devonfw-ide integration setting up the environment:
https://github.com/lextm/windowsterminal-shell
So as a result in Windows Explorer I would have a menu "Open devonfw Terminal here" with the sub-menu options:
* git-bash  ([x] open new tab, [ ] getting variable)

14.06:
Spellchecker bug can be ignored.
Merge must be done.

PR #787, opened 2022-06-09
Improve Windows Terminal with gitbash #692

CREITZ25:


hohwille (approved):
@CREITZ25 Excellent. Thanks for solving this also for Git-Bash. Awesome 👍

hohwille (approved):
@CREITZ25 great that you could also solve it for git-bash. All looking good to me.👍

Issue #797, opened 2022-06-14
Azure CLI with Python


tobka777:
Azure CLI (see issue #712) should be able to be installed locally. For this, python must be installed as a dependency if not already installed. The local installation should be performed in the same way for all operating systems.

Acceptance criteria:

- [x] The new commandlet allows to install and launch Azure CLI from the CLI
- [x] Azure CLI is installed with pip 
- [x] Azure CLI is executed with python (use new python commandlet #713)
- [x] The commandlet is a wrapper for Azure CLI
- [x] Azure CLI (latest version) can be installed automatically (during setup via settings) as well as via the commandlet call - independently of the used OS plattform 
- [x] Azure CLI is integrated in a way that complies with the multi-tenancy of the devon IDE - this implies that configuration is stored per IDE instance and not globaly in users home directory.
- [x] During the installation process I get information about the current status of the installation
- [x] Azure CLI can be installed and used together with any of the devon IDE standard tools without negative interference
- [x] Add licence documentation
- [x] The new commandlet is documented on the website
- [ ] Tested under MacOS, Linux, Windows

alfeilex:
- [ ] Azure CLI must be installed locally via Python

hohwille:
See also issue #712

Akuhana:
The Azure CLI package can now be installed via command `devon pip install azure-cli` (see issue #841). Nevertheless, further testing on different operating systems is still required.

PR #801, opened 2022-06-22
#797 Azure CLI with Python

alfeilex:


alfeilex:
The command works on windows and linux. On linux I had to add spellchecker ignores to accept the missing double quotes (see line 48-52) because otherwise commands like `devon az config get` lead to errors. 

On Mac there is still a bug concerning SSL, which is still being fixed in the Python command.

PR #798, opened 2022-06-20
Switching message on success to "Successfully <past>"

ThomasRot:
Just a minor grammar fix as to my knowledge "succeed to <infinitiv>" does not exist..

PR #802, opened 2022-06-23
#798: English grammar fix for success message

hohwille:
Alternative fix for PR #798


hohwille:
Now produces output like this:
```
Success: verify installation of node.js ('/d/projects/devonfw/software/node/node.exe')
```
IMHO simple fix avoiding other problems when changing the time forms from present to past.

PR #812, opened 2022-06-27
added doInstallWithPackageManager to functions

### LINKED ISSUE #811: add function for installing with different package managers
As a devonfw-ide user on Linux I want to install software without worrying if my package manager is supported. For that a function has to be added that automatically installs with the correct package manager for a given linux distribution. 

This function needs to: 
- find the installed package manager 
- install the given packag(es)
- return an error if no corresponding package could be found
- return an error if the installed package manager is not supported

BizarreBits:
function takes one parameter and tries to install the given packages.  If packages are not found it will result in an error, same as when no supported package manager is found.
Closes #811


scripts/src/main/resources/scripts/functions
```diff
@@ -741,6 +741,42 @@ function doUnzip() {
   doRunCommand "unzip -qn '${1}' -d '${target_dir}'"
 }
 
+function doInstallWithPackageManager() {
+  pckmngers=("apt-get" "yum" "pacman" "emerge" "zypp" "apk" "dnf" "brew")
+  installString=${1}
```
hohwille, line 746: Sorry, it seems I was not clear enough about this feature/story.
The `installString` may differ between packagemanager and therefore this will not properly work for all situations.

My expectation was that I can provide an array of linux/OS specific commands to this function.
Then this function will iterate all of them and always extract the first arg of the command (the packagemanager).
It will check if that command exists and if yes call it prefixed by `sudo`.

Example usage for Python:
```
doInstallWithPackageManager "apt-get install -y make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev" "yum -y groupinstall 'Development Tools && yum -y install gcc openssl-devel bzip2-devel libffi-devel" "zypp install automake fdupes gcc gcc-c++ gcc-fortran gdbm-devel gettext-tools gmp-devel intltool libbz2-devel libexpat-devel libffi-devel libnsl-devel lzma-devel make ncurses-devel netcfg openssl-devel pkgconfig readline-devel sqlite-devel xz zlib-devel" 
```

Or simpler case for unzip (you could change doUnzip using this in the same PR):
```
doInstallWithPackageManager "apt-get install unzip" "yum install unzip" "pacman -S unzip" "emerge install unzip" "zypp install unzip" "apk add unzip" "dnf install unzip" "brew install unzip"
```

hohwille (changes requested):
@BizarreBits thanks for this PR.
I have added a review comment to clarify the requirement better.
Could you have a look and rework?

hohwille:
DoD?

> Running test #2: test-shellcheck
In scripts/functions line 746:
  installStrings="${@}"
  ^------------^ SC2034: installStrings appears unused. Verify use (or export if used externally).
                 ^----^ SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.
In scripts/functions line 763:
    if [ ${pckmnger} = ${first} ] || [ ${pckmnger} = ${second} ]
         ^---------^ SC2086: Double quote to prevent globbing and word splitting.
                       ^------^ SC2086: Double quote to prevent globbing and word splitting.
                                       ^---------^ SC2086: Double quote to prevent globbing and word splitting.
                                                     ^-------^ SC2086: Double quote to prevent globbing and word splitting.
Did you mean: 
    if [ "${pckmnger}" = "${first}" ] || [ "${pckmnger}" = "${second}" ]

hohwille (changes requested):
@BizarreBits just for the record I note what we discussed in our review meeting:

* avoid explicitly hardcoding pckmngers here
* instead start with the loop in 755 (`for installString in "${@}"`)
* only extract the first arg using `local first=${installString/ */}`
* then check if that package manager exists (`if command -v "${t}" > /dev/null`)
* and in that case call the actuall command with sudo (`doRunCommand "sudo ${installString}"`)
* and then exit this function (`return`)
* Otherwise after the look simply fail (e.g. `doFail "No supported package manager found on your system.\nInstallation failed: ${1}"`).

scripts/src/main/resources/scripts/functions
```diff
@@ -741,6 +741,23 @@ function doUnzip() {
...
+function doInstallWithPackageManager() {
+  
+  for installString in "${@}"
+  do
+    first="${installString/ */}"
+    
+    if command -v "${first}" > /dev/null
+    then
+      doEcho "Installing: ${installString}"
+      doRunCommand sudo "${installString}"
```
hohwille, line 753: `doRunCommand` expected the command as first argument, second one is comment so this is actually wrong.

hohwille:
Build failing:
```
In scripts/functions line 93:
  [ -n "${force}" ]
        ^------^ SC2154: force is referenced but not assigned.
In scripts/functions line 98:
  [ -n "${batch}" ]
        ^------^ SC2154: batch is referenced but not assigned (for output from commands, use "$(batch ...)" ).
In scripts/functions line 110:
  [ -n "${quiet}" ]
        ^------^ SC2154: quiet is referenced but not assigned.
In scripts/functions line 115:
  [ -n "${debug}" ] && [ -z "${quiet}" ]
        ^------^ SC2154: debug is referenced but not assigned.
In scripts/functions line 1514:
}
^-- SC1089: Parsing stopped here. Is this keyword correctly matched up?
```

As you did not change these parts of functions I can only assume that github actions now install a newer version of shellcheck and that behaves more picky than the previous version. Actually this is bad style - updates shall not break builds worldwide that used to work before. However, lets not mess with shellcheck or GHA guys and just silently fix (make as false positive?).

hohwille (approved):
@BizarreBits Thanks for your rework. I simply completed this as it was already 99% ready.
Ready for merge.

hohwille:
```
In scripts/functions line 1516:
}
^-- SC1089: Parsing stopped here. Is this keyword correctly matched up?
```

Maybe I messed something up - will check and fix...

hohwille:
https://github.com/devonfw/ide/commit/021fcefccf8991a01387a70c4114c8bc043d8220

I will fix this and then we can merge...

scripts/src/main/resources/scripts/functions
hohwille: now all fine. Thanks.

scripts/src/main/resources/scripts/functions
hohwille: I fixed it so also resolved.

hohwille:
Now GitGuardian check is hanging for over 10 minutes. I will wait for another 30 minutes and then merge anyhow if still stuck...

PR #816, opened 2022-06-28
Feature/aws

CREITZ25:


hohwille (approved):
@CREITZ25 thanks for the PR. Excellent. Same fix as with `helm` and `terraform` - now we should have it all correct. Will merge this asap.

PR #822, opened 2022-07-01
#794: improve installation of rancher desktop

### LINKED ISSUE #794: Installation Process of Rancher Desktop and VPNKit
*Problem*

If the user doesn't have WSL enabled inside Windows, Rancher Desktop will require a restart at the end of its installation to enable WSL.
The VPNKit needs WSL to be enabled, or can't be installed otherwise. 

*Solution*

Therefore the VPNKit has to be installed after Rancher Desktop finished its installation and after a possible restart. 
A RunOnce Registry Key could be a comfortable solution handling the restart process (git-bash with admin rights needs to be started for that purpose).
An easier option would be to ask the user to rerun 'devon docker setup' after the restart and modifying the code acordingly.

cinnamon-coder-hub:
fixed #794 

- Changed installation order
- Included restart case for enabling wsl in installation script via setting a startup file without admin rights to finish installation after reboot (instead of setting a RunOnce registry key with admin rights that doesn't necessarily work)
- Via 'devon docker setup' now ability to check if wsl-vpnkit is installed and to catch up on its installation
- Some minor additions


scripts/src/main/resources/scripts/command/docker
```diff
@@ -41,7 +51,7 @@ function doSetup() {
         local version="${RANCHER_VERSION}"
   	  	if [ -z "${version}" ]
         then 
-          version="1.3.0"
+          version="1.4.1"
```
hohwille, line 54: great. 👍

scripts/src/main/resources/scripts/command/docker
```diff
@@ -150,6 +171,39 @@ function doIsRancherInstalled() {
     fi
 }
 
+
+function doIsVPNKitInstalled() {
+  # echo $result=$(wsl -l -v | % {$_.replace("`0", '')} | Select-String 'wsl-vpnkit')
+  result=$(powershell.exe -command "wsl -l | Where {\$_.Replace(\"\`0\",\"\") -match '^wsl-vpnkit'}" | tr -d '\0' )
```
hohwille, line 177: why do we need powershell for this?
Simply use bash...

scripts/src/main/resources/scripts/command/docker
```diff
@@ -150,6 +171,39 @@ function doIsRancherInstalled() {
...
+    # echo "start \"\" "C:\Program Files\Git\git-bash.exe" -c "cd /c/projects/my-project/ide-devon && devon docker setup  && /usr/bin/bash --login -i""
+    echo '(goto) 2>nul & del "%~f0"' 
+  } > VPNKitInstaller.bat
+  {
+    echo '@ECHO OFF'
+    echo 'move VPNKitInstaller.bat "%APPDATA%\Microsoft\Windows\Start Menu\Programs\Startup\VPNKitInstaller.bat" >nul'
+    echo '(goto) 2>nul & del "%~f0"'
+  } > moveKitInst.bat
+
+  cmd.exe "/c moveKitInst.bat"
```
hohwille, line 201: no need for CMD to do this:
```
$ echo $APPDATA
C:\Users\hohwille\AppData\Roaming
```

scripts/src/main/resources/scripts/command/docker
```diff
@@ -162,17 +216,19 @@ function doInstallVpnkit() {
...
+    echo 'pushd $env:USERPROFILE | Out-Null'
+    echo 'wsl --import wsl-vpnkit $env:USERPROFILE\wsl-vpnkit wsl-vpnkit.tar.gz --version 2 | Out-Null'
+    echo 'popd | Out-Null'
+    echo 'wsl.exe -d wsl-vpnkit service wsl-vpnkit start | Out-Null' 
   } > vpnkit-setup.ps1
 
   #execute the powershell-script 
-  powershell.exe -Command ".\vpnkit-setup.ps1"
+  # shellcheck disable=2091
+  $(powershell.exe -Command ".\vpnkit-setup.ps1")
```
hohwille, line 231: I do not get the rationale for this change. `$()` syntax is used to get the output of an execution.
As you are not using the output (e.g. as in `variable="$(command)"`) you can simply omit this syntax, but why did you add it then?

scripts/src/main/resources/scripts/command/docker
```diff
@@ -150,6 +171,34 @@ function doIsRancherInstalled() {
     fi
 }
 
+
+function doIsVPNKitInstalled() {
```
hohwille, line 175: Just for the record: In `pascalCase` and `CamlCase` never use CAPITALIZED abbreviations like `VPN`.
For Java coding we had documented this as explicit convention here:
https://github.com/devonfw/devon4j/blob/master/documentation/coding-conventions.asciidoc#naming

For bash we did not write explicit coding conventions but it makes sense to follow this same rule(s) for all languages.

This is just a nice-to-have improvement you should consider for your next update. It will not block this PR from being approved and merged.

PR #795, opened 2022-06-10
Python integration for Windows

BizarreBits:
Complete Python Integration for Windows and Linux, with Pip included.

PR #824, opened 2022-07-12
#797: Azure CLI Python Integration

alfeilex:
Implementation of #797 (Azure CLI)


scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
-
-# autocompletion list
-if [ "${1}" = "shortlist" ]
-then
-  if [ -z "${2}" ]
-  then
-    echo "setup help"
-  fi
-  exit
-fi
```
hohwille, line 11: you should not remove this - it is required for auto-completion.

scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
-    winPath=$(cygpath -w "${path_to_package}")
-    powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${winPath}\az-${AZ_VERSION}-windows.msi /QB-!'" &> /dev/null
-    if [ "${?}" = 0 ]
-    then
-      setConfigDir
-      doSuccess "Azure CLI ist installed now. Great! \nTo get started you may need to restart your current shell."
-    else
-      doError "The setup was canceled or failed. Rerun the setup to install Azure CLI."
-    fi
+    AZ_FILE="${DEVON_IDE_HOME}/software/python/Scripts/az"
```
hohwille, line 39: Actually it is not so good to have internal dependencies and knowledge of python/pip inside this azure commandlet.
As @BizarreBits changed #795 and there is no `Scripts` folder anymore this will not work anymore.
Actually we should assume that everything will automatically be added to the `PATH` so you could directly call it.

Tricky would be to distinguish if the `az` installation comes from a global installation or from the local IDE.
I would suggest to use `command -v az` and then check that the result of this starts with `${DEVON_IDE_HOME}/software/`.

scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
-# Set directory for azure cli configuration file
-function setConfigDir() {
-  azureconf_dir="${DEVON_IDE_HOME}/conf/.azure"
-  azurecli_export="export AZURE_CONFIG_DIR=${azureconf_dir}"
-  if ! grep -q "${azurecli_export}" "${DEVON_IDE_HOME}/conf/devon.properties"
-    then
-      doRunCommand "${azurecli_export}"
-      echo -e "\n${azurecli_export}" >> "${DEVON_IDE_HOME}/conf/devon.properties"
-      doEcho "Location of Azure's configuration file is set to ${azureconf_dir}"
+    AZ_FILE="${DEVON_IDE_HOME}/software/python/bin/az"
```
hohwille, line 41: Same here. IMHO you can simply remove the OS switch and replace all with the approach I suggested above.

scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
 
 function doSetup() {
-  if command -v az &> /dev/null
+  setLocation
+  if [ "${1}" != "silent" ] || [ ! -f "${AZ_FILE}" ] 
   then
-    if [ "${1}" != "silent" ] && ! doIsQuiet
+    if [ -d "${PYTHON_HOME}" ]
```
hohwille, line 9: Actually you should even remove `setLocation` function and then do the check here as otherwise this is also not correct and we should have as little inter-dependencies between commandlets (here `az` and `python`).

scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
-    if [ "${1}" != "silent" ] && ! doIsQuiet
+    if [ -d "${PYTHON_HOME}" ] 
     then
-      doEcho "Azure CLI is already installed at $(command -v az)"
+      doDevonCommand python pip install azure-cli
+      setConfig
+    else
+      doDevonCommand python setup silent
+      doDevonCommand python pip install azure-cli
+      setConfig
```
hohwille, line 16: so your IF should now simply check if `az` is already installed locally in my IDE.
If yes there is nothing to install. Otherwise you should always do the installation in one way.
Please note that `doDevonCommand python pip install azure-cli` should always implicitly do `devon python setup silent` for you so you do not need to call that explicitly and you have no need for these two variants here before and after `else`.

scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
+  else
+    # shellcheck disable=SC2068
+    az $@
   fi
 }
 
-AZ_HOME="${DEVON_IDE_HOME}/updates/install/az"
-
 # CLI
+PYTHON_HOME="${DEVON_IDE_HOME}/software/python"
```
hohwille, line 59: can be removed

scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
-  echo " setup                    install Azure CLI on your machine (global installation)."
-  echo " <<args>>                 call Azure CLI with the specified arguments. Call az --help for details or use Azure CLI directly as preferred."
+  echo " setup                    setup Azure CLI (install and verify)"
+  echo 
   echo
 ;;
 
 "setup" | "s")
-  doSetup "${2}"
+  doSetup setup
```
hohwille, line 71: Please revert this change. It was perfectly correct before. See also other commandlets like here:
https://github.com/devonfw/ide/blob/71c62a97bc93bff4b302dc93d9d6e01ecd210943/scripts/src/main/resources/scripts/command/mvn#L369

Otherwise it is not possible for other commandlets to call setup with parameter `silent` here.

scripts/src/main/resources/scripts/command/az
```diff
@@ -1,106 +1,77 @@
...
 ;;
 
 *)
-  doRun "${@}"
+  doRun "${*}"
```
hohwille, line 75: I also do not see why you changed this here? `${@}` is perfectly correct here. Example from other commandlet:
https://github.com/devonfw/ide/blob/71c62a97bc93bff4b302dc93d9d6e01ecd210943/scripts/src/main/resources/scripts/command/mvn#L392

hohwille (changes requested):
@alfeilex Thanks for your rework to use pyhton and pip and ensure we have a local installation working accross all operating-systems. 👍 
As the python commandlet has been WIP and changed meanwhile, you have to update this PR.
I left you several review comments with the according hints and suggestions.
Please rework so we can get all merged for the upcomming release (hopefully by the end of this week).

scripts/src/main/resources/scripts/command/az
alfeilex: I use the `command -v az` option, but the problem is still that the local az script is not automatically added in the path.

alfeilex: The az script is still located at `software/python/Scripts`. Pip create this folder automatically during package installations. Either we need to add the script folder to the path or we try to use a target switch as described here? https://stackoverflow.com/a/19404371

alfeilex: > I use the `command -v az` option, but the problem is still that the local az script is not automatically added in the path. The behavior I assume is that after I run `devon`, `command -v az` shows me the local path. Currently the global path is output like `/c/Program Files (x86)/Microsoft SDKs/Azure/CLI2/wbin/az.`
> 
> Furthermore, it would be important that when Python is also installed with `doDevonCommand python pip install azure-cli,` the environment variables are updated before`pip install azure cli` is executed, otherwise the setup will abort.

Could be fixed with by adding `PATH=/c/projects/devonfw-ide-scripts-2022.04.003/software/python/Scripts:$PATH` to devon.properties

scripts/src/main/resources/scripts/command/az
```diff
@@ -10,88 +10,45 @@ then
...
-        AZ_VERSION=$(curl -s https://api.github.com/repos/Azure/azure-cli/releases/latest | awk -F ":" '/tag_name/ {print $2}'| awk -F "\"" '{print $2}' | awk -F "-" '{print $3}')
-      fi
-
-      doInstall "-" "${AZ_HOME}" "az" "${AZ_VERSION}"
-      if [ "${?}" = 0 ]
-      then
-        doPackageInstall "${AZ_HOME}"
-      fi
-    fi
+    doDevonCommand python
```
alfeilex, line 19: I changed the command so that it works for the latest python command. In the latest python command you can't run `devon python pip,` but you can run `pip` individually.

Further in the python command still need to add the script folder to the path.
`PATH=/c/projects/devonfw-ide-scripts-2022.04.003/software/python/Scripts:$PATH

Currently, if you run `devon az` when python is not installed, you have to restart the console again so that pip3 is recognized in the path.
`

scripts/src/main/resources/scripts/command/az
```diff
@@ -10,88 +10,45 @@ then
...
+    doRunCommand "pip3 install azure-cli"
+    setConfig
```
alfeilex, line 21: I changed the command so that it works for the latest python command. In the latest python command you can't run `devon python pip`, but you can run pip individually.

Further in the python command still need to add the script folder to the path.
`PATH=/c/projects/devonfw-ide-scripts-2022.04.003/software/python/Scripts:$PATH`

Currently, if you run `devon az` when python is not installed, you have to restart the console again so that pip3 is recognized in the path.
`

hohwille: > In the latest python command you can't run `devon python pip`. 

Indeed. Thanks for pointing this out and sorry for that. Seems that I broke this. I created #839 and will fix this on Monday. Should be no big deal. You can then change this back to a single `doDevonCommand` again.
BTW: If you want to to a dependent setup always do with `setup silent`. Example:
https://github.com/devonfw/ide/blob/f1d26c141dc923bb3ab8f754cba4754e8fcc5278/scripts/src/main/resources/scripts/command/mvn#L23

> Further in the python command still need to add the script folder to the path.

I still disagree to that approach. This will only cause added complexity and future bugs.

However, I just tested and see what you mean. By default pip is still installing to the Scripts folder. This is hardcoded somewhere inside pip and totally undesired and violating obvious standard and on all platforms other than Windows it uses the `bin` folder instead.
Anyhow, this is good as it pointed me to a solution that seems to be a general improvement to me:
https://intellipaat.com/community/9428/install-a-python-package-into-a-different-directory-using-pip

So we can simply install `az` to `software/az` folder by providing the `--target=` option (we have to test, maybe the second option is better). Then we would again follow our own conventions and best-practices and have a folder for each tool inside `software`. This is much cleaner and as always "uninstalling" simply means delete the folder, done.

scripts/src/main/resources/scripts/command/az
hohwille: Same problem as discussed in the thread below.
We will fix this properly and can stop hardcoding paths and knowledge about tools that belong to other commandlets.
The right way to go is always to ensure that all binaries are either in the `bin` subfolder or if not present in the tool folder itself. If that is not met, we actually have a bug and tweaking the PATH with manual hacks will not solve it properly but only cause later problems somewhere else.

scripts/src/main/resources/scripts/command/az
alfeilex: I have added the azure installation with .pth file. Make sure that Python must be already installed or if not, that it comes to an error if the shell is not restarted after the Python installation. This may need to be adjusted in the Python command. The Python `Script` folder has only two scripts, which are created by reinstalling pywin32.

All other files belonging to Azure are stored in the `software/az` folder.

alfeilex: After installation, the shell must also be restarted so that az is added in the path.

hohwille: Now working with just `doDevonCommand pip install azure-cli`.
However, this is causing that the software folder is called `azure-cli` instead of `az`.
I also observed other small issues.
I will create a new PR as replacement to be able to push the changes.

scripts/src/main/resources/scripts/command/az
hohwille: already obsolete and resolved.

hohwille:
As I was unable to bring my changes to new files into this existing PR, I created PR #873 that replaces this PR.
Hence, I am closing this one.
@alfeilex your work is great and this is only for technical reasons.

Issue #827, opened 2022-07-14
Fix GitHub Action Runner for Windows


alfeilex:
### Expected behavior

As a developer, I want to test the IDE on Windows via GitHub Action Runner.

### Actual behavior
There is a problem with Action Runner on Windows. Bash tests are not recognized as scripts.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

Use the test-windows workflow under [.github/workflows ](https://github.com/devonfw/ide/tree/master/.github/workflows)to built the IDE. Check the output to see the error.

### Related/Dependent Issues
#691 

### Comments/Hints:
One possibility would be to create a separate profile for Windows in the pom.xml and run the scripts using `bash test-*.`

PR #831, opened 2022-07-21
#827: Fix GitHub Action Runner for Windows

alfeilex:


scripts/pom.xml
```diff
@@ -129,40 +129,24 @@
...
-  <profiles>
-    <profile>
-      <id>integration-test</id>
-      <activation>
-        <os>
-          <family>!Windows</family>
-        </os>
-        <property>
-          <name>!skipTests</name>
-        </property>
```
hohwille, line 143: Please do not remove this feature.
When using maven it is a common standard that you can disable the tests using `-DskiptTests=true`.
With your change, you removed this feature.
Please keep the plugin execution in the profile and just remove`<os>...</os>`.

Also please test the build locally on windows (`mvn clean install`):
* Does it work and find `bash` automatically?
* Does it work if you have installed shellcheck locally on windows?

hohwille (changes requested):
@alfeilex Seems you found a working solution that looks good to me. 👍 
However, as commented in my review, please keep the profile to let users disable the tests if needed.
If skippable and locally tested this should be ready for merge.

scripts/pom.xml
alfeilex: I have added the `!skipTests` feature again. The build runs locally on windows and find `bash` automatically. The `shellcheck` also works. (I used chocolatey to install shellcheck)

hohwille (approved):
@alfeilex thanks for the rework. Now looking perfect. Ready for merge 👍

PR #837, opened 2022-07-27
#738 commandlet for lazydocker added

### LINKED ISSUE #738: CLI Lazy Docker integration
As a devon IDE user, I want to have Lazy Docker available in the devon IDE for my docker installation.

https://github.com/jesseduffield/lazydocker
 
Acceptance criteria:

- [x] Mirror is added.
- [x] There is a new commandlet that allows to install and launch Lazydocker from the CLI
- [x] Installation of the [binary releases](https://github.com/jesseduffield/lazydocker#binary-release-linuxosxwindows)
- [x] During the installation process I get information about the current status of the installation
- [x] Lazydocker can be installed and used together with any of the devon IDE standard tools without negative interference
- [x] Check Docker requirements (Docker >= 1.13 (API >= 1.25); Docker-Compose >= 1.23.2 (optional))
- [x] The new commandlet is documented on the website
- [x] Add licence documentation

arexhepi7:
Fixed #738


hohwille:
Thanks for the PR.
Can you check with [DoD](https://github.com/devonfw/ide/blob/master/documentation/DoD.asciidoc)?
```

In scripts/command/lazydocker line 20:
  cd ${LAZYDOCKER_HOME}
  ^-------------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
     ^----------------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
  cd "${LAZYDOCKER_HOME}" || exit

For more information:
  https://www.shellcheck.net/wiki/SC2164 -- Use 'cd ... || exit' or 'cd ... |...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
Error:  Failed running test #2: test-shellcheck - exit code 1
```

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,72 @@
...
+function doRun() {
+  doSetup silent
+  cd ${LAZYDOCKER_HOME}
+  if doIsWindows
+  then
+    read -r -p "inside if condition of doRun"
+    powershell.exe -command 'start powershell {.\lazydocker.exe}'
+  else
+    ./lazydocker 
+  fi
```
hohwille, line 27: Seems that lazydocker has some issues to run on windows seamless in any shell.
Actually my vision would be that you could just setup lazydocker and then do not need `devon` command and just run `lazydocker` in any shell on any os.
Is that vision out of reach?
Can you share what errors or problems arise when you try this?

cinnamon-coder-hub: This comments issue will be addressed in #842 as recently discussed and the comment can therefore be closed.

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,108 @@
...
+fi
+
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+# Call lazydocker with specified arguments.
+function doRun() {
+  doSetup silent
+  cd "${LAZYDOCKER_HOME}" || exit 1
```
hohwille, line 20: should actually be on `PATH` so `lazydocker` should work directly (instead of `./lazydocker`).
May only be required to start via powershell.

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,108 @@
...
+      fi
+    fi
+    doCheckDockerVersions
+    doInstall "-" "${LAZYDOCKER_HOME}" "lazydocker" "${version}"
+  fi
+}
+
+function doCheckDockerVersions() {
+  local DOCKERENGVERSION
+  DOCKERENGVERSION=$(docker version | grep "Version" | head -1 | sed -ne 's/.*\([0-9][0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p')
```
hohwille, line 53: `docker version --format '{{.Client.Version}}'`

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,108 @@
...
+  local DOCKERAPIVERSION
+  DOCKERAPIVERSION=$(docker version | grep "API version" | head -1 | sed -ne 's/.*\([0-9]\+\.[0-9]\+\).*/\1/p')
```
hohwille, line 55: `docker version --format '{{.Client.APIVersion}}'`

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,108 @@
...
+  doVersionCompare "1.3.0" "${DOCKERENGVERSION}"
+  local res1=$?
+  doVersionCompare "1.25" "${DOCKERAPIVERSION}"
+  local res2=$?
+  local ressum=$((res1 + res2))
+  doWarning "Please be aware, that if you want to use docker-compose with lazydocker, then you'll need at least version 1.23.2 of docker-compose"
```
hohwille, line 61: So this WARNING is printed every time regardless if the versions are fine or not?

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,108 @@
...
+    then
+      doWarning "\nIt seems as if your docker-engine version is too old for a proper usage of lazydocker.\nPlease upgrade it at least to version 1.3.0"
+      read -r -p "Continue by pressing enter"
+    fi
+    if [ "${res2}" == "1" ]
+    then
+      doWarning "\nIt seems as if your docker-API version is too old for a proper usage of lazydocker.\nPlease upgrade it at least to version 1.25.0"
+      read -r -p "Continue by pressing enter"
+    fi
+  elif [ "${ressum}" == "4" ] || [ "${ressum}" == "0" ] || [ "${ressum}" == "2" ]
```
hohwille, line 74: so `4`  means that both versions are higher and `0` that both versions are identical. So `2` means one is equal and one is higher - because you are in the else so it can not be the `1+1` case? Somehow seems correct but does this `ressum` really make anything better? IMHO it is only causing confusion and making the code hard to read.

Why not just kick out `ressum` and do `else` here?
The only possible results are 0 (equal), 1 (greater), 2 (lower):
https://github.com/devonfw/ide/blob/3c49143d7a3167378f90a25daf09b1d890d2dd27/scripts/src/main/resources/scripts/functions#L1381

To avoid confusion it would even be helpful to switch the arguments:
```
doVersionCompare "${DOCKERENGVERSION}" "1.3.0"
local res1=$?
```

so `1` means greater (${DOCKERENGVERSION} > 1.3.0) and `2` means lower (${DOCKERENGVERSION} < 1.3.0).

So these are the cases (without switching version arguments):
|res1|res2|result|
|----|----|------|
|  0  | 0  | fine, both versions equal |
|  0 |  1  | bad, ApiVersion is lower |
|  1 |  0  | bad, Version is lower |
|  1 |  1  | bad, both versions lower |
|  0 |  2  | fine, version equal, ApiVersion higher |
|  1 |  2  | bad, version lower, ApiVersion higher |
|  ... |  ... | ... |

So to make it short, the first if is covering all bad cases, and `else` would be fine.

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,108 @@
...
+  then
+    doSuccess "\nDocker-Engine and Docker-API versions are sufficient"
+  else
+    doWarning "\nSomething went wrong, is your docker-daemon running or did you forget to start Rancher Desktop?"
+    read -r -p "Continue by pressing enter"
```
hohwille, line 79: this case can never happen and could therefore be dropped.

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,108 @@
...
+  if doIsWindows
+  then
+    powershell.exe -command 'start powershell {.\lazydocker.exe}'
+  else
+    ./lazydocker 
+  fi
+}
+
+function doSetup() {
+  if [ ! -d "${LAZYDOCKER_HOME}" ] || [ "${1}" != "silent" ]
```
hohwille, line 30: you should add
```
doDevonCommand docker setup silent
```
as before what will ensure that docker/rancher is first installed if not already present.

hohwille (changes requested):
@alfeilex and antigoni: Thanks for this PR. Almost prefect 👍 
I added some review comments.
Can you check and rework so that we can merge?

alfeilex:
> @alfeilex and antigoni: Thanks for this PR. Almost prefect 👍 I added some review comments. Can you check and rework so that we can merge?

I have been mentioned here incorrectly. @cinnamon-coder-hub has created the PR with @arexhepi7  .

scripts/src/main/resources/scripts/command/lazydocker
cinnamon-coder-hub: Now there is a separate if condition for the case that docker-compose is at an insufficient version or not running. So users with the propper version won't be bothered by this warning anymore.

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -50,33 +56,34 @@ function doSetup() {
 
 function doCheckDockerVersions() {
   local DOCKERENGVERSION
-  DOCKERENGVERSION=$(docker version | grep "Version" | head -1 | sed -ne 's/.*\([0-9][0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p')
+  DOCKERENGVERSION=$(docker version --format '{{.Client.Version}}' | sed -ne 's/.*\([0-9][0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p')
```
hohwille, line 59: this is outputting nothing but the version. So what is the need of the `sed` command?

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -50,33 +56,34 @@ function doSetup() {
...
   local DOCKERAPIVERSION
-  DOCKERAPIVERSION=$(docker version | grep "API version" | head -1 | sed -ne 's/.*\([0-9]\+\.[0-9]\+\).*/\1/p')
-  doVersionCompare "1.3.0" "${DOCKERENGVERSION}"
+  DOCKERAPIVERSION=$(docker version --format '{{.Client.APIVersion}}' | sed -ne 's/.*\([0-9]\+\.[0-9]\+\).*/\1/p')
```
hohwille, line 61: Same here, can you remove the `sed` pipe?

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,116 @@
...
+function doRun() {
+  doSetup silent
+  cd "${LAZYDOCKER_HOME}" || exit 1
+  if doIsWindows
+  then
+    ldInPathRes=$(echo "$PATH" | tr ":" "\n" | grep lazydocker)
+    if [ -z "${ldInPathRes}" ]
+    then
+      export PATH=${PATH}:"${LAZYDOCKER_HOME}"
+    fi
```
hohwille, line 27: This seems to be a missunderstanding.
If `doSetup` was called then `lazydocker` has to be installed and on the `PATH`. Otherwise this is a bug of `doSetup`.
IMHO we do not need this block.

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,116 @@
...
+    powershell.exe -command 'start powershell {lazydocker}'
+  else
+    ./lazydocker
```
hohwille, line 30: so what I actually wanted to say is the you do not need to do `cd ${LAZYDOCKER_HOME}` and can just call `lazydocker` without the `./` what also makes the assumption that it has to be directly in `software/lazydocker` and we get that content from an external source - maybe in future releases they decide to move the binary to a `bin` subfolder.

scripts/src/main/resources/scripts/command/lazydocker
```diff
@@ -0,0 +1,116 @@
...
+  echo "                          call lazydocker by executing the command 'devon lazydocker'. Call lazydocker --help for details or use lazydocker directly as preferred."
+  echo
+;;
+"setup" | "s" )
+  doEcho "Installing lazydocker..."
+  doSetup
+;;
+"version" | "v" )
+  lazydversion=$(cat "${LAZYDOCKER_HOME}"/.devon.software.version)
+  doInfo "Current version of lazydocker is ${lazydversion}"
```
hohwille, line 109: if we want to have this feature in the future for all commandlets, we could simply add it to `functions` so it does not have to be implemented into every commandlet. However, this is just a side-nore for the record. We can file a separate issue for this. Let us get this lazydocker completed and merged.

hohwille (changes requested):
@alfeilex thanks for the rework. The version compatibility checks are nicely improved, etc. 👍 

Still I have some last remarks. I would have done that as constructive review by directly changing the code but as I am going on vacation and do not want to interfere or break anything due to lack of time for testing last minute changes, I leave this up to you for the last round.

scripts/src/main/resources/scripts/command/lazydocker
hohwille: I created issue #847 for this.

scripts/src/main/resources/scripts/command/lazydocker
cinnamon-coder-hub: The lazydocker cmdlet should work, but this additional command seems to make problems in my case. For Antigoni it is fine again. I suspect that this is the case because of some changes that I made for testing purposes, but I will look deeper into this next week and make another commit if necessary.

hohwille (approved):
Thanks for the final rework. Now everything is just perfect 👍

PR #838, opened 2022-07-29
Fixed settings bug

### LINKED ISSUE #835: Setup not working if SettingsURL not ending with .git
### Expected behavior

As a devonfw-ide user, I want to be able to setup my project with any working GIT Settings URL so that I can get started without problems and my project has the flexibility to use any kind of git service with whatever URL formats.
I would even expect any valid GIT URL including local ones (https://stackoverflow.com/questions/978052/how-can-i-make-my-local-repository-available-for-git-pull) to work without issues.

### Actual behavior

The setup tries a CURL download instead of a git clone if the GIT URL does not end with `.git` ending up in strange errors:
```
./setup /d/projects/devonfw/settings
Setting up your devonfw-ide in /d/projects/test
Installing devon autocompletion
Starting installation of devonfw-ide-settings in version  to /d/projects/test/settings
Trying to download settings from /d/projects/devonfw/settings
curl: (3) URL using bad/illegal format or missing URL
Failed to download /d/projects/devonfw/settings with exit code 3

******** ATTENTION ********
ERROR: Failed to download  from /d/projects/devonfw/settings
Exit code was 3
```

Please also see that with git the URL is working fine:
```
git clone /d/projects/devonfw/settings
Cloning into 'settings'...
done.
```

1. download devonfw-ide
2. extract the content to new folder
3. run setup and supply a valid git URL that does not end with `.git`

### Comments/Hints:

Problem comes from here:
https://github.com/devonfw/ide/blob/94a9ff7d46bc6d38146075d531206dfa7908b36f/scripts/src/main/resources/scripts/command/ide#L99

Why don't we just call `doGitPullOrClone` instead of `doDownload`?

Some additional magic to support branch encoded in URL was placed inside `doInstall` instead of `doGitPullOrClone`.
https://github.com/devonfw/ide/blob/86b690280cbc8def707570c1384957283eebc223/scripts/src/main/resources/scripts/functions#L784

I would strongly question this approach and suggest to move that magic to `doGitPullOrClone` and simply call this directly to clone or update `settings` or `mirrors` instead of calling `doDownload` that is IMHO just wrong here.

### Affected version:

* Version 2022.04.003
* OS: Windows, Linux, and Mac

BizarreBits:
closes #835

### DIFF
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -1080,6 +1080,7 @@ function doMoveGlobSafe() {
 function doGitPullOrClone() {
   local dir="${1}"
   local url="${2}"
+  
   local git_opts=""
   if ! doIsDebug
   then
@@ -1134,6 +1135,13 @@ function doGitPullOrClone() {
       popd > /dev/null || exit 255 
     fi
   fi
+  
+  if [ "${url/.git/}" != "${url}" ]
+  then
+    doGitPullOrClone "${target_path}" "${url}"
+    return
+  fi
+  
 }
 
 # $1: download URL
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -96,7 +96,7 @@ function doUpdateSettings() {
     then
       SETTINGS_URL="${defaultSettingsUrl}"
     fi
-    doInstall "${SETTINGS_URL}" "${SETTINGS_PATH}" "devonfw-ide-settings"
+    doGitPullOrClone "${SETTINGS_PATH}" "${SETTINGS_URL}"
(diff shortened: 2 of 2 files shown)


scripts/src/main/resources/scripts/functions
```diff
@@ -1134,6 +1135,13 @@ function doGitPullOrClone() {
       popd > /dev/null || exit 255 
     fi
   fi
+  
+  if [ "${url/.git/}" != "${url}" ]
+  then
+    doGitPullOrClone "${target_path}" "${url}"
```
hohwille, line 1141: 1. Variable `target_path` is undefined here. Did you mean `dir`instead?
2. This seems to cause an infinity loop.

hohwille:
Was replaced by #859.

Issue #840, opened 2022-08-01
Pip integration for python


hohwille:
As discussed in #839 we need to create a commandlet to install and use `pip`.

For acceptance criteria as always see our [DoD](https://github.com/devonfw/ide/blob/master/documentation/DoD.asciidoc).

alfeilex:
I have found to solutions to be able to deposit the installation of Azure CLI and other packages in the software directory. 

First solution: 
- Install package with `pip --target <directory_path> <package_name>`
- create a .pth file in software/python/Lib/site-packages and paste the previous path there
- add `${DEVON_IDE_HOME}/software/<directory_path>/bin` into `$PATH`

=> Leads to an error:
`Error loading command module 'rdbms': No module named 'pywintypes'`
even though the dll is available in the directory. (`<directory_path>/pywin32_system32/pywintypes310.dll`)

This is probably due to a still faulty dependency. But the script will be executed anyway (e.g `devon az`).

Second solution:
- install [virtualenv](https://virtualenv.pypa.io/en/latest/) with `pip install virtualenv`
- add `${DEVON_IDE_HOME}/software/python/Scripts` into `$PATH`
- navigate to the software and create virtual environment with `virtualenv <directory_name>` (copying python interpreter)
- run `activate.ps1` in `<directory_name>/Scripts` (add copied python **with pip** to $PATH)
- add `${DEVON_IDE_HOME}/software/<directory_path>/Scripts` into `$PATH`
- run pip install <package_name>

Runs perfectly without any problems. The created folder can be easily deleted.

Disadvantage: The structure with `bin` folder is not respected.

Akuhana:
Preliminary pull request for pip commandlet implementation.
- Use `devon pip` to install pip.
- Use `devon pip install <package>` to install pip packages.
- Use `devon pip remove` to remove pip.
- Use `devon pip remove <package>` to remove a pip package.

Issue #713, opened 2022-02-24
Python integration


albertdb:
As a devon IDE user, I want to have Python available in the devon IDE so that I and my team can easily ramp up our development environment and we are using the same and correct versions of the tool.
 
Acceptance criteria:

* [x] Mirror is added.
* [x] There is a new commandlet that allows to install and launch Python from the CLI
* [x] The new commandlet is a wrapper for Python CLI
* [x] Python (Version 3.x) can be installed automatically (during setup via settings) as well as via the commandlet call - only for windows
* [x] During the installation process I get information about the current status of the installation

17.05:
* [x] Python (Version 3.x) can be installed automatically (during setup via settings) as well as via the commandlet call - open for Windows & Linux
* [x] The new commandlet is documented on the website
* [x] Python can be installed and used together with any of the devon IDE standard tools without negative interference (tested for windows and linux)
* [x] Python is integrated in a way that complies with the multi-tenancy of the devon IDE - Windows and Linux
* [x] PIP (Package manager) is installed - Linux & Windows

14.06:
* Python (Version 3.x) can be installed automatically (during setup via settings) as well as via the commandlet call - open for MAC
* Review for windows and linux commandlet documentation
* Python can be installed and used together with any of the devon IDE standard tools without negative interference (must be tested for MAC)
* Python is integrated in a way that complies with the multi-tenancy of the devon IDE - MAC
* Add licence documentation
*  PIP (Package manager) is installed - MAC

maybeec:
please make sure, that pip is also installed! I made it once for python integration into devonfw ide and there is some script which needs to be executed and some path adaptations to make python portable work properly. https://stackoverflow.com/questions/23888109/portable-python-cant-install-pip

maybeec:
@albertdb, just have seen the requirement is coming from hangar https://github.com/devonfw/hangar/blob/f651d23d53f74c8c9904da82313a7a878aab2f92/scripts/accounts/aws/custom-policies-check.py

is there a reason, why there is just one python file? Couldn't this be implemented in bash?

Bene90:
HI @BizarreBits 
@alfeilex already handled the multitendency topic within the Azure CLI integration. Maybe he can help here.

hohwille:
@albertdb did you observe the things happening here? Could we get in touch? In case you are a "python expert" you could give us some assistance for small design decisions.
Actually Python is a disaster as its installation process is simply not working for all platforms and e.g. on MacOs it is just broken.
[...]

PR #841, opened 2022-08-01
#840: added pip commandlet

hohwille:
Added pip commandlet for #840 and fixes for #713

### DIFF
--- a/scripts/src/main/resources/scripts/command/pip
+++ b/scripts/src/main/resources/scripts/command/pip
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+
+# autocompletion list
+if [ "${1}" = "shortlist" ]
+then
+  if [ -z "${2}" ]
+  then
+    echo "setup help version pip pip3 version -v --version"
+  fi
+  exit
+fi
+
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+function doSetup() {
+  doDevonCommand python setup silent
+  local PIP_CMD
+  PIP_CMD="$(command -v pip)"
+  local software="${DEVON_IDE_HOME}/software/"
+  if [ "$?" == 0 ] && [ "${PIP_CMD::${#software}}" = "${software}" ]
+  then
+    if [ "${1}" = "silent" ]
+    then
+      doEcho "Pip is already installed at ${PIP_CMD}"
+    fi
+  else
+    if doIsWindows
+    then
+      doInstall "-" "${PIP_HOME}" "pip" "latest" "" "pip"
+      cd "${PIP_HOME}"
+      PYTHON_HOME="${DEVON_IDE_HOME}/software/python"
+      #doDevonCommand python pip-latest-windows.py --no-warn-script-location "--target=${DEVON_IDE_HOME}/software/python"
+      doDevonCommand python pip-latest-windows.py --no-warn-script-location "--target=${PIP_HOME}"
+      #rm "${PYTHON_HOME}/pip-latest-windows.py"
+      #PYTHON_HOME="${DEVON_IDE_HOME}/software/python"
+      #if [ -d "${PYTHON_HOME}/Scripts" ]
+      #then
+      #  echo mv "${PYTHON_HOME}/Scripts/"* "${PYTHON_HOME}/"
+      #  mv "${PYTHON_HOME}/Scripts/"* "${PYTHON_HOME}/"
+      #fi
+    else
+      doEcho "unsupported!"
+    fi
+  fi
+  if [ "${1}" != "silent" ] && ! doIsQuiet
+  then
+    doRunCommand "pip --version"
+  fi
+}
+
+function doRunPip() {
+  doSetup silent
+  if doIsQuiet
+  then
+    doRunCommand "pip ${*}" > /dev/null
+  else
+    doRunCommand "pip ${*}"
+  fi
+}
+
+PIP_HOME="${DEVON_IDE_HOME}/software/pip"
+
+case ${1} in 
+  "help" | "-h")
+    echo "Install Pip (Python Package Installer)."
+    echo
+    echo "Arguments:"
+    echo " setup                    setup pip."
+    echo " «args»                   call pip with the specified arguments (call 'pip --help' for details)."
+    echo
+  ;;
+  "setup" | "s" | "")
+    doSetup "${2}"
+  ;;
+  "version" | "-v" | "--version")
+     doRunPip --version
+  ;;
+  *)
+     doRunPip "${@}"
+  ;;
+esac
--- a/scripts/src/main/resources/scripts/command/python
+++ b/scripts/src/main/resources/scripts/command/python
@@ -5,7 +5,7 @@ if [ "${1}" = "shortlist" ]
 then
   if [ -z "${2}" ]
   then
-    echo "setup help version pip pip3 version -v --version"
+    echo "setup help version version -v --version"
   fi
   exit
 fi
@@ -26,7 +26,7 @@ function doSetup() {
     if doIsMacOs
     then
       doBrewInstall openssl
-      doInstall "-" "${PYTHON_HOME}" "python" "${PYTHON_VERSION}" "" "Python"
+      doInstall "-" "${PYTHON_HOME}" "python" "${PYTHON_VERSION}"
       sh "${PYTHON_HOME}"/configure MACOSX_DEPLOYMENT_TARGET=12 --enable-optimizations --with-ensurepip=install --with-openssl=/opt/homebrew/opt/openssl@3 --prefix="${PYTHON_HOME}"
       make -j 8
       make altinstall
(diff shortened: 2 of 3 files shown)


alfeilex:
I have found to solutions to be able to deposit the installation of Azure CLI and other packages in the software directory. 

First solution: 
- Install package with `pip --target <directory_path> <package_name>`
- create a .pth file in software/python/Lib/site-packages and paste the previous path there
- add `${DEVON_IDE_HOME}/software/<directory_path>/bin` into `$PATH`

=> Leads to an error:
`Error loading command module 'rdbms': No module named 'pywintypes'`
even though the dll is available in the directory. (`<directory_path>/pywin32_system32/pywintypes310.dll`)

This is probably due to a still faulty dependency. But the script will be executed anyway (e.g `devon az`).

Second solution:
- install [virtualenv](https://virtualenv.pypa.io/en/latest/) with `pip install virtualenv`
- add `${DEVON_IDE_HOME}/software/python/Scripts` into `$PATH`
- navigate to the software and create virtual environment with `virtualenv <directory_name>` (copying python interpreter)
- run `activate.ps1` in `<directory_name>/Scripts` (add copied python **with pip** to $PATH)
- add `${DEVON_IDE_HOME}/software/<directory_path>/Scripts` into `$PATH`
- run pip install <package_name>

Runs perfectly without any problems. The created folder can be easily deleted.

Disadvantage: The structure with `bin` folder is not respected.

hohwille:
> run activate.ps1 in <directory_name>/Scripts (add copied python with pip to $PATH)
...
Runs perfectly without any problems.

Thanks for this hint (albert also pointed me to virtualenv).
Does this still work when the user opens a new shell (e.g. git-bash here inside devonfw-ide installation) and types `az`?
Or does the user first have to call `activate.ps1` again to setup the "virtual environment"?
In the latter case, we still do not yet have a working solution.
But maybe we could "hack" `environment-project` scripts to automatically setup the environment for python.
I always wanted to avoid tool specific logic in such scripts but if there is no other way, this may be an option.

However, I would still love to have an installation like `az` that can find python and other potential python specific dependencies automatically via PATH or some config file (like `.pth` file). With all other technologies and tools we made this happen - why not for Python?

hohwille:
I tested virtualenv:
It does create a full copy of the python and pip installation. 
The idea of devonfw-ide is that we have a sandbox with various software tools each in `software/«tool»` folder.
Then it should be possible to setup and update each of such tool.
So my expecation would be:
* devon python setup --> installs python to `software/python` in the version configured via `PYTHON_VERSION` variable.
* devon pip setup --> installs pip to `software/pip`
* devon az setup --> installs az via pip into `software/az`
* change value of `PYTHON_VERSION` and rerun `devon python setup`--> updates python in `software/python` and then running `az` would use the new version of python.

Why is that impossible with pyhton? We could solve this for all other tools so far.

hohwille:
DoD :)
```

In scripts/command/pip line 32:
      cd "${PIP_HOME}"
      ^--------------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.

Did you mean: 
      cd "${PIP_HOME}" || exit

For more information:
  https://www.shellcheck.net/wiki/SC2164 -- Use 'cd ... || exit' or 'cd ... |...
Error:  Failed running test #2: test-shellcheck - exit code 1
```

alfeilex:
I tried to install pip via `devon pip` but it gives the following error message:

`******** ATTENTION ********
Unknown archive format: .py. Can not extract /c/Users/alfeil/Downloads/devonfw-ide/pip-latest-windows.py
`

alfeilex:
> * devon az setup --> installs az via pip into `software/az`

The only way I found is to create a .pth file in `../software/python/Lib/site-packages` and add the path from Azure, which I installed via 
`pip install --target <path_to_software/az> azure-cli` in devon software folder.

With the `--target` option, there is also a `bin` folder where the executable scripts are located. 

There was an error that says that the module `pywintypes` is missing. I could fix this error with `pip uninstall pywin32` and `pip install pywin32`. So our idea to put Azure and other software in its own folder might work.

Do you know any other CLI or software that I can test with this method? Maybe only azure-cli has this kind of error which can be fixed by reinstalling pywin32.

Akuhana:
@hohwille The pip commandlet should work as of now correctly.

hohwille:
Tests failing:
```
In scripts/command/pip line 24:
  local PIP_CMD="$(command -v pip)"
        ^-----^ SC2155: Declare and assign separately to avoid masking return values.


In scripts/command/pip line 40:
        echo "../setuptools" >> "${software}/python/modules.pth"
        ^-- SC2129: Consider using { cmd1; cmd2; } >> file instead of individual redirects.


In scripts/command/pip line 165:
    doRemove ${2}
             ^--^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
    doRemove "${2}"

For more information:
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
  https://www.shellcheck.net/wiki/SC2129 -- Consider using { cmd1; cmd2; } >>...
Error:  Failed running test #2: test-shellcheck - exit code 1
```

scripts/src/main/resources/scripts/command/pip
```diff
@@ -0,0 +1,170 @@
...
+  # Install pip package
+  local package="${1}"
+  local software="${DEVON_IDE_HOME}/software/"
+  local package_path="${software}${package}"
+
+  # If pywin32 is not installed in the python Lib directory, install it. This is required for the azure cli to work.
+  if [ ! -d "${software}/python/lib/site-packages/win32" ]
+  then
+    doRunPip "install pypiwin32"
+  fi
```
hohwille, line 75: This is highly windows specific and will make it fail on linux or MacOs.

scripts/src/main/resources/scripts/command/pip
```diff
@@ -0,0 +1,170 @@
...
+    doRunCommand "pip3 --version"
+  fi
+}
+function doInstallPackage(){
+  # Install pip package
+  local package="${1}"
+  local software="${DEVON_IDE_HOME}/software/"
+  local package_path="${software}${package}"
+
+  # If pywin32 is not installed in the python Lib directory, install it. This is required for the azure cli to work.
```
hohwille, line 71: is this only required by azure? Should it then also be installed from az commandlet instead of here?

scripts/src/main/resources/scripts/command/pip
```diff
@@ -0,0 +1,170 @@
...
+  # Create new package folder in software directory
+  if [ ! -d "${package_path}" ]
+  then
+    mkdir -p "${package_path}"
+  fi
+
+  # Write a new line to the modules.pth if the line doesn't exist
+  if ! grep -q "../${package}" "${software}/python/modules.pth"
+  then
+    echo "../${package}" >> "${software}/python/modules.pth"
```
hohwille, line 86: can we now the current working directory ($PWD) here?
This relative path may be wrong and we should use an absolute path instead.

hohwille:
@Akuhana you have done an awesome job. I did rework and improve according to the above comments but you solved pip and some of the tricky parts of python installation and dependencies very well.
However, when I now try to install something with pip I get strange results:
```
$ devon pip install az
Collecting pypiwin32
  Using cached pypiwin32-223-py3-none-any.whl (1.7 kB)
Collecting pywin32>=223
  Using cached pywin32-304-cp310-cp310-win32.whl (11.0 MB)
Installing collected packages: pywin32, pypiwin32
Successfully installed pypiwin32-223 pywin32-304
Success: run command pip
Collecting az
  Using cached az-0.1.0.dev1-py3-none-any.whl (1.5 kB)
Installing collected packages: az
Successfully installed az-0.1.0.dev1
Success: run command pip
Package az installed successfully! If you get an error, that the command was not found, run 'devon' or restart your terminal window and try running that command again.

$ find software/az/
software/az/
software/az/az
software/az/az/__init__.py
software/az/az/__pycache__
software/az/az/__pycache__/__init__.cpython-310.pyc
software/az/az-0.1.0.dev1.dist-info
software/az/az-0.1.0.dev1.dist-info/INSTALLER
software/az/az-0.1.0.dev1.dist-info/METADATA
software/az/az-0.1.0.dev1.dist-info/RECORD
software/az/az-0.1.0.dev1.dist-info/REQUESTED
software/az/az-0.1.0.dev1.dist-info/top_level.txt
software/az/az-0.1.0.dev1.dist-info/WHEEL
```

As you can see the installation of azure CLI is not as expected and there is no `az` binary at all.

hohwille (commented):
For python and pip setup all looks sane and clean to me.
We still have to check why the installation of packages like `az` with `pip` gives unexpected results.

scripts/src/main/resources/scripts/command/pip
Akuhana: This might be also necessary for other packages. It is similar to the package wheel or setuptools, which is why it is a good idea to have it installed correctly beforehand.

hohwille: Thanks for the explanation. I updated the command and added `e.g.` before azure to make that clear.

scripts/src/main/resources/scripts/command/pip
hohwille: I added `doIsWindows`.

PR #845, opened 2022-08-11
doUnzip: add variable definition + shellcheck mod

cinnamon-coder-hub:
fixed #737 and changed if condition (tested as well) so that shellcheck won't be alerted.


CREITZ25 (commented):
From my point of view we have only add the command:
target_dir=${2}   before
doRunCommand "unzip...

Deleting echo is ok and the changes in 717/718 are ok, too.

hohwille (changes requested):
@cinnamon-coder-hub thanks for creating this PR and improving doUnzip. 👍 
Interesting: according to your changes `target_dir` was undefined before and this could have never worked before. I am confused as I thought this has already been tested. Further, I wonder why shellcheck did not report the undefined variable. 
We have meanwhile `doInstallWithPackageManager` and it was planned that also `doUnzip` gets changed to use this new function. Could you be so kind and update this accordingly?

For the record: We started to define the parameters as local variables at the beginning of each function (what together with the functions documentation also makes the comments of the arguments obsolete). Could you also update this as well?

hohwille:
To prevent blocking out release, I replaced this PR with #871.

Issue #786, opened 2022-06-09
Improve Windows Terminal with powershell


CREITZ25:
This issue shall complete issue #701 to also add an option to the windows explorer context menu to open a devonfw-ide powershell in Windows-Terminal.

PR #855, opened 2022-08-24
Feature/wt

CREITZ25:
Little bugfix for devon CMD in Windows Terminal.  Title is now set to "devon CMD"


tobka777 (approved):
Thanks. We should take care that the path is "wt-cmd" everywhere at the end.

hohwille:
>  Title is now set to "devon CMD"

When it is about documentation, titles, etc. can you please always use the registered trademark `devonfw` instead of `devon`?

CREITZ25:
I corrected it, but I add it to the other (the last) issue/PR:

#786 Improve Windows Terminal with powershell

hohwille:
> I corrected it, but I add it to the other (the last) issue/PR: #786

Great, thanks. So I close this issue and we focus on getting #786 merged.
If I got something wrong, feel free to reopen and comment :)

Issue #846, opened 2022-08-11
Add Linux Support of Rancher Desktop to docker cmdlet + Bug of setup for macOS


cinnamon-coder-hub:
As a user with a common linux distribution I want to be able to install Rancher Desktop by executing `devon docker setup` . Currently Rancher Desktop can only be installed in the above way on Windows and MacOS.
Further while using 'devon docker setup' on macOs the last setup command inside the docker script 
'doRunCommand 'cp -a "${rancher_install_dir}"/*.app /Applications/'' is not working so that the setup of Rancher Desktop fails.

PR #856, opened 2022-08-25
#846: Added Linux support for setup of Rancher Desktop

cinnamon-coder-hub:
fixing #846.


tobka777 (approved):
Look's fine!

cinnamon-coder-hub:
This PR now includes two small commits fixing a setup bug for Rancher Desktop on macOS that was added here due to its small size and and because of also being setup related.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -62,9 +62,11 @@ function doSetup() {
           doEnsureVpnKitInstallation
         elif doIsMacOs
         then
-          doRunCommand "cp -a '${rancher_install_dir}'/*.app /Applications/"
+          # shellcheck disable=SC2016
+          doRunCommand 'cp -a "${rancher_install_dir}"/*.app /Applications/'
```
hohwille, line 66: While this change seems sane it is not releated to Linux as indicated by the bug issue or the PR title.
Fine, but in such cases you may also leave a comment in the PR summary why this had to be changed.

scripts/src/main/resources/scripts/command/docker
```diff
@@ -62,9 +62,11 @@ function doSetup() {
...
         else
-          doFail "Your OS is not supported currently"
+          doRunCommand "cp -a \"${rancher_install_dir}/.\" \"${DEVON_IDE_HOME}/software/docker/\""
+          doRunCommand "${DEVON_IDE_HOME}/software/docker/rancher-desktop"
```
hohwille, line 69: Is it valid to copy it to software and run it from there?
I see two possible cases:
* running `rancher-desktop` is just an installer that will install Rancher Desktop globally into your Linux system (e.g. to `/usr` or `/usr/local`). If so then why do you need to copy it first instead of directly running this from the `rancher_install_dir`?
* this is already the actual Rancher Desktop installation and running `rancher-desktop` just launches the already installed Rancher Desktop. If this is the case doesn't this conflict with our global installation of Rancher Desktop for Win and Mac as the behavior differs (on Win and Mac Rancher only gets installed once by the first IDE installation while on Linux it gets installed inside every IDE installation and deleting the IDE installation will also remove/uninstall Rancher)?

For the record: It is great that you found a solution for Linux and it is already a great improvement compared to failing with an "not supported" error. I just want to trackle your solution to understand the big picture...

hohwille (approved):
@cinnamon-coder-hub thanks for this PR.
I left 2 small review comments. You may want to comment or even react with a small change but it should also be fine to merge as is. I will wait a few days for you to give a reaction. Feel free to comment something like "agreed, but lets first merge and consider further improvements via later PRs".

scripts/src/main/resources/scripts/command/docker
hohwille: I linked the bug issue and indeed there it is described both in the title and in the issue description.
Hence, I am resolving this conversation.

hohwille:
Due to lack of feedback or reaction, I created issue #872 and will now merge this PR as is.

scripts/src/main/resources/scripts/command/docker
hohwille: Created #872

PR #858, opened 2022-08-30
#858: bugfix for devon ide create

CREITZ25:
The "path" parameter for devon ide create <path> results sometimes in error. Now it is fixed


hohwille (approved):
@CREITZ25 Excellent catch. Thanks for finding and fixing 👍

PR #859, opened 2022-08-30
#835 Setup not working if SettingsURL not ending with .git

CREITZ25:
#835 Setup not working if SettingsURL not ending with .git


hohwille (approved):
@CREITZ25 thanks. This should fix it. 👍

PR #860, opened 2022-08-30
#786 add readme to windows terminal with Powershell, cmd and GitBash

CREITZ25:
I created the readme file for this three extensions


hohwille:
I was initially referring to such magic hex code:
https://github.com/devonfw/ide/blob/d5a35311f7ae80f682fd76746e92537b7404fde4/scripts/src/main/resources/system/windows/windows-terminal/devon-wt-cmd.reg#L7

I might miss something but the README your created does not relate to that hex stuff and how to create it.

hohwille:
Maybe you wanted to explain in the readme that you are doing this temporary stuff manually in `regedit` and then at the end export the hex stuff from there to create the reg file. If that is the case, then you are one the right way but the readme needs some more details so that others can understand what you want to say. The explanation of the rationale with variables otherwise not working is already very helpful to get a motivation why all this is required here...

CREITZ25:
Extension of the readme.asciidoc with a step-by-step guide

hohwille (approved):
@CREITZ25 thanks for the update. Now everything gets clear. Awesome 👍

Issue #857, opened 2022-08-29
Create function doIsEmptyDirectory


hohwille:
I know that I am picky, but in the review of PR #830 I stumbled over the complicated way to check if a directory is empty:
```
    # shellcheck disable=SC2012
    count_lines="$(ls -alF "${create_path}" | wc -l)"
    if [ "${count_lines}" -gt 3 ]
    then
      doAskToContinue "Directory is not empty, continue?"
    fi
```
Indeed there is no trivial build in bash function to determine this.
However, I searched the web for the most elegant solution to test if a directory is empty and found this:
```
if [ -z "$(find "${dir}" -maxdepth 0 -empty)" ]
then
  ...
fi
```

As I do not want to block #830 for the sake of my perfection, I decided to create this issue.
We should add a function `doIsEmptyDirectory` to `functions` that uses the more elegant trick and then change the ide commandlet to do this instead:
```
    if ! doIsEmptyDirectory "${create_path}"
    then
      doAskToContinue "Directory is not empty, continue?"
    fi
```

hohwille:
Closing as PR is already merged.

PR #863, opened 2022-08-30
#857 Create function doIsEmptyDirectory

CREITZ25:
#857 Create function doIsEmptyDirectory


scripts/src/main/resources/scripts/command/ide
```diff
@@ -460,12 +460,11 @@ function doCreate() {
       mkdir -p "${create_path}"
     fi
     # shellcheck disable=SC2012
```
hohwille, line 462: this can now also be deleted. I can do that...

hohwille: Done.

hohwille (approved):
@CREITZ25 Excellent. Looks perfect to me. Thanks 👍

Issue #832, opened 2022-07-22
Solution for unstable update sites


hohwille:
Actually the design of the plugin ecosystem of most tools or IDEs (esp. Eclipse) is a disaster:
* Plugins are provided decentralized via URLs
* When the plugin is installed at a specific point in time it is loaded from that URL
* The maintainer of that URL is in full control of the availability and the content behind that URL. 
* If the service gets broken, obviously the plugin can not be downloaded.
* Even worse if existing content gets updated, the result is not reproducible anymore.
* The major problem is that exactly the latter happens all the time.
* All famous IDEs do not have a clear versioning concept: While the plugin artifact itself is typically versioned, the request to install a plugin can not specify a version but always downloads the "latest" version. If at all some kind of versioning for stability is in place it happens via the URL itself so for different major versions different URLs are provided.
* A concrete example are the JBoss Plugins available at this update site: http://download.jboss.org/jbosstools/updates/webtools/photon/
* While at some point of time the plugin installation worked, at a later point in time all the plugins had been updated and stopped working with the configured eclipse version of that project.
* Even worse after fixing this in the project settings it worked again for some time and now is entirely broken again as RedHad published new plugin versions that are entirely broken, had expired certificates, etc.
* As a result the entire project with hundreds of developers was blocked and it was impossible to rebuild the IDE with Eclipse and the required projects from scratch.
* This is just an example and we had the same story also for other Eclipse Plugins and also for other tools like SonarQube, IntelliJ, or VS Code.

The perfect solution would obviously to change all these tools to have a reliable and reproducible plugin installations:
* Ensure that all plugins get maintained in a central repository (like maven)
* Ensure that every artifact in this repo is strictly versioned and once a version gets deployed it can never be changed anymore.
* Ensure that a plugin (group) installation can specify an exact version (if omitted the latest version is used). So if a specific version is given you always get the same reproducible result at any point in time in the future.

However, the sad fact is chances to make this happen are practically zero!

There is already some kind of workaround offered by devonfw-ide to address the problem:

We already offer ability for a project specific [custom repo](https://github.com/devonfw/ide/blob/master/documentation/software.asciidoc#custom). This can be used to archive a fully build IDE including all required plugins into a *.tar.gz file and provide it on your own server in such a custom repo. Then devonfw-ide would not download the IDE vanilla release and then install all the plugins but instead simply download and unpack that complete archive with IDE + plugins.

This approach has several drawbacks:
[...]

PR #881, opened 2022-09-12
#832: Solution for unstable update sites

alfeilex:


scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -36,6 +36,10 @@ function doInstallEclipsePlugin() {
     echo
     doEchoAttention "Installation of plugin ${plugin} failed. Please try to install manually or rerun setup (devon eclipse setup)."
     echo
+  else
+    echo
+    doSuccess "Installation of plugin ${plugin} was successful."
```
alfeilex, line 41: I add this line to get an output if a plugin has been installed successfully.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -254,6 +258,42 @@ function doImportEclipse() {
...
 
+# $1: URL
+# $2: name
+# $3: optional dest
+function doMirrorUpdateSite() {
+  if [ -z "${2}" ]
+  then
+    doFail "Parameter <URL> or <name> is missing."
+  fi
+  local dest="${ECLIPSE_HOME}/update-sites/"
```
alfeilex, line 269: Should we change the default destination of the mirrored update-sites? Maybe in conf?

scripts/src/main/resources/scripts/command/eclipse
hohwille: Can you remove the empty `echo` statements?

scripts/src/main/resources/scripts/command/eclipse
hohwille: Yes, I would not put this into the Eclipse installation.
Maybe just use the download folder of devonfw-ide as default and create an updates-sites subfolder there.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -254,6 +258,42 @@ function doImportEclipse() {
...
+  if [ -n "${3}" ]
+  then
+    dest="${3}"
+  fi
+  if doIsWindows
+  then
+    eclipsecli="${ECLIPSE_HOME}/eclipsec.exe"
+  fi
+  "${eclipsecli}" -nosplash -verbose -application org.eclipse.equinox.p2.metadata.repository.mirrorApplication -source "${1}" -destination "${dest}/${2}_${timestamp}"
+  "${eclipsecli}" -nosplash -verbose -application org.eclipse.equinox.p2.artifact.repository.mirrorApplication -source "${1}" -destination "${dest}/${2}_${timestamp}"
```
hohwille, line 282: why is this line duplicated?

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -254,6 +258,42 @@ function doImportEclipse() {
...
+  local timestamp
+  timestamp=$(date +'%m-%d-%Y')
+  local eclipsecli="${ECLIPSE_HOME}/eclipse"
+  if [ -n "${3}" ]
+  then
+    dest="${3}"
+  fi
+  if doIsWindows
+  then
+    eclipsecli="${ECLIPSE_HOME}/eclipsec.exe"
```
hohwille, line 279: why dont you just call `eclipse` and keep this OS independent?

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -254,6 +258,42 @@ function doImportEclipse() {
   "${ECLIPSE_HOME}/eclipse" -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.ant.core.antRunner -buildfile "${DEVON_IDE_HOME}/scripts/lib_script/import.xml" -data "${WORKSPACE_PATH}" -keyring ~/.eclipse/.keyring -vm "${ECLIPSE_JAVA_HOME}/bin/java" -vmargs -DdevonImportPath="${importPath}" -DdevonImportWorkingSet="${importWorkingSets}" ${ECLIPSE_VMARGS} 
 }
 
+# $1: URL
+# $2: name
+# $3: optional dest
```
hohwille, line 263: could we also support the existing plugins configuration?
So if only one argument is provided get the URL from according plugin properties file and the name from `$1`.
So if `checkstyle` is provided then it reads the update site from here:
https://github.com/devonfw/ide-settings/blob/master/eclipse/plugins/checkstyle.properties#L1

There is already code to add plugins handling this stuff that could be reused.

scripts/src/main/resources/scripts/command/eclipse
alfeilex: There are two commands. The first one downloads the meta-data and the second one downloads the main content of the plugin.

scripts/src/main/resources/scripts/command/eclipse
hohwille: Maybe we could even add a `plugin_mirror_url` property overriding `plugin_url` for plugin installations so I can always just use one param instead of two?

scripts/src/main/resources/scripts/command/eclipse
alfeilex: `Eclipsec.exe` is the console executable version of eclipse.exe for Windows. Other OS's use `eclipse`. It was recommended in the docs to do it this way.

scripts/src/main/resources/scripts/command/eclipse
alfeilex: Should I extend the  `add-plugin` function so that if `plugin_mirror_url` is set, this URL is used instead of `plugin_url`?

scripts/src/main/resources/scripts/command/eclipse
hohwille: but there is absolutely no difference between the first and the second line. So your explanation does not make sense to me.

scripts/src/main/resources/scripts/command/eclipse
hohwille: Which "docs"? You can just call `eclipse` without any path and it works on any OS.

scripts/src/main/resources/scripts/command/eclipse
hohwille: OMH. Sorry. I thought I did look twice and more with care, but indeed you are correct that the one line contains `metadata` and the other one instead contains `artifact`. OK. Issue resolved.

scripts/src/main/resources/scripts/command/eclipse
alfeilex: I have now added two parameters for the mirror function. The first ist the id of the plugin (e.g. checkstyle) and the second is an optional URL. If there is only one parameter, the  `plugin_url` will be extracted automatically from the related plugin property file. If the optional URL is set as second parameter, the function will mirror this URL instead. 

I choose this structure because in my opinion the first parameter (plugin id) is needed in any case, because the download folder is named by this. With only an URL as paramter I would have to implement something for the name. Moreover, the user only needs to pass a single world (e.g. checkstyle) as a parameter instead of the path of the related property file, if we were to take the property file as a parameter.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -254,6 +260,45 @@ function doImportEclipse() {
...
+    local file="${SETTINGS_PATH}/eclipse/plugins/${1}.properties"
+    if [ -f  "${file}" ]
+    then
+      plugin_id=""
+      plugin_url=""
+      doLoadProperties "${file}"
+      if [ -z "${plugin_id}" ] || [ -z "${plugin_url}" ]
+      then
+        doFail "Invalid eclipse plugin config: ${file}"
+      fi
```
hohwille, line 284: Can we avoid copy&paste and code duplication by moving this to a function?

alfeilex: done

alfeilex:
Should there be an error message if the plugin_url is empty? If there is no link to the original update site, then we should change this condition. 

https://github.com/alfeilex/ide/blob/38f37a7f9b78f3448fde95b8f64a4ff7b603c6a1/scripts/src/main/resources/scripts/command/eclipse#L296

hohwille (approved):
@alfeilex excellent. Thanks for the rework. 👍 
Ready for merge.

hohwille:
> Should there be an error message if the plugin_url is empty?

Thanks for taking care and asking. But that is fine as it was: this property should always be present and filled.

PR #882, opened 2022-09-13
Fixing NPE by updating mvn default version to 3.8.4

maybeec:
As of https://issues.apache.org/jira/browse/MNG-6921 this could be a blocking issue for some projects, so we should upgrade the default maven version to the latest bugfix release of maven 3


hohwille (approved):
@maybeec excellent. Thanks for this nice improvement and the link to the maven bug. 👍 
Ready to merge. As I just had to merge with master, and am on the leave (almost midnight now and going on business trip tomorrow) feel free to merge yourself.
Otherwise I will do so next week.

Issue #877, opened 2022-09-12
Update version of vscode to 1.71.0+


hohwille:
As a devonfw-ide user, I want to get latest version of vscode by default so that it integrates properly with other things (e.g. github).
Currently with 2022.08.001 default version is still 1.54.1 what does not work with github anymore:
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/vscode#L21

alfeilex:
There is an error with [devonfw-extension-pack.properties](https://github.com/devonfw/ide-settings/blob/1e7e0ede083acb1b06e55d2eebc536096c548f76/vscode/plugins/devonfw-extension-pack.properties#L1) 

It causes this output: `Installing extensions...
Installing extension 'devonfw.devonfw-extension-pack'...
(node:10184) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use Code --trace-deprecation ... to show where the warning was created)
connect ETIMEDOUT 13.107.6.175:443
Failed Installing Extensions: devonfw.devonfw-extension-pack`

The functionality of the other plugin [lombok](https://github.com/devonfw/ide-settings/blob/1e7e0ede083acb1b06e55d2eebc536096c548f76/vscode/plugins/lombok.properties#L1) also needs to be tested. It contained errors too.

alfeilex:
The installation of vscode on the Remote Mac also didn't work: Here the output: 
<img width="407" alt="vscode-m1-output" src="https://user-images.githubusercontent.com/101652401/189950994-8824d9c5-b882-4a59-8cac-34886dab7d49.PNG">

PR #884, opened 2022-09-13
#877: Update version of vscode to 1.71.0+

alfeilex:
There are some errors with vscode plugins. But on windows and linux it works without the plugin [devonfw-extension-pack.properties](https://github.com/devonfw/ide-settings/blob/1e7e0ede083acb1b06e55d2eebc536096c548f76/vscode/plugins/devonfw-extension-pack.properties). So it can be merged. I created two bug tickets for the issues: 

- #887 
- #888


hohwille (approved):
@alfeilex thanks. All fine 👍

Issue #879, opened 2022-09-12
Renaming integration-tests and adding badges


alfeilex:
The current nightly tests have the name "CI Build ...". But they are not really CI Builds and should be renamed to e.g. "Integration Test <OS>".  Furthermore, the status of the tests should be displayed in the README.asciidoc. As this is the case for License, Maven Central and CI Build. 

Acceptence criteria:

- Rename the current `CI Build <OS>` in e.g. `Integration Test <OS>`
- Add the status of these tests to [README.asciidoc](https://github.com/devonfw/ide/blob/master/README.asciidoc)

PR #889, opened 2022-09-15
#879: Renaming integration-tests and adding badges

alfeilex:


hohwille (approved):
@alfeilex Thanks. All looking good. 👍

PR #890, opened 2022-09-16
GitHub cli

Akuhana:
GitHub CLI should be implemented and working correctly as of this writing

### DIFF
--- a/scripts/src/main/resources/scripts/command/gh
+++ b/scripts/src/main/resources/scripts/command/gh
@@ -40,7 +40,7 @@ function doSetup(){
 
 function doRun(){
   doSetup silent
-  doRunCommand "${GH_HOME}/bin/gh ${@}"
+  doRunCommand "${GH_HOME}/bin/gh ${*}"
 }
 
 case ${1} in
--- a/scripts/src/main/resources/scripts/command/gh
+++ b/scripts/src/main/resources/scripts/command/gh
@@ -25,9 +25,7 @@ function doSetup(){
     if [ -z "${GH_VERSION}" ] || [ "${1}"!="silent" ]
     then
       doEcho "Getting latest release..."
-      local GH_VERSION=$( curl -LsH "Accept: application/vnd.github.v3+json" https://api.github.com/repos/cli/cli/releases/latest | awk -F':' '/tag_name/ {print $2}' | awk -F'"' '{print $2}' )
-      # Remove v from version
-      GH_VERSION="${GH_VERSION:1}"
+      local GH_VERSION=$( curl -LsH "Accept: application/vnd.github.v3+json" https://api.github.com/repos/cli/cli/releases/latest | awk -F':' '/tag_name/ {print $2}' | awk -F'"' '{print $2}' | tr -d 'v' )
 
       doEcho "Installing GitHub CLI..."
       doInstall "-" "${GH_HOME}" "gh" "${GH_VERSION}"
--- a/scripts/src/main/resources/scripts/command/gh
+++ b/scripts/src/main/resources/scripts/command/gh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+# autocompletion list
+if [ "${1}" = "shortlist" ]
+then
+  if [ -z "${2}" ]
+  then
+    echo "setup help"
+  fi
+  exit
+fi
+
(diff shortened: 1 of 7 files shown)


scripts/src/main/resources/scripts/command/gh
```diff
@@ -0,0 +1,57 @@
...
+
+      doEcho "Installing GitHub CLI..."
+      doInstall "-" "${GH_HOME}" "gh" "${GH_VERSION}"
+    fi
+  fi
+}
+
+function doRun(){
+  doSetup silent
+  doRunCommand "gh ${*}"
```
tobka777, line 39: Does this work? I think you should add the path like `doRunCommand "${HELM_HOME}/helm" "${@}"` to call the binary.

scripts/src/main/resources/scripts/command/gh
```diff
@@ -0,0 +1,57 @@
...
+  # Install GitHub CLI in the software folder
+
+  if [ ! -d "${GH_HOME}" ]
+  then
+    # Get leatest release
+    if [ -z "${GH_VERSION}" ] || [ "${1}" != "silent" ]
+    then
+      doEcho "Getting latest release..."
+      local GH_VERSION
+      GH_VERSION=$( curl -LsH "Accept: application/vnd.github.v3+json" https://api.github.com/repos/cli/cli/releases/latest | awk -F':' '/tag_name/ {print $2}' | awk -F'"' '{print $2}' | tr -d 'v' )
```
tobka777, line 29: Nice solution!!!

scripts/src/main/resources/scripts/command/gh
```diff
@@ -0,0 +1,57 @@
...
+if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+GH_HOME="${DEVON_IDE_HOME}/software/gh"
+
+function doSetup(){
+  # Install GitHub CLI in the software folder
+
+  if [ ! -d "${GH_HOME}" ]
```
tobka777, line 22: add `|| [ "${1}" != "silent" ]` and remove it from the next if statement, because you should return the version otherwise. Add this codesnippet after this if statement as alternative like
```
 if [ "${1}" != "silent" ] && ! doIsQuiet
  then
    doRunCommand "helm version"
  fi

File: scripts/src/main/resources/scripts/command/gh [hunk 4]
```diff
@@ -0,0 +1,57 @@
...
+  then
+    # Get leatest release
+    if [ -z "${GH_VERSION}" ] || [ "${1}" != "silent" ]
+    then
+      doEcho "Getting latest release..."
+      local GH_VERSION
+      GH_VERSION=$( curl -LsH "Accept: application/vnd.github.v3+json" https://api.github.com/repos/cli/cli/releases/latest | awk -F':' '/tag_name/ {print $2}' | awk -F'"' '{print $2}' | tr -d 'v' )
+
+      doEcho "Installing GitHub CLI..."
```
tobka777, line 31: I think you can move this echo to the switch case (line 52) like in other commandlets.

tobka777 (changes requested):
Thank you for your great work. I add some small improvements, but it's almost done.

Akuhana:
@tobka777 Thank you for reviewing and pointing out the important details of GitHub CLI commandlet implementation. I have made the changes and committed them.

scripts/src/main/resources/scripts/command/gh
```diff
@@ -0,0 +1,62 @@
...
+
+  if [ "${1}" != "silent" ] && ! doIsQuiet
+  then
+    doRunCommand "gh --version"
+  fi
+}
+
+function doRun(){
+  doSetup silent
+  doRunCommand "${GH_HOME}/bin/gh ${*}"
```
tobka777, line 43: I think it's better with *${@}*:
```
doRunCommand "${GH_HOME}/bin/gh" "${@}"

tobka777 (changes requested):
Thank you. Only a last issue...

scripts/src/main/resources/scripts/command/gh
hohwille: `doRunCommand` takes the command as first argument. Second is optional description of the task to use in log message instead of using the command itself (default if omitted).

tobka777 (approved):

Issue #880, opened 2022-09-12
Customize the trigger of the integration tests


alfeilex:
Currently, an integration test consists of two jobs. The first one checks if a commit has been executed in the last 24h and the other one runs the integration test for the specific os. Unfortunately, the output of the first job is always true and the second job runs only if the commit was 24h before. Otherwise the integration-test is skipped. 

So if there is an error in the commit, the test will output the error correctly. But if the integration-test is done after 24h of the commit, it will output true as a whole, because the commit-check will output true and the second job will be skipped. The error will not be displayed at all. You can observe this behavior if you check all integration-tests that run after 24h of a commit. They all return true even though the integration test is not run, but has been skipped.

Acceptence criteria:

- modify the [check-commit.yml](https://github.com/devonfw/ide/blob/master/.github/workflows/check-commit.yml) so that the integration-test action is only displayed when the test has run for the first time after the latest commit
- further integrations-tests should either be skipped or not triggered at all.

PR #891, opened 2022-09-16
#880: Customize the trigger of the integration tests

alfeilex:
I have modifed the integration tests so that if the last commit happens after 24 hours, the running workflow is canceled.


hohwille (approved):
@alfeilex thanks for this PR. This all looks good to me. 👍 
If I got it correct, you removed the `check-commit.yml` and replaced it with `verify_commit` job that you directly integrated into all 3 tests. We have gathered quite some copy&paste redundancy between those 3 builds. Maybe we can later find a way how to reuse the most logic from a central file that we include into the 3 test workflows. However, just a nice to have for later.
Your new solution for cancelling and checking changes looks promising.
I will simply see if it works as expected after merging so fingers crossed :)

PR #896, opened 2022-09-19
#898: Bugfix for aws commandlet

CREITZ25:
Bugfix for aws commandlet


scripts/src/main/resources/scripts/command/aws
```diff
@@ -16,8 +16,22 @@ source "$(dirname "${0}")"/../functions
 
 # Call aws with specified arguments.
 function doRun() {
-  doSetup silent
```
hohwille, line 19: why is this not wroking?
This is the common way to do it in all other commandlets.
Cant we make it work the same way here?

scripts/src/main/resources/scripts/command/aws
```diff
@@ -29,43 +43,46 @@ function doPackageInstall()
...
+      fi
+
+      if [ "${1}" != "silent" ] && ! doIsQuiet
+      then
+        doEcho "logon again to test aws with \"aws --version\""
+      fi 
     fi
+  else
+    echo "aws is already installed"
+    doRunCommand "aws --version"
```
hohwille, line 82: `silent` means that if the tool is already is installed, then nothing should be logged and almost nothing should happen (that takes time or requires network access). So this `else` block should only happen if not silent.
The regular thing to do that in all other commandlets is that separate `if` that you removed here:
```
if [ "${1}" != "silent" ] && ! doIsQuiet
```
You can simply remove the message "aws is already installed".

scripts/src/main/resources/scripts/command/aws
```diff
@@ -29,43 +43,46 @@ function doPackageInstall()
...
+    powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${windows_path_to_package}\aws-${AWS_VERSION}-windows.msi /quiet'"
   elif doIsMacOs
   then
     doFail "Sorry, AWS CLI installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
   else
     doEcho "Installating AWS for Linux..."
     doRunCommand "sudo ${path_to_package}/install"
   fi
-  doRunCommand "rm -rf ${path_to_package}"
+  doRunCommand "rm -rf ${path_to_package}"
```
hohwille, line 54: accidental change to add trailing spaces?

scripts/src/main/resources/scripts/command/aws
```diff
@@ -29,43 +43,46 @@ function doPackageInstall()
   then
     doEcho "Installing AWS for Windows..."
     windows_path_to_package=$(cygpath -w "${path_to_package}")
-    powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${windows_path_to_package}\aws-${current_version}-AWSCLIV2.msi /quiet'"
+    powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${windows_path_to_package}\aws-${AWS_VERSION}-windows.msi /quiet'"
```
hohwille, line 46: function is not really reusable as `${AWS_VERSION}` has to be initialized what happens inside `doSetup`.
Unfortunately bash has no concept of visibilities like `private` vs. `public`.
However, I understood that you extracted this function to increase readability (`doSetup` gets shorter and this "package installation" is grouped in a function with a self explanatory name. So all fine...

hohwille (changes requested):
@CREITZ25 Thanks for testing, finding bugs and fixing. 👍 
Still I have some questions/doubts... Can you have a look. Lets also try to do a short call together...

scripts/src/main/resources/scripts/command/aws
CREITZ25: I changed this line multiple times to see what happend by running the script.  I will delete the spaces again.

scripts/src/main/resources/scripts/command/aws
CREITZ25: Now I add the version as the second parameter. The better way.

scripts/src/main/resources/scripts/command/aws
CREITZ25: It misunderstood the word "silent". Now I corrected it

scripts/src/main/resources/scripts/command/aws
hohwille: nice solution. Thanks.

hohwille (approved):
@CREITZ25 thanks for addressing all review comments perfectly. 👍 
Ready for merge.

Issue #868, opened 2022-09-08
Further improvements for nightly integration-tests


hohwille:
From review of PR #844:
A new file `functions` was added in test:
> maybe this file should have a different name (e.g. test-functions, ``)? Otherwise we might end up in confusions if someone talks about functions and it is unclear if he means the productive functions or the test functions. We also have a functions documentation and some places that just talk about `functions` what is currently a unique file and so it is clear what we are talking about.

Further we currently have tests for
* terraform
* vscode

But how about other tools and features?
* eclipse
* intellij
* node
* npm
* az
* ng
* python
* pip
* yarn
* oc
* graalvm
* helm
* dotnet
* cobigen
* ...

alfeilex:
To add new commands for the integration test, you can create a new file named `integration-test-<command>` and follow this example. 

https://github.com/devonfw/ide/blob/6ddbbc26e7e51fd008118d7165014691ac8a495c/scripts/src/test/bash/integration-test-terraform#L1-L4

Windows users must be careful when pushing the new test-scripts. Because Windows does not have a permission system like Linux, the following command must be run beforehand, otherwise a permission denied error will occur in github action run.

`git update-index --chmod=+x <script-path>`

PR #899, opened 2022-09-21
Improvements for nightly integration-tests

ahmedagdmoun:
I improved the nightly integration-tests as described in #868.
I also updated the index of all newly created files to avoid  the permission denied error from github actions.


scripts/src/test/bash/integration-test-docker
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+source "$(dirname "${0}")"/test-functions
+doCommandTest docker
```
hohwille, line 4: Typically `docker` is already present inside github action environment.
Do we have any idea what happens if we run this on windows (or mac)?
IMHO a real installation of Rancher Desktop on Windows can never work via github action as it shows installation UI screens that a user has to interact with.

hohwille (approved):
@ahmedagdmoun thanks for PR and good work. All looks fine to me. 👍 
I have a little feat that with all the UI tools like eclipse and intelli we make the integration test really slow but still we want to test everything an I hope that it will work with our available resources.
I also left some review comment about docker doubts but I will already merge it as is and we can simply see the effects.

Issue #908, opened 2022-09-26
AWS CLI integration for MacOS


CREITZ25:
As a devon IDE user, I want to add AWS CLI for MacOS, too.

Acceptance criteria:

The aws commandlet has to be changed, so that it allows to install and launch AWS CLI from the CLI also on MacOS

PR #909, opened 2022-09-26
#908 AWS CLI integration for MacOS, bugfixes for Windows and Linux

CREITZ25:
Adding feature to install aws on MacOS


scripts/src/main/resources/scripts/command/aws
```diff
@@ -42,7 +55,7 @@ function doPackageInstall()
 
 function doSetup() {
 
-  if [ ! -d "${AWS_HOME}" ] || [ "${1}" != "silent" ]
+  if [[ $(command -v aws) != "/c/Program Files/Amazon/AWSCLIV2/aws" ]] || [ "${1}" != "silent" ]
```
ahmedagdmoun, line 58: The first condition will never be met under MacOS because the given path is Windows specific. As a result AWS is being installed again and again every time devon aws is called. **command -v aws** returns different results (paths) depending on the OS and doSetup should be OS independent. So I think using the function doIsAwsInstalled would now make more sense here.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -51,21 +64,19 @@ function doSetup() {
       AWS_VERSION=$(curl "https://github.com/aws/aws-cli/tags" | awk -F'tags/' '/archive/ { print $2}' | sort -r | head -1 | awk -F'.zip' '{print $1}')
     fi
 
-    doInstall "-" "${AWS_HOME}" "aws" "${AWS_VERSION}"
```
ahmedagdmoun, line 54: wouldn't it be better if we keep the **AWS_HOME** variable and use it here as before? it was set to the same given value here.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -51,21 +64,19 @@ function doSetup() {
       AWS_VERSION=$(curl "https://github.com/aws/aws-cli/tags" | awk -F'tags/' '/archive/ { print $2}' | sort -r | head -1 | awk -F'.zip' '{print $1}')
     fi
 
-    doInstall "-" "${AWS_HOME}" "aws" "${AWS_VERSION}" 
+    doInstall "-" "${DEVON_IDE_HOME}/updates/install/aws" "aws" "${AWS_VERSION}" 
     if [ "${?}" = 0 ]
     then
-      doPackageInstall "${AWS_HOME}"
+      doPackageInstall "${DEVON_IDE_HOME}/updates/install/aws" "${AWS_VERSION}"
```
ahmedagdmoun, line 70: Same notice as above.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -51,21 +64,19 @@ function doSetup() {
...
-
-  if [ "${1}" != "silent" ] && ! doIsQuiet
-  then
-    doEcho "logon again to test aws with \"aws --version\""
-  fi 
 
+    if [ "${1}" != "silent" ] && ! doIsQuiet
+    then
+      doEcho "logon again to test aws with \"aws --version\""
+    fi
```
ahmedagdmoun, line 76: I think this should stay outside the first if statement. Or is there a reason to move it inside?

scripts/src/main/resources/scripts/command/aws
```diff
@@ -51,21 +64,19 @@ function doSetup() {
...
-  fi 
 
+    if [ "${1}" != "silent" ] && ! doIsQuiet
+    then
+      doEcho "logon again to test aws with \"aws --version\""
+    fi 
+  fi
 }
 
-AWS_HOME="${DEVON_IDE_HOME}/updates/install/aws"
```
ahmedagdmoun, line 68: As mentioned above, I think you can still use this variable so no need to remove it. Or is there a particular reason why you removed it?

ahmedagdmoun (changes requested):
@CREITZ25 Thank you for the PR. I have added a couple of notes concerning the new changes. Furthermore, I have tested the installation with PuTTY and the remote mac and noticed that after a successful setup of aws, no aws folder was found in `${DEVON_IDE_HOME}/updates/install` or `${DEVON_IDE_HOME}/software`. I think this was intended but shouldn't the installation folder also be available somewhere inside the devonfw-ide installation folder? I'm just asking to be sure.

@ahmedagdmoun Jörg said the installation software shouldn't be saved, since it would also use up disk space

scripts/src/main/resources/scripts/command/aws
CREITZ25: Using AWS_HOME does not work, because it means, we install the software on all systems in the same directory, and that is not possible.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -17,22 +17,35 @@ source "$(dirname "${0}")"/../functions
 # Call aws with specified arguments.
 function doRun() {
   doSetup silent
-  doRunCommand "${AWS_HOME}/aws ${*}"
+  doRunCommand "aws ${*}"
+}
+
+
+# OS independent check if aws is installed.
+function doIsAwsInstalled() {
```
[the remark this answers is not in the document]
CREITZ25: The function was meant to be used here:
if [[ $(command -v aws) != "/c/Program Files/Amazon/AWSCLIV2/aws" ]] || [ "${1}" != "silent" ]

I will correct it

scripts/src/main/resources/scripts/command/aws
CREITZ25: If the doIsAwsInstalled function is used instead, it also works on Mac

scripts/src/main/resources/scripts/command/aws
CREITZ25: Using AWS_HOME does not work, because it means, we install the software on all systems in the same directory, and that is not possible.

scripts/src/main/resources/scripts/command/aws
CREITZ25: I will check it again

scripts/src/main/resources/scripts/command/aws
```diff
@@ -44,17 +56,22 @@ function doPackageInstall()
     powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${windows_path_to_package}\aws-${version}-windows.msi /quiet'"
   elif doIsMacOs
   then
-    doFail "Sorry, AWS CLI installation support is not yet implemented for your OS. Please install manually or help devonfw-ide to support it for your OS by contributing a pull-request."
+    doEcho "Installing AWS for MacOS..."
+    sudo installer -pkg "${path_to_package}"/aws-"${version}"-mac.pkg -target /
```
tobka777, line 60: If this command does not put the installation in the root folder of the hard drive `-target /`? I think we should put it in `/usr/local/bin` or something like that.

tobka777 (approved):
Looks good. Only one question...

scripts/src/main/resources/scripts/command/aws
CREITZ25: This depends on the tar command. tar- x extrakt the files to usr/local/bin, so if I give / as target directory, it will put all files to /usr/local/bin/

scripts/src/main/resources/scripts/command/aws
```diff
@@ -17,16 +17,28 @@ source "$(dirname "${0}")"/../functions
...
 # OS independent check if aws is installed.
 function doIsAwsInstalled() {
   if command -v aws &> /dev/null
-  then 
+  then
     return
   else
+    if doIsWindows
+    then
+      if [ -f "/c/Program Files/Amazon/AWSCLIV2/aws" ]
```
tobka777, line 37: Please use `%ProgramFiles%` and not an exact path.

tobka777 (changes requested):

scripts/src/main/resources/scripts/command/aws
CREITZ25: The created version runs without errors at any platform. I'm not able to redesign it with ${PROGRAMFILES}. Please accept this version for now, and please, create a new issue for this point.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -17,16 +17,28 @@ source "$(dirname "${0}")"/../functions
 # Call aws with specified arguments.
 function doRun() {
   doSetup silent
-  doRunCommand "aws ${*}"
+  if doIsWindows
+  then
+    doRunCommand "\"/c/Program Files/Amazon/AWSCLIV2/aws\" ${*}"
```
hohwille, line 22: This is redundant to line 37 and as @tobka777 mentioned you should use `%ProgramFiles%` variable as on some windows systems it may be on a different drive or in a different folder.
Unfortunately this variable is not directly available but I figured out that you can get it in git-bash via `cygpath "$(powershell  -c '${env:ProgramFiles}')"`.

I would suggest to do the following:
```
function findAws() {
  AWS="aws"
  AWS_INSTALLED=
  if command -v aws &> /dev/null
  then
    AWS_INSTALLED="yes"
  elif doIsWindows
  then
    local aws_win
    aws_win="$(powershell  -c '${env:ProgramFiles}')"
    aws_win="$(cygpath "${aws_win}")"
    aws_win="${aws_win}/Amazon/AWSCLIV2/aws"
    if [ -x "${aws_win}" ]
    then
      AWS="${aws_win}"
      AWS_INSTALLED="yes"
    fi
  fi
}
```

Then you can call this function before invoking `aws` and to run it simply do:
```
doRunCommand "'${AWS}' ${*}"
```

scripts/src/main/resources/scripts/command/aws
CREITZ25: I think now it works

scripts/src/main/resources/scripts/command/aws
hohwille: @CREITZ25 Great. Thanks that you found this `${PROGRAMFILES}` variable.
I tested this and it works.
No idea, why I did not get this with tab completion when I searched for this during my review.
Lets cross fingers and hope for the best.
If we or someone else can ever reproduce this newline instead of whitespace error, we simply create a new bug issue and try to fix or workaround it.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -63,28 +81,31 @@ function doSetup() {
...
-case ${1} in 
+case ${1} in
 "help" | "-h")
   echo "Setup or run AWS CLI (command-line interface for Amazon-Web-Services)."
   echo
   echo "Arguments:"
-  echo " setup                          install aws on your machine."
-  echo " «args»                         call aws with the specified arguments. Call aws --help for details or use aws directly as preferred."
+  echo " setup                    install aws on your machine."
+  echo " <<args>>                 call aws with the specified arguments. Call aws --help for details or use aws directly as preferred."
```
hohwille, line 107: IMHO @sarahffm changed the indendation of the help output in all commandlets to the same width so that added generic args fit the same indendation width and appear aligned.
Can you check this?

hohwille (approved):
@CREITZ25 Excellent. Thanks for intensive testing and perfection of this 👍 
For me this is functionally ready to be merged.
Can you check the indentation of the help output (see review comment)?

scripts/src/main/resources/scripts/command/aws
hohwille: @tobka777 can you approve as this is now perfectly resolved?

scripts/src/main/resources/scripts/command/aws
CREITZ25: Done

tobka777 (approved):

Issue #905, opened 2022-09-26
Better error handling when installing Eclipse plugins


Akuhana:
As a dewonfw user, I want to see reasonable error messages so that I can take action instantly when the installation of a certain plugin failed. Most of the times, you need to go to the logs and search for the problem, that occurred when the installation of a plugin failed. This may take a lot of time. To solve this problem, I want to see meaningful error messages as well as suggestions on what the underlying problem might be. One way of doing so would be to write a function which checks what plugins are already installed.

Akuhana:
PR https://github.com/devonfw/ide/pull/911

Issue #903, opened 2022-09-26
Uninstall Eclipse plugins


Akuhana:
As a devonfw user, I want to be able to uninstall Eclipse plugins so that I can install a never version of the plugin or have a way to test if the plugin update sites work correctly faster.

- Add doRemovePlugin function
- Documentation in the [CHANGELOG](https://github.com/devonfw/ide/blob/master/CHANGELOG.asciidoc)
- Edit the DoD of the commandlet to reflect the changes

Akuhana:
PR https://github.com/devonfw/ide/pull/952

PR #911, opened 2022-09-28
devonfw/ide#905: Better error messages for eclipse

Akuhana:
The improved logging of plugin installation with eclipse was implemented. To be able to create a pull request for the [plugin removal](https://github.com/devonfw/ide/issues/903) issue, this pull request will have to be completed first, as a new function doCheckPluginsInstalled() ist required.


scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,80 @@ source "$(dirname "${0}")"/../functions
...
 function doInstallEclipsePlugin() {
-  local plugin="${1/\.feature\.group*/}"
-  echo "Installing eclipse plugin ${plugin} from ${2}"
-  # shellcheck disable=SC2086
-  "${ECLIPSE_HOME}/eclipse" -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.equinox.p2.director -repository "${2}" -installIU "${1}" -vmargs ${ECLIPSE_VMARGS}
-  result=${?}
-  if [ "${result}" != 0 ]
+  # Remove feature.group from each plugin id
+  local plugins="${1//.feature.group/}"
+  echo "Installing Eclipse plugin '${plugin_name}' from URL ${2}"
```
hohwille, line 32: after reviewing this I wonder what happened here (even before your change and what I had done myself here long time ago)...

`plugin_name` is not defined within this function but in `doAddPlugin` or `doAddPlugins` as a side-effect. This is not very clean or intuitive to understand. Maybe it would be better to pass it as an explicit argument and always declare it as local variable...

However, your change is good to print the simpler name for betting understanding. However, in case someone now calls `devon eclipse add-plugin net.sf.eclipsecs.feature.group https://checkstyle.org/eclipse-cs-update-site` this used to print out `net.sf.eclipsecs` and with your change will print `net.sf.eclipsecs.feature.group`. So if we would pass `plugin_name` as 3rd param (`$3`)  and if undefined fallback to `"${1/\.feature\.group*/}"` we would get the best result and maintainability.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,80 @@ source "$(dirname "${0}")"/../functions
...
+
+  if [ -z "${plugins}" ]
   then
-    doEchoAttention "Installation of plugin ${plugin} failed. Please try to install manually or rerun setup (devon eclipse setup)."
+    doEcho "No plugins to install."
+    return
   fi
```
hohwille, line 38: Did this really happen for you?
Anyhow, fair to add this check so feel free to set this to resolved immediately.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,80 @@ source "$(dirname "${0}")"/../functions
...
+}
+
+# Check if the plugins are already installed by looking at the eclipse/plugins folder
+# $1: plugin feature group ID
+# return: dependencies that are already installed
+function doCheckPluginsInstalled() {
+  local plugin_dir="${DEVON_IDE_HOME}/software/eclipse/plugins"
+
+  # Remove spaces from plugin variable
+  plugins="${1// /}"
```
hohwille, line 78: always declare variables in functions with `local` - also all the other ones you added. See my comment above. Not declaring local variables makes them global and that leads to these hacks where someone users a variable that is declared somewhere hard to find and in the end you can never tell if the function works correct or not...

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,80 @@ source "$(dirname "${0}")"/../functions
...
+        break
+      fi
+      # shellcheck disable=SC2010
+      new_plugin_dependencies=$(ls -1 "${plugin_dir}" | grep -o -E "${plugin}\..*\.jar")
+      if [ -n "${new_plugin_dependencies}" ] 
+      then
+        plugin_dependencies="${plugin_dependencies}\n${new_plugin_dependencies}"
+      fi
+      plugin=${plugin%.*}
+    done
```
hohwille, line 98: So you implemented logic to detect if a plugin is present in Eclipse. In Bash. Wow!
However, I am concerned that we can do this in a reliable way.
Are you aware of the distinction between eclipse `plugins` and `features` folder?
Are you aware that a plugin can exist as *.jar file OR as a extracted folder and both variants are legal cases?
Do you know that you are relying on a naming convention of `plugin group ID` and `plugin artifact ID` that might be kept as best practices by most eclipse plugins but maybe not by all of them...

OK, if we use this to see if the plugin is already installed and if we find this applied we do not install again most probably the problematic case will lead to us not detecting that the plugin is already installed in an odd case so we keep the existing behavior to install it then anyway so maybe this is fine...

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,80 @@ source "$(dirname "${0}")"/../functions
...
+
+  # Remove spaces from plugin variable
+  plugins="${1// /}"
+  # Iterate each plugin via comma seperation
+  IFS=',' read -ra plugins_arr <<< "${plugins}"
+  for plugin in "${plugins_arr[@]}"; do
+    # Remove feature from plugin variable
+    plugin=${plugin/\.feature*/}
+    # shellcheck disable=SC2010
+    plugin_dependencies=$(ls -1 "${plugin_dir}" | grep -o -E "${plugin}\..*\.jar")
```
hohwille, line 85: FYI: this can also be done with bash build-in glob pattern matching `"${plugin_dir}/${plugin}"*.jar`.
To avoid "too many arguments" problem you would however need to pass it to a function that returns `[ -e "$1" ]`.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,80 @@ source "$(dirname "${0}")"/../functions
...
+      plugin_dependencies=$(echo -e "${plugin_dependencies}" | sort -u)
+      doWarning "Plugin '${plugin_name}' might be already installed, because of existing dependencies: ${plugin_dependencies} \nMake sure the plugin is not installed yet by checking the list of installed plugins in the Eclipse IDE. If there's an error and the issue persists, try to remove the plugin manually from the plugins folder or uninstall it with devon remove-plugin command."
+    fi
+    return ${exit_code}
+  fi
+}
+
+# Check if the plugins are already installed by looking at the eclipse/plugins folder
+# $1: plugin feature group ID
+# return: dependencies that are already installed
```
hohwille, line 73: bash is kind of crappy by design. You can only return exit codes (0-255). Apart you can either echo and redirect the result into a variable when calling (`result="$(doMyFunction)"`) and finally you can also use ugly side-effects of manipulating global variables. IMHO you are doing the latter. Then you should document the variable(s) your side-effect is applying to. However, this way or returning should be avoided whenever there is a suitable alternative.

hohwille (changes requested):
@Akuhana Thanks for your PR. This is a good improvement and makes it better for the end-user. 👍 
Aspects like line 178 and 181 are awesome and I wonder why we did not have this initially already.
However, I left some picky review comments for further improvements.
Thanks for taking them into account :)

scripts/src/main/resources/scripts/command/eclipse
Akuhana: Thank you for pointing out an really important issue, that I haven't considered yet. I have removed the `plugins="${1/\.feature\.group*/}"` line from the `doInstallEclipsePlugin()` function altogether, because only `doCheckPluginsInstalled()` needs to strip the strings down to their base name.

scripts/src/main/resources/scripts/command/eclipse
Akuhana: This will most likely never happen unless the .properties file has an empty plugin_id parameter. Nevertheless, I think It's not a bad idea to check the edge cases

scripts/src/main/resources/scripts/command/eclipse
Akuhana: Indeed, this way of plugin detection is not reliable at all. It can only point out to a potential problem if the problem is detected. And the problem might not always be because one of the given dependencies are already installed. Despite of this, I think it is good idea to point out to the user, that some of the required dependencies are installed already. At the end of the day, it only gives out a warning in the console and doesn't do any error correction itself, as I'd need a lot more knowledge on how the plugin/feature system in Eclipse IDE works to be able to correct the error.

Akuhana: The suggestions make a lot of sense and I wish I would have known earlier, that a folder can also be seen as a plugin. In most cases however, I have observed, that when installing a plugin from the Eclipse IDE, new files are created in `features` folder as well as in `plugins` folder. If you don't want to iterate the folders themselves, the same logic from above can also be used with the `artifacts.xml` file, which contains all of the plugins id's.

scripts/src/main/resources/scripts/command/eclipse
Akuhana: I have made some changes, but I am still unsure how to correct the side-effects that are caused here. Do you have any suggestions on how to change the return value?

scripts/src/main/resources/scripts/command/eclipse
Akuhana: I've tested your suggestion and it works indeed, but it returns the variable with whole directory attached to it, which has to be then removed later. It adds additional steps. However, if by doing so it could reduce execution time, then it might be worth considering that approach.

Akuhana: I've implemented the new version that works without grep

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,91 @@ source "$(dirname "${0}")"/../functions
...
+  plugin_dependencies=$(doCheckPluginsInstalled "${1}")
+
+  # Install plugins and write debug info into temp file
+  local debug_info_file
+  debug_info_file=$(mktemp)
+  "${ECLIPSE_HOME}/eclipsec" -consoleLog -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.equinox.p2.director -repository "${2}" -installIU "${1}" -vmargs "${ECLIPSE_VMARGS}" > "${debug_info_file}" 2>&1
+  local exit_code=$?
+  # Write only lines with !MESSAGE into debug info
+  local debug_info
+  debug_info=$(grep -E "!MESSAGE" "${debug_info_file}")
```
hohwille, line 54: log files from eclipse could get large in case of errors.
Why do we need to read them into memory?

We could simply pipe the `grep` already before writing the log if we do not need anything else than the `!MESSAGE` lines.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,91 @@ source "$(dirname "${0}")"/../functions
...
+  
+  rm "${debug_info_file}"
+
+  if [[ "${debug_info}" == *"Overall install request is satisfiable"* ]] && [ "${exit_code}" == 0 ]; then
```
hohwille, line 58: now we could grep this on the log file (before deleting it - do we have to delete the log after all?) to avoid reading it into memory.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -23,19 +23,91 @@ source "$(dirname "${0}")"/../functions
...
+    doEcho "No plugins to install."
+    return
+  fi
+
+  local plugin_dependencies
+  plugin_dependencies=$(doCheckPluginsInstalled "${1}")
+
+  # Install plugins and write debug info into temp file
+  local debug_info_file
+  debug_info_file=$(mktemp)
```
hohwille, line 49: So far eclipse plugin stuff gets logged to `software/eclipse/configuration/*.log`.
Why don't we keep such location and also preserve the log instead of deleting it (see `rm` command below)?

hohwille (commented):
I left some last review comments. Maybe I can take over from here and do a constructive review (directly apply changes myself to address review comments and then merge).
Thanks again for this PR and the nice improvement. 👍

scripts/src/main/resources/scripts/command/eclipse
Akuhana: A very good point! Sometimes the debug_info variable was too large to finish the execution of the command, that is why I opted out to get only `!MESSAGE` lines.

scripts/src/main/resources/scripts/command/eclipse
Akuhana: I've removed the `rm` command, because the file is going to be removed either way even without users intervention.

scripts/src/main/resources/scripts/command/eclipse
Akuhana: I've made the suggested changes and pushed them

hohwille (approved):
@Akuhana thank you so much for even completing the review comments 👍 

Tested with devstyle:
```
$ devon eclipse add-plugin devstyle
Installing Eclipse plugin 'devstyle' from URL https://www.genuitec.com/updates/devstyle/ci/

Installation of plugin 'devstyle' failed:
!MESSAGE Using unsafe http transport to retrieve http://downloads.genuitec.com/downloads/products/devstyle/updates/ds20220613a, see CVE-2021-41033. Consider using https instead.
!MESSAGE Cannot complete the install because one or more required items could not be found.
!MESSAGE Software being installed: DevStyle with CodeTogether Bundle 1.11.0.202206271604 (com.genuitec.eclipse.theming.feature.feature.group 1.11.0.202206271604)
!MESSAGE Missing requirement: DevStyle with CodeTogether Bundle 1.11.0.202206271604 (com.genuitec.eclipse.theming.feature.feature.group 1.11.0.202206271604) requires 'org.eclipse.equinox.p2.iu; com.genuitec.eclipse.codetogether.feature.feature.group 1.0.0' but it could not be found
```

When I run a regular installation with this new feature, I get:
```
Installing Eclipse plugin 'anyedit' from URL https://raw.githubusercontent.com/iloveeclipse/plugins/latest/
Plugin anyedit installed successfully.
Installing Eclipse plugin 'checkstyle' from URL https://checkstyle.org/eclipse-cs-update-site
Plugin checkstyle installed successfully.
Installing Eclipse plugin 'cobigen' from URL https://devonfw.com/cobigen/updatesite/stable/
Plugin cobigen installed successfully.
Installing Eclipse plugin 'startexplorer' from URL https://basti1302.github.io/startexplorer/update/
Plugin startexplorer installed successfully.
```

I could find the log in `C:\Users\hohwille\AppData\Local\Temp\tmp.aMGHO7MQWw` with this content:
```
!MESSAGE Overall install request is satisfiable
!MESSAGE Add request for StartExplorer  Feature 1.6.1 (de.bastiankrol.startexplorer.feature.feature.group 1.6.1) is satisfiable
!MESSAGE No digest algorithm is available to verify download of osgi.bundle,de.bastiankrol.startexplorer,1.6.1.201409160940.
!MESSAGE No digest algorithm is available to verify download of org.eclipse.update.feature,de.bastiankrol.startexplorer.feature,1.6.1.
```

So it seems to work and improve the plugin installation.

Issue #847, opened 2022-08-11
Generic way to output tool version 


hohwille:
I observed that we started crafting individual logic into commandlets to get the version of a specific tool (e.g. `devon «tool» version` or `devon «tool» -v`). The way this is implemented slightly differs between commandlet and most of them do not support this.
We should check if we can implement this inside functions as a cross-cutting aspect so in an ideal case no commandlet has to implement anything (existing implementations can be removed) but it is working for every commandlet.

Please note that there is already a generic CLI handling inside functions where this could be addressed easily:
https://github.com/devonfw/ide/blob/3c49143d7a3167378f90a25daf09b1d890d2dd27/scripts/src/main/resources/scripts/functions#L1597

hohwille:
This slightly relatest to #826

PR #913, opened 2022-09-29
#847 Generic way to output tool version

CREITZ25:
#847 Generic way to output tool version

Functions to test whether a specified commandlet is installed and to determine the tool's software directory as a prerequisite for determining the current version.


hohwille:
Meanwhile overtaken by PR #932.

Issue #826, opened 2022-07-14
List available versions for a tool


tobka777:
As a devon-IDE user, I want to know which versions are available for a tool i.e. java so that can easily switch the version of a tool.

Acceptance criteria:
- [x] start with JAVA
- [x] save all versions to a file (name: available-versions) in https://github.com/devonfw/ide-mirrors
- [x] create nightly github action to crawl versions and update the file and push changes to git.
- [ ] suggest the versions to the user during installation and update
- [ ] optionally the autocomplete function can be extended

hohwille:
@tobka777 thanks for filing the issue.
Some thoughts on this:
* The `versions` file should be sorted from latest to oldest version.
* To determine the latest version, we can then just get the first line via `head` command.
* We could define some specific syntax for versions such as e.g.:
```
devon «tool» version list - list all available versions of the tool
devon «tool» version set - choose a version from the list with potential auto-completion, etc. and save it as `«tool»_VERSION` variable in `settings/devon.properties`.
devon «tool» version set «version» - like above but set the given «version». If `latest` is given as `«version»` the latest version is used. Maybe the above can also be omitted and the versions could be outputted for bash completion that already does the completion for us.
devon «tool» version - print the version of «tool» that is currently installed
```
Still to be discussed but sounds promising.
Some people tend to install things like https://github.com/nvm-sh/nvm#node-version-manager--- and SDKMan (#772) also works in this way. So making it easier for the user to install a specific or latest version surely makes sense.

hohwille:
Reopening as following aspects are still open:
* exceptions for special commands (e.g. `maven` vs. `mvn`)
* autocompletion

PR #919, opened 2022-09-30
Functions to list and set versions

sarahffm:
#826 List available versions for a tool

Work in progress...
Doesn't fully support the tools maven, nodejs and sonarqube yet (all tools where the command differs from the name of the corresponding mirrors folder).


hohwille (approved):
@sarahffm Wow! Excellent work. Well Done. Thanks 👍 
The only thing I am missing is a function defined and used to map the irregular tool names like `maven -> mvn`. We can however add this after merging.

Issue #933, opened 2022-10-10
available-versions for java invalid


hohwille:
### Expected behavior

As a devonfw-ide user, I want to make use of the new version support including completion, etc. so that my life gets easier maintaining versions. Therefore available-versions should contain only valid stable versions of the according tool that are available.

### Actual behavior

Currently [available-versions for java](https://github.com/devonfw/ide-mirrors/blob/master/java/available-versions) contains `19_36` as latest version, so I tried to install it:

```
devon java set version 19_36
devon java setup
Updating java from version 11.0.13_8 to version 19_36...
Starting installation of java in version 19_36 to /d/projects/devonfw/software/java
/d/projects/devonfw/scripts/command/../functions: line 398:
******** ATTENTION ********
None of the followin path(s) exists:
/d/projects/devonfw/mirrors/java/Temurin/os-mappings
/d/projects/devonfw/mirrors/java/os-mappings
/os-mappings
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.: No such file or directory
Trying to download java-19_36-windows.zip from https://github.com/adoptium/temurin19-binaries/releases/download/jdk-11/OpenJDK19U-jdk_x64_windows_hotspot_19_36.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404
Failed to download https://github.com/adoptium/temurin19-binaries/releases/download/jdk-11/OpenJDK19U-jdk_x64_windows_hotspot_19_36.zip with exit code 22

******** ATTENTION ********
Download of java in version 19_36 failed after trying 1 mirror(s).
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. `devon java set version 19_36`
2. `devon java setup`

### Related/Dependent Issues

* #826
* #931

### Comments/Hints:

There is a typo in the error message that I already fixed (`followin` is missing a `g`).
Besides if `os-mapping` is required, we need to add it for `java`.
Further, the last entry `/os-mappings` looks odd and seems to indicate another bug.

### Affected version:

* WIP of 2022.08.004

hohwille:
When I go to https://adoptium.net I get `17.0.4.1_1` as latest version.
[...]

Issue #893, opened 2022-09-16
Ability to configure version prefix


hohwille:
As a devonfw-ide user, I want to use latest security fixes of tools but still be in control of e.g. major version so that my project team can work reliable but also up-to-date with minimum maintenance effort.

So e.g. for my project I might want to decide for Java 11. However, whilst using Java 11, I would like to have always the latest version available for Java 11. Currently there is no option to configure this:
* Either I leave `JAVA_VERSION` unconfigured and once a new (even non LTS, kind of experimental) release gets available, my team will get it and things break
* Or I have to configure `JAVA_VERSION` to the exact version number to the last digit.

Therefore, the idea is that I can set e.g. `JAVA_VERSION=11*` or `JAVA_VERSION=11.0*`.
Then, with this story being implemented, devonfw-ide would notice that this from the asterisk in the version.
Therefore, devonfw-ide (`doInstall`) would choose the first version from `available-versions` starting with the given version but not immediately followed by a digit. So `JAVA_VERSION=11*` would match `11.0.16.1_1` but not `111.2.3`.

This slightly relates to issue #878

hohwille:
I have done some more thinking about this and came to the following concern:

There may be a tool X with versions such as `11` and `11.1`. Now if I set `X_VERSION=11` it would be ambiguous if I want to have exactly version `11` or the latest `11*` version. So it would be better to change this story in a way to avoid such ambiguity.
Therefore we should either add `*` as suffix to a version variable value in order to activate this feature or to end with a separator (`11.` would match `11.1` but `11` would only match `11` - but then also other separators such as `-` would need to be supported as well). Therefore IMHO using `*` as the common glob matching character would be the easiest but still most consistent solution.
In case of Java it might be slightly more intuitive for users to simply write `JAVA_VERSION=11` than `JAVA_VERSION=11*` (and with `*` we have to be careful to avoid accidental bash globbing) but as a generic solution this makes the most sense to me and also it seems to be quite explicit and readable as `11*` is obviously not a version but means `11` followed by anything (where the first character of "anything" is not a digit).

hohwille:
Works but needs some special cases to be considered:
```
*** Setting up eclipse ***
You are using eclipse version 2022-06 that requires Java 11 or newer.
However, [EXTRA_]JAVA_VERSION is 11* - in this setup eclipse can not work.
[...]

Issue #878, opened 2022-09-12
Use default tool version from `available-versions`


hohwille:
So far we either hardcode the default version of a tool:
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/vscode#L21
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/java#L27
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/mvn#L18
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/eclipse#L47
...

or we dynamically get it via curl:
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/helm#L30
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/docker#L45
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/oc#L33
https://github.com/devonfw/ide/blob/ebdeab1d0bb5a9ced96e914cab287153711c0a7f/scripts/src/main/resources/scripts/command/terraform#L30
...

The idea of this story is that if story #826 is implemented and working stable, that we change this:
IMHO we should simply pass the tool version to `doInstall` as version argument - e.g. `"${ECLIPSE_VERSION}"` for eclipse.
Now if that version is empty, then `doInstall` can automatically determine the latest version (via `head` command) from the according `available-versions` file from mirrors repo.
Also we should think about updating `mirrors` repo via git.
Ideally, we only trigger `git pull` (`doGitCloneOrPull`) only once per `doInstall`.

hohwille:
This issue reveals bug #950 what is a kind of showstopper for the new version features (PR #934)

Issue #927, opened 2022-10-05
Improve python download of available versions


sarahffm:
As a devon IDE user, I want to be able to download specific python versions without errors.
Currently, a number of versions in the available-versions file for Python can't be downloaded with the download URL specified in https://github.com/devonfw/ide-mirrors/blob/master/python/urls.

For windows the download URL looks like this when the extension and os mappings are filled in:
https://www.python.org/ftp/python/${version}/python-${version}-embed-win32.zip
But the actual download links (from the same website) of some versions have a different format, for example:
https://www.python.org/ftp/python/3.9.14/Python-3.9.14.tgz (or other extensions but not .zip)

Examples of versions that don't work with the current URL:
3.11.0, 
3.9.14, 
3.8.11 - 3.8.14, 
3.7.2,
...

Version 3.11.0 might be a pre-release or something like that. If that's really the case, it should actually not be listed in the available-versions file.

Related Issue: 
#826

alfeilex:
It seems that there are security bugfix releases (e.g. [3.9.14](https://www.python.org/downloads/release/python-3914/)) which don't provide embeddable zip files for windows to install. The installation on MacOS and Linux should work with all versions because the installation of Python there builds from source code with .tgz file and this format is provided by each version.

Maybe we can modify the python command to build Python from source on windows systems as well. But currently I have troubles to install python on Linux and probably on MacOS as well.

alfeilex:
I have started a discussion [here](https://discuss.python.org/t/url-of-labeled-python-releases/20175) to get information about available release url's.

alfeilex:
Here is the PR https://github.com/devonfw/ide-mirrors/pull/41 in case we need stable python versions after the next _doInstall_ merge.

alfeilex:
#963 will fix this issue. Issue can be closed.

alfeilex:
https://github.com/devonfw/ide/issues/963 will fix this issue. Issue can be closed.

PR #934, opened 2022-10-10
#878: default version from mirrors #893: version prefix #940: generic doInstall

hohwille:
Implements the following stories and features:
* #878 - default version from available-versions
* #893 - version prefix
* #940 - generic doIntall

Revealed bug #933 that is now fixed.


hohwille:
This is now working and fine.
However, this is a rather significant change that could easily break things.
Therefore, it would be good to do the following:
* do a review
* do manual (and automated) tests of each commandlet to verify that nothing broke
* explicitly test with a custom tool to see that this functionality is still working.

Besides I am considering more changes.
I already refactored `doInstall` significantly (number and order of parameters).
I could do the same also with `doDownload` for further cleanup and simplification but this is rather optional.
I am tempted to also implement #940 in this PR as it is very closely related.
However, we could still do it in a separate PR after merging this one...

hohwille:
#893 was now easy to add so I just implemented it :)
Works like a charm

hohwille:
Now ready for review and test.
I also tested Java, Eclipse, and IntelliJ (Community + Ultimate) on MacOS due to my changes for #940

hohwille:
Found a new bug:
```
Updating java from version 8u262b10 to version 11.0.13_8...
Starting installation of java in version 11.0.13_8 to /d/projects/zensus/software/java
Trying to download java-11.0.13_8-/d/projects/zensus/software/extra/java-windows.zip from https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.13%2B8/OpenJDK11U-jdk_x64_windows_hotspot_11.0.13_8.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  187M  100  187M    0     0  10.9M      0  0:00:17  0:00:17 --:--:-- 11.5M
...
Download of java-11.0.13_8-/d/projects/zensus/software/extra/java-windows.zip from https://github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.13%2B8/OpenJDK11U-jdk_x64_windows_hotspot_11.0.13_8.zip succeeded.
unzip:  cannot find or open /c/Users/hohwille/Downloads/devonfw-ide/java-11.0.13_8-/d/projects/zensus/software/extra/java-windows.zip, /c/Users/hohwille/Downloads/devonfw-ide/java-11.0.13_8-/d/projects/zensus/software/extra/java-windows.zip.zip or /c/Users/hohwille/Downloads/devonfw-ide/java-11.0.13_8-/d/projects/zensus/software/extra/java-windows.zip.ZIP.
```

from `devon.properties`:
```
JAVA_VERSION=8u262b10
EXTRA_JAVA_VERSION=11.0.13_8
```

So why is default java (`/d/projects/zensus/software/java`) updated to the `EXTRA_JAVA_VERSION` while it should be `JAVA_VERSION`?
Where does the `.ZIP` extension come from that is appended to existing `.zip` extension?

hohwille:
Problems fixed :)

hohwille:
Tested custom software feature and found next bug:
```
*** Setting up jboss-eap ***
Starting installation of jboss-eap in version 7.4.5.GA to /d/projects/test/software/jboss-eap

******** ATTENTION ********
No mirror is configured at /d/projects/test/mirrors/jboss-eap
```

Indeed there is no such mirror but in case of custom software this should not lead to an error as the repo for download is configured and shall be used.

hohwille:
OMG, stupid bug from myself. When I refactored `doInstall` I kicked out the repository arg first and then noticed that it is still required but forgot to revert one line. Actually declaring the same local variable twice in the function should lead to an error in bash what would have revealed the bug.

Also fixed now!

hohwille:
One last TODO is left in `functions`:
```
      # TODO could be removed and done in doResolveDownloadUrl
      mirror_url="${mirror_url//\$\{major\}/${major}}"
```

This `${major}` is currently only used in `java`. However, the current way it is done in `doDownload` is fully generic and IMHO fine. So if we might need it for some other tool in the future, we would to have to copy the solution to `doResolveDownloadUrl`.
Therefore I would tend to simply remove the `TODO` command and leave it as is.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -71,45 +67,26 @@ function doPackageInstall()
...
-      AWS_VERSION=$(curl "https://github.com/aws/aws-cli/tags" | awk -F'tags/' '/archive/ { print $2}' | sort -r | head -1 | awk -F'.zip' '{print $1}')
-    fi
-
-    doInstall "-" "${DEVON_IDE_HOME}/updates/install/aws" "aws" "${AWS_VERSION}"
-    ERR="${?}"
-    if [ "${ERR}" = "0" ]
-    then
-      doPackageInstall "${DEVON_IDE_HOME}/updates/install/aws" "${AWS_VERSION}"
-    fi
-    doRunCommand "rm -rf ${DEVON_IDE_HOME}/updates/install/aws"
```
hohwille, line 91: Should we keep this after `doPackageInstall`?
The problem is that we then do not have `.devon.software.version` file if we remove it.
When we keep it, `doInstall` will return `1` if the same version was already present and installed.
Otherwise on every setup `aws` will be reinstalled or we need to re-add more custom logic for AWS.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -71,45 +67,26 @@ function doPackageInstall()
...
-
 # CLI
 case ${1} in
 "help" | "-h")
   echo "Setup or run AWS CLI (command-line interface for Amazon-Web-Services)."
   echo
   echo "Arguments:"
   echo " setup                          install aws on your machine."
-  echo " �args�                         call aws with the specified arguments. Call aws --help for details or use aws directly as preferred."
+  echo " «args»                         call aws with the specified arguments. Call aws --help for details or use aws directly as preferred."
```
hohwille, line 85: where does this come from?
I will fix it...

hohwille: I do not have this locally and can therefore not fix it. Looks like some UTF-8 vs. ISO-Latin encoding conversion bug...

hohwille:
I am still having issues with `aws`:
```
$ devon aws
Starting installation of aws in version 2.8.3 to /d/projects/devonfw/updates/install/aws
Trying to download aws-2.8.3-windows.msi from https://awscli.amazonaws.com/AWSCLIV2-2.8.3.msi
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 29.7M  100 29.7M    0     0  11.4M      0  0:00:02  0:00:02 --:--:-- 11.5M
Download of aws-2.8.3-windows.msi from https://awscli.amazonaws.com/AWSCLIV2-2.8.3.msi succeeded.
Success: run command mv
Successfully moved archive aws-2.8.3-windows.msi to /d/projects/devonfw/updates/extracted/aws
Success: run command mv
Successfully installed aws
/d/projects/devonfw/scripts/command/../functions: line 85: aws: command not found

******** ATTENTION ********
Failed to verify installation of aws (aws)
Exit code was 127
```

Interesting fact is that there is no `Amazon` folder in `C:\Program Files` on my machine.
So whatever has been installed by the `msi` it either did nothing or at least not install `aws` in the expected location.

hohwille:
Via this bug I now finally figured out, why some artifacts are downloaded again every time while most others get reused from download directory as cache (unless `force` mode is used).

hohwille:
AWS fixed and tested (on Windows).

scripts/src/main/resources/scripts/command/aws
hohwille: For the record: When I got to the tab `Conversation` of this PR then the diff looks correct and nothing has changed about `«args»`.
However, in the tab `Files changed`, I see it changed to `Â«argsÂ»`.
I was referring the these broken `Â` characters that I do not have locally and that I do not understand.
IMHO this is an encoding bug of github when dealing with special unicode characters...
I hope that if we merge this, we do not get these `Â` characters into our code...

scripts/src/main/resources/scripts/command/aws
hohwille: I resolved this and kept the line to remove it.
@CREITZ25 has found an excellent solution to unpack MSI files. We will create a separate story to extend `doExtract` so that it can also unpack MSI files just like tar, zip, oder dmg files. Maybe we can also find a solution for extracting PKG files on Mac.
Here I simply try to keep compatibility with previous behavior.

hohwille:
I just got this one:
```
$ devon az setup
Trying to download pip-latest-pip.py from https://bootstrap.pypa.io/get-pip.py
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2509k  100 2509k    0     0  3306k      0 --:--:-- --:--:-- --:--:-- 3310k
Download of pip-latest-pip.py from https://bootstrap.pypa.io/get-pip.py succeeded.
D:\projects\devonfw\software\python\python.exe: can't open file 'D:\\projects\\devonfw\\software\\pip\\pip-latest-windows.py': [Errno 2] No such file or directory

******** ATTENTION ********
Failed to run command python
Exit code was 2
Command 'python pip-latest-windows.py --no-warn-script-location --target=/d/projects/devonfw/software/pip' failed with exit code 2

******** ATTENTION ********
Failed to run devon python pip-latest-windows.py --no-warn-script-location --target=/d/projects/devonfw/software/pip
Exit code was 2
Command 'pip install azure-cli az' failed with exit code 2

******** ATTENTION ********
Failed to run devon pip install azure-cli az
Exit code was 2
```

scripts/src/main/resources/scripts/command/aws
CREITZ25: The help option has been standardized for all commandlets. My change reversed that, so I copied and pasted the change back in. I used copy and paste because I couldn't type the original change because of the special characters. And what also surprised me is that the change was not displayed as a merge conflict in Github and I was able to fix it there.

scripts/src/main/resources/scripts/functions
```diff
@@ -62,29 +62,31 @@ function doResult() {
...
-  if [ -z "${2}" ]
+  if [ -z "${message}" ]
   then
-    message="run command ${1/ */}"
+    doDebug "Trying to run command ${command}"
+    message="run command ${command/ */}"
   else
-    message="${2} (${1})"
+    doDebug "Trying to ${message} (${command})"
+    message="${message} (${command/ */})"
```
hohwille, line 83: ssomething is wrong here, as I got this:
```
doRunCommand npm -v verify installation of npm
Message: run command tar (mv)
run command tar (mv) (npm)
8.19.2
Success: run command tar (mv) (npm)
```

hohwille:
```
Trying to download gcviewer-1.36-/Users/hohwille/projects/devon/workspaces/main/ide/scripts/target/integration-test/test-setup/software/gcviewer-noUnpack.jar from https://sourceforge.net/projects/gcviewer/files/gcviewer-1.36.jar
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   713    0   713    0     0   2145      0 --:--:-- --:--:-- --:--:--  2187
100   729    0   729    0     0   1152      0 --:--:-- --:--:-- --:--:--  1152
100   970    0   970    0     0   1044      0 --:--:-- --:--:-- --:--:--  1044
100   327  100   327    0     0    202      0  0:00:01  0:00:01 --:--:--  1837
100  549k  100  549k    0     0   271k      0  0:00:02  0:00:02 --:--:--  271k
mv: rename /var/folders/gk/7stbxyd17w3cfxw4m_4bfzlr0000gn/T/gcviewer-1.36.XXXXXX.dLFxGuBx to /Users/hohwille/Downloads/devonfw-ide/gcviewer-1.36-/Users/hohwille/projects/devon/workspaces/main/ide/scripts/target/integration-test/test-setup/software/gcviewer-noUnpack.jar: No such file or directoryTrying to download gcviewer-1.36-/Users/hohwille/projects/devon/workspaces/main/ide/scripts/target/integration-test/test-setup/software/gcviewer-noUnpack.jar from https://sourceforge.net/projects/gcviewer/files/gcviewer-1.36.jar
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   713    0   713    0     0   2145      0 --:--:-- --:--:-- --:--:--  2187
100   729    0   729    0     0   1152      0 --:--:-- --:--:-- --:--:--  1152
100   970    0   970    0     0   1044      0 --:--:-- --:--:-- --:--:--  1044
100   327  100   327    0     0    202      0  0:00:01  0:00:01 --:--:--  1837
100  549k  100  549k    0     0   271k      0  0:00:02  0:00:02 --:--:--  271k
mv: rename /var/folders/gk/7stbxyd17w3cfxw4m_4bfzlr0000gn/T/gcviewer-1.36.XXXXXX.dLFxGuBx to /Users/hohwille/Downloads/devonfw-ide/gcviewer-1.36-/Users/hohwille/projects/devon/workspaces/main/ide/scripts/target/integration-test/test-setup/software/gcviewer-noUnpack.jar: No such file or directory
```

hohwille:
gcviewer now working.
function `doIsPrefix` was added temporary and is now unused.
As it seems handy and useful, I do not remove it.

hohwille:
Latest fun fact:
```
      doDebug "${target_path}"
      doDebug "$(cygpath -w "${target_path}")"
      doDebug "$(cygpath -w ${install_path})"
```

gives me the following output:
```
/d/projects/test/software/java
D:\projects     est\software\java
D:\projects\software\default\java\11.0.13_8
```

So is bash drunk? It took me a while to guess an idea of what could go wrong.
IMHO `cygpath` is changing all slashes (`/`) to backslashes (`\`) and when this is going back into bash it is treated as escape character and `\t` is the escaped syntax of the tab character what explains the effect.
So the master question is how to call `cygpath` from bash without interpreting backslashes?

hohwille:
I usually do not like monster PRs that even implement multiple stories at once.
However, I am now happy with the achievement of this PR that brings us way ahead with great stuff to improve devonfw-ide. I am hoping to get this merged soon. Still waiting for more feedback (review & test) from the team...

I also have tested the `DEVON_SOFTWARE_PATH` feature intensively now on windows and mac as well as without and also the custom tools feature.

scripts/src/main/resources/scripts/functions
```diff
@@ -1257,21 +1306,61 @@ function doInstall() {
...
-    doRunCommand "cp '${target_software_software}/.devon.software.version' '${target_path}'"
   fi
-
+  if [ "${target_path}" = "${DEVON_IDE_HOME}/software/${dir}" ]
+  then
+    doExtendPath "${target_path}"
+  fi
+  if [ -n "${TOOL_VERSION_COMMAND}" ] && [ "${TOOL_VERSION_COMMAND}" != "-" ] && ! doIsQuiet
+  then
+    doRunCommand "${TOOL_VERSION_COMMAND}" "verify installation of ${software}"
```
tobka777, line 1362: Is it possible that the version of the tool will not be output if `TOOL_VERSION_COMMAND` is not defined, or is a future requirement that `TOOL_VERSION_COMMAND` should be defined in every commandlet. If so, I would ensure by a warning message or test that it is not forgotten.

tobka777 (approved):
I reviewed your changes and it all looks good. Only one question I have left.

scripts/src/main/resources/scripts/functions
hohwille: Yes, we have situations where `TOOL_VERSION_COMMAND` and then it is obviously intended that no such command is executed.
Examples are:
* we have commandlets that are not related to a "tool" at all (e.g. `build`). They should never run `doInstall` but there are other places where this variable is used so it also triggers if `version get|set|list` is supported or not.
* we have custom tool support and in such case we do not even know anything about the tool except "name" and "version" and where to download the *.tgz file that we extract. So this would be the case for the code you annotated with your review comment.

hohwille: > If so, I would ensure by a warning message or test that it is not forgotten.

If you forget this, then `devon «tool» version get|set|list` will all 3 not work.
We can easily test this and add it to our DoD.
Due to the explanation above we can not easily render a warning as this would also show up in undesired situations where nothing is actually wrong.

hohwille: DoD updated.

Amueller36:
I was testing the changes on the M1 with Alexander and we got the following error when trying to install vscode :
```
Success: run command unzip
Successfully extracted archive vscode-1.72.2-mac.zip to /Users/m1/Documents/Alex_Dev/test_doInstall/updates/extracted/vscode
Success: run command mv
Successfully installed vscode
Creating symlink as workaround for vscode on MacOS
Success: run command rm
Success: run command mv
Success: run command ln
Success: run command cp
The software vscode has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
/Users/m1/Documents/Alex_Dev/test_doInstall/scripts/command/../functions: line 85: code: command not found
```
We were trying to spot the issue  and  we think it might have something to do with the MacOs workaround in the functions file,
Line 1309 Following (functions).

hohwille:
> One other thing was that the recent available vscode versions got pulled in a wrong format : v14 v1.19.3 master-with-conflict-190418 1.72.2 1.72.1 1.72.0 1.71.2 1.71.1
> 
> The first 3 have a wrong format.

Thanks, see #959

hohwille:
> I was testing the changes on the M1 with Alexander and we got the following error when trying to install vscode :
> 
> ```
> Success: run command unzip
> Successfully extracted archive vscode-1.72.2-mac.zip to /Users/m1/Documents/Alex_Dev/test_doInstall/updates/extracted/vscode
> Success: run command mv
> Successfully installed vscode
> Creating symlink as workaround for vscode on MacOS
> Success: run command rm
> Success: run command mv
> Success: run command ln
> Success: run command cp
> The software vscode has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
> /Users/m1/Documents/Alex_Dev/test_doInstall/scripts/command/../functions: line 85: code: command not found
> ```
> 
> We were trying to spot the issue and we think it might have something to do with the MacOs workaround in the functions file, Line 1309 Following (functions).

Strange so `doUpdatePath` did not work somehow in your test scenario.
Can you reproduce this error?

CREITZ25:
I reviewed your changes too, and I have only a few comments/questions:

1. TERRAFORM HOME is set on line 35, but should also be included on line 14
2. NODE_HOME in the npm commandlet is set on line 114 in an if condition, but
    should also be included on line 14 without if
3. aws Commandlet: 
    It doesn't make much sense for me to create first your own functions in the commandlet
    and then source the functions file. This has the unpleasant side effect that you cannot use
    existing functions. Is it not better to change it back and use doIsWindows?

Amueller36:
> > I was testing the changes on the M1 with Alexander and we got the following error when trying to install vscode :
> > ```
> > Success: run command unzip
> > Successfully extracted archive vscode-1.72.2-mac.zip to /Users/m1/Documents/Alex_Dev/test_doInstall/updates/extracted/vscode
> > Success: run command mv
> > Successfully installed vscode
> > Creating symlink as workaround for vscode on MacOS
> > Success: run command rm
> > Success: run command mv
> > Success: run command ln
> > Success: run command cp
> > The software vscode has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
> > /Users/m1/Documents/Alex_Dev/test_doInstall/scripts/command/../functions: line 85: code: command not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We were trying to spot the issue and we think it might have something to do with the MacOs workaround in the functions file, Line 1309 Following (functions).
> 
> Strange so `doUpdatePath` did not work somehow in your test scenario. Can you reproduce this error?

image

Yes the error is reproducable.
We get it everytime we try to run "devon vscode setup" on the M1.

hohwille:
@CREITZ25 also thanks for your intensive testing 👍 

# Cobigen

> No mirror is configured at /c/Users/creitz/Projects/devonfw2022.08.003/mirrors/https://repo.maven.apache.org/maven2/com/devonfw/cobigen/cli/2021.12.006/cli-2021.12.006.tar.gz

OMG. This is a severe bug of my refactoring. I will fix this asap. Thanks.

# Python (incl. pip and az)

> Failed to download https://www.python.org/ftp/python/3.11.0/python-3.11.0-embed-win32.zip with exit code 22

Problem of mirrors. See #927 - @alfeilex is already working on this.
Pip itself did only first try to install python and that failed, so not really a problem of `pip`. Same for `az`.

# Eclipse

> Download of eclipse in version 2022-12 failed after trying 3 mirror(s).

OK, when I tested this, I still got version 2022-06. Maybe I had `ECLIPSE_VERSION` configured.
Again our mirrors workflow is broken as there is no release `2022-12` available (yet) - we can easily guess this by the version that is a date in the future. Latest version can be checked here:
https://www.eclipse.org/downloads/

# VSCode

> Download of vscode in version v14 failed after trying 1 mirror(s).

Jep, reported mutliple times. See #959

# Rancher-Desktop (docker, kubectl, lazydocker)

> Failed to download https://github.com/rancher-sandbox/rancher-desktop/releases/download/v1.6.1/Rancher.Desktop.Setup.1.6.1.exe with exit code 22
> ******** ATTENTION ********
> Download of rancher in version 1.6.1 failed after trying 1 mirror(s).

Great catch. Also seems to be a bug of `ide-mirrors`.
As we can see here, the latest version is still `1.6.0` and not `1.6.1`:
https://github.com/rancher-sandbox/rancher-desktop/releases/

# AWS

> /c/Users/creitz/Projects/devonfw2022.08.003/scripts/command/../functions: line 85: aws: command not found
> Failed to verify installation of aws ('aws')

OK. Here I made a logical design mistake:
We set `TOOL_VERSION_COMMAND` to tell the `functions` how to do the tool version check.
This is triggered both by "version get" (`doGetSoftwareVersion`) as well as by installation (`doInstall`).
However, here we are using `doInstall` to "install" the installer currently.
The execution of `TOOL_VERSION_COMMAND` can never succeed if we just downloaded the installer.
However, for "version get" the `TOOL_VERSION_COMMAND`  is still correct.
So maybe we have to set `TOOL_VERSION_COMMAND="-"` temporary in `doSetup` of `aws` before calling `doInstall` and after `doPackageInstall` we can call `findAws` again.

> It doesn't make much sense for me to create first your own functions in the commandlet
and then source the functions file. This has the unpleasant side effect that you cannot use
existing functions. Is it not better to change it back and use doIsWindows?

I was fully aware of this "unpleasant effect". However, I have to set `TOOL_VERSION_COMMAND` before sourcing the `functions` and therefore I can not use `doIsWindows` before. If you come up with a better solution please let me know, but I can not see any.

# Terraform

> TERRAFORM HOME is set on line 35, but should also be included on line 14

I had already done that: https://github.com/devonfw/ide/pull/934/commits/be20341c02daaded0160ae169903a786101da7d4
In case you have review comments on specific code, it is much better and easier to discuss if you directly annotate the code in the "Files changed" tab of the PR. Then we have a single thread per code-line and can discuss this and track it until it is resolved. This comment already shows that it is quite complex to address all the comments that had been added meanwhile...

# Node

> NODE_HOME in the npm commandlet is set on line 114 in an if condition, but should also be included on line 14 without if

OK. Also nice catch. Has nothing to do with this PR but interesting.
IMHO the `if` is wrong in `npm` and we could also move this up to the top.
If we continue like this this monster PR will grow bigger and bigger but I am happy to also include this change.

hohwille:
@Amueller36 thanks for re-testing and further details.

> Yes the error is reproducable.
We get it everytime we try to run "devon vscode setup" on the M1.

Great. If an error is reproducible we are already close to a fix.
After reproducing the very next step would have been to check why it is not there or what is there instead.
I have now booted my mac and reproduced the bug you have found:

```
ls -la software/vscode
... software/vscode -> macos/vscode/Contents/MacOS
ls software/macos/vscode
Contents
ls -F software/macos/vscode/Contents
CodeResources Frameworks/ Info.plist MacOS/ PkgInfo Resources/ _CodeSignatures/
```

OK. So with this info it is getting clear:
the macos workaround implementation found out that we have a MacOS App with `Contents` folder.
It found a folder with a binary inside (`MacOS` folder with `Electron`) and linked it.
However, it should have found and linked the folder `Contents/Resources/app` containing a `bin` subfolder.
The Problem here is that this folder is one level deeper and the current implementation is only searching for folders directly in `Contents`. So again no conventions or best practices followed here so we have more effort to get it right...

hohwille:
I improved the MacOS workaround and the symlink is created as desired.
However, now I get the next error:
``` 
Unable to determine app path from symlink : /Users/hohwille/projects/devon/workspaces/main/ide/scripts/target/integration-test/test-setup/software/vscode/bin/code
```
So in other words VSCode can not deal with our symlink workaround.
Any new ideas on this?
https://github.com/microsoft/vscode/blob/07655f3a23b9b168e7ed351a754fc77bd7faf3ff/resources/darwin/bin/code.sh#L28

Otherwise we have to return to the old approach for vscode...

CREITZ25 (commented):
Looks good

hohwille:
> OMG. This is a severe bug of my refactoring. I will fix this asap. Thanks.

Gotcha:
https://github.com/devonfw/ide/blob/d718d4586baa4d5eb42472973b35298d102b2936/scripts/src/main/resources/scripts/functions#L1047

So I refactored `doInstall` and missed the above usage that is now entirely broken.
Actually I removed the `URL` arg parameter entirely - not so easy to fix.
IMHO we should also refactor cobigen to use `doInstall` and `mirrors`.

hohwille:
```
command/cobigen:    doUpgradeMavenArtifact "${COBIGEN_CLI_HOME}" "https://repo.maven.apache.org/maven2/com/devonfw/cobigen" "cli" "${COBIGEN_VERSION:-LATEST}" ".tar.gz"
command/ide:  doUpgradeMavenArtifact "${DEVON_IDE_HOME}" "${DEVON_IDE_REPO_URL}" "devonfw-ide-scripts" "${target_version}" ".tar.gz" "$[devon_ide_version]"
command/ide:  doUpgradeMavenArtifact "${create_path}/" "${DEVON_IDE_REPO_URL}" "devonfw-ide-scripts" "${target_version}" ".tar.gz"
functions:function doUpgradeMavenArtifact() {
```

So most probably also `devon ide update scripts` will be broken as well.
If we would have merged and released before finding this, I had created a disaster as then nobody could update from that release to a newer version automatically...

hohwille:
> So in other words VSCode can not deal with our symlink workaround. Any new ideas on this? https://github.com/microsoft/vscode/blob/07655f3a23b9b168e7ed351a754fc77bd7faf3ff/resources/darwin/bin/code.sh#L28
> 
> Otherwise we have to return to the old approach for vscode...

This is the last real bug of this PR.
Interesting is that we also did some things wrong before:
https://github.com/devonfw/ide/blob/d9ebcad8bdc0c43bb7a205580b3212c2f9c6ebdd/scripts/src/main/resources/scripts/command/vscode#L34

If we compare this with the official launch script on MacOS:
https://github.com/microsoft/vscode/blob/07655f3a23b9b168e7ed351a754fc77bd7faf3ff/resources/darwin/bin/code.sh#L34

We can see that we are lacking the `cli.js` and the `--ms-enable-electron-run-as-node` args.
This might be related to #943

Unfortunately the `code` launch script provided by Microsoft to run VSCode assumes that it has to be inside a folder named `*.app`.
However, with our approach this is not the case and therefore the `code` script is producing this error.
Options:
* Should we fix this by "hacking" their own launch script?
* Should we tweak `doInstall` and the macos workaround to not "eliminate" the `*.app` folder?
* Should we go back to the old behavior at least for `vscode`?

hohwille:
I also implemented the change to keep the `*.app` folder to make the `code` launch script happy.
However, this also does not help or work:
```
/Users/hohwille/projects/devon/workspaces/main/ide/scripts/target/integration-test/test-setup/software/vscode/bin/code: line 2: /Users/hohwille/projects/devon/workspaces/main/ide/scripts/target/integration-test/test-setup/software/vscode/Visual Studio Code.app/Contents/MacOS/Electron: No such file or directory
```

So as you can see the `code` launch script is still bugy and does not resolve the symlink properly via `readlink`.
IMHO we have to give up here and go back to the old approach for vscode even though it would be create to get the quirks and workarounds out of the commandlets...

hohwille:
IMHO now I got everything right and fixed without ugly hacks.
The MacOs workaround seems to work in a generic way as initially planned for #940
I tested eclipse, vscode and java on my Mac and it worked now.

Side-node: If you are using shared software feature (`DEVON_SOFTWARE_PATH`) and then update devonfw-ide to the next comming release things could break when still having the older release in parallel on MacOS as the folder layout changed for MacOS apps. However, this is such an exotic situation that I do not care about it.

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -151,7 +151,7 @@ do
   if [ "${1}" = "run" ] || [ "${1}" = "start" ]
   then
     doSetup silent
-    doStartVsCode
+    doStartVsCode "${@}"
```
hohwille, line 154: nice but then you need to call `shift` before as otherwise `run` or `start` will be passed to `doStartVsCode`.

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -166,8 +166,7 @@ do
     doCreateIdeScript vscode
   elif [ "${1}" = "add-plugin" ]
   then
-    shift
-    doAddPlugin "${@}"
+    doAddPlugin "--install-extension ${2}"
```
hohwille, line 169: shouldn't the `--install-extension` option be added inside `doAddPlugin` function?

hohwille:
> [ERROR LOG] test-functions-doc
73d72
< does not have options
Documentation of functions does not match function declarations!
Please keep function documentation and declarations in sync!

Oh, newline diff, need to set option to ignore

> [ERROR LOG] test-shellcheck
In scripts/command/help line 24:
echo $@
     ^-- SC2068: Double quote array expansions to avoid re-splitting elements.

Awesome. Forgot to remove debug output. Thanks shellcheck for finding it. :)

scripts/src/main/resources/scripts/command/vscode
Amueller36: There are 2 Functions the doAddPlugins and doAddPlugin

The doAddPlugins would break if I add the --install-extension into the doAddPlugin Function

because the doAddPlugins Functions passes a String with multiple plugins to the doAddPlugin.
Multiple plugins have to be installed with the following format : --install-extension plugin1 --install-extension plugin 2 ...
If I add the "--install-extension" into the doAddPlugin function it would call vscode with --install-extension --install-extension which would lead to an error.

scripts/src/main/resources/scripts/command/vscode
hohwille: Done.

scripts/src/main/resources/scripts/command/vscode
hohwille: Refactored to support multiple plugin-ids as parameter.

CREITZ25 (commented):
Looks fine

scripts/src/main/resources/scripts/functions
```diff
@@ -1257,21 +1306,77 @@ function doInstall() {
...
-  if [ ! -d "${target_path}/bin" ] && doIsMacOs && [ -d "${target_path}/Contents/Home/bin" ]
+  if [ ! -d "${target_path}/bin" ] && doIsMacOs
   then
-    doEcho "Creating symlink as workaround for ${software} on MacOS"
-    local target_software_macos=${target_path%\/*}/macos
-    if [ ! -d "${target_software_macos}" ]
+    local contents="Contents"
+    if ! [ -d "${target_path}/${contents}" ]
+    then
+      pushd "${target_path}" > /dev/null || exit 255
```
alfeilex, line 1314: I tried the symlink functionality on the M1 and get the following error message: 

```
m1@3dc6a48e-5e10-4e97-a481-1c8483f2748d test_doInstall % devon java setup
Success: run command mkdir
Starting installation of java in version 19_36 to /Users/m1/Documents/Alex_Dev/test_doInstall/software/java
Trying to download java-19_36-mac.tar.gz from https://github.com/adoptium/temurin19-binaries/releases/download/jdk-19%2B36/OpenJDK19U-jdk_aarch64_mac_hotspot_19_36.tar.gz
Artifact already exists at /Users/m1/Downloads/devonfw-ide/java-19_36-mac.tar.gz
To force update please delete the file and run again.
Success: run command tar
Successfully extracted archive java-19_36-mac.tar.gz to /Users/m1/Documents/Alex_Dev/test_doInstall/updates/extracted/java
Success: run command mv
Successfully installed java
Linking ../symlinktest/default/java/19_36 to /Users/m1/Documents/Alex_Dev/test_doInstall/software/java
/Users/m1/Documents/Alex_Dev/test_doInstall/scripts/command/../functions: line 1314: pushd: /Users/m1/Documents/Alex_Dev/test_doInstall/software/java: No such file or directory
```

scripts/src/main/resources/scripts/command/az
```diff
@@ -9,10 +9,11 @@ then
   fi
   exit
 fi
-if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
-TOOL_VERSION_COMMAND="az -v"
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
+TOOL_VERSION_COMMAND="az -v"
+# shellcheck source=scripts/commandlet-cli
+source "$(dirname "${0}")"/../commandlet-cli
```
cinnamon-coder-hub, line 17: Firstly, it may be that someome recently mentioned this problem(?)
When trying to setup azure on linux mint the setup of python is needed. 
It seems as if the python download url changed slightly, but significantly.
It turned out, that the version in the download-url isn't 3.12.0, as our script would expect, but 3.12.0a1. 
Expected download-url: https://www.python.org/ftp/python/3.12.0/Python-3.12.0.tgz .
Actual download-url: https://www.python.org/ftp/python/3.12.0/Python-3.12.0a1.tgz .
Therefore the download just fails. This is another argument for an manually editable list of versions I guess.

scripts/src/main/resources/scripts/command/cobigen
```diff
@@ -10,10 +10,11 @@ then
   exit
 fi
 
-if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
-TOOL_VERSION_COMMAND="cg -v"
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
+TOOL_VERSION_COMMAND="cg -v"
+# shellcheck source=scripts/commandlet-cli
+source "$(dirname "${0}")"/../commandlet-cli
```
cinnamon-coder-hub, line 18: After the installation the verification of the installation of cobigen is handled via `cobigen -v`, which gives 
`Unknown option: '-v'` as output. 
So I guess, as also suggested as possible solution by the devon ide, that the verification should be done by `cobigen --version` instead.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -118,48 +124,42 @@ function doCheckPluginsInstalled() {
...
-        echo "Doing workarounds for MacOS quirks..."
-        echo -e "#!/usr/bin/env bash\nECLIPSE_HOME=\"\$(dirname \"\${0}\")\"\n\"\${ECLIPSE_HOME}/Eclipse.app/Contents/MacOS/eclipse\" \$@" > "${ECLIPSE_HOME}/eclipse"
-        chmod a+x "${ECLIPSE_HOME}/eclipse"
+        doConfirmWarning "You are using eclipse version ${ECLIPSE_VERSION:-latest} that requires Java 11 or newer.\nHowever, [EXTRA_]JAVA_VERSION is ${javaVersion} - in this setup eclipse can not work.\nPlease update JAVA_VERSION, set EXTRA_JAVA_VERSION or downgrade ECLIPSE_VERSION in your settings/devon.properties!"
       fi
     fi
-    doAddPlugins
   fi
-  if [ ! -x "${ECLIPSE_HOME}/eclipse" ]
+  doInstall "eclipse" "${ECLIPSE_VERSION}" "${silent}" "${ECLIPSE_EDITION_TYPE:-java}"
```
cinnamon-coder-hub, line 159: Sadly regarding eclipse one folder in the eclipse download urls changed from "R" to "M2" or "M1", at least for version 2022-12. As an example the download-urls for linux: 
Generated with ide-mirrors: https://ftp.snt.utwente.nl/pub/software/eclipse/technology/epp/downloads/release/2022-12/R/eclipse-java-2022-12-R-linux-gtk-x86_64.tar.gz .
Actual url from utwente.nl:https://ftp.snt.utwente.nl/pub/software/eclipse/technology/epp/downloads/release/2022-12/M2/eclipse-committers-2022-12-M2-linux-gtk-aarch64.tar.gz .
Actual url from eclipse.com: https://archive.eclipse.org/technology/epp/downloads/release/2022-12/M1/eclipse-dsl-2022-12-M1-linux-gtk-x86_64.tar.gz .

scripts/src/main/resources/scripts/command/intellij
```diff
@@ -14,49 +14,32 @@ then
...
-source "$(dirname "${0}")"/../functions
+# shellcheck source=scripts/commandlet-cli
+source "$(dirname "${0}")"/../commandlet-cli
 cd "${DEVON_IDE_HOME}" || exit 255
 
 function doSetup() {
   doDevonCommand java setup silent
-  if [ "${1}" != "silent" ] || [ ! -d "${IDEA_HOME}" ]
+  doInstall "intellij" "${INTELLIJ_VERSION}" "${1}" "${INTELLIJ_EDITION_TYPE:-C}"
+  if [ "${?}" = 0 ]
```
cinnamon-coder-hub, line 34: The download mirrors

- https://ftp.osuosl.org/idea/ideaIC-2022.2.3.tar.gz
- https://mirror.math.princeton.edu/idea/ideaIC-2022.2.3.tar.gz

don't work anymore.

scripts/src/main/resources/scripts/command/ionic
```diff
@@ -10,10 +10,11 @@ then
...
-if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
-TOOL_VERSION_COMMAND="ionic -v"
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
+TOOL_VERSION_COMMAND="ionic -v"
+# shellcheck source=scripts/commandlet-cli
+source "$(dirname "${0}")"/../commandlet-cli
 
 # $1: optional setup
 function doSetup() {
```
cinnamon-coder-hub, line 20: During the setup of npm a 'MODULE_NOT_FOUND' blocks the  installation of ionic. 
To be more precise, I've been getting the following error after the installation of node seemed
to be complete:


`node:internal/modules/cjs/loader:998
  throw err;
  ^

Error: Cannot find module '/home/dr/projects/PR934/software/node/bin/node_modules/npm/bin/npm-cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
    at Module._load (node:internal/modules/cjs/loader:841:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:82:12)
    at node:internal/main/run_main_module:23:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v19.0.0
Could not determine Node.js install directory

******** ATTENTION ********
Failed to install ionic-cli ('/home/dr/projects/PR934/software/node/npm')
Exit code was 1
`


This error could  be unrelated to the changes of this PR,
but seems to be worth mentioning.

cinnamon-coder-hub: It seems as if '/home/dr/projects/PR934/software/node/bin/node_modules/npm/bin/npm-cli.js' would be just the wrong location, as the file 'npm-cli.js' exists, but with one bin directory less in the path. So the proper path in my case is '/home/dr/projects/PR934/software/node/bin/node_modules/npm/bin/npm-cli.js'.

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -14,27 +14,19 @@ then
...
+VSCODE_HOME="${DEVON_IDE_HOME}/software/vscode"
+TOOL_VERSION_COMMAND="code -v"
+# shellcheck source=scripts/commandlet-cli
+source "$(dirname "${0}")"/../commandlet-cli
+
 cd "${DEVON_IDE_HOME}" || exit 255
 
 function doSetup() {
-  if [ "${1}" != "silent" ] || [ ! -d "${VSCODE_HOME}" ]
+  doInstall "vscode" "${VSCODE_VERSION}" "${1}"
```
cinnamon-coder-hub, line 27: It seems that the download url of VS Code for Linux changed, so the installation doesn't work. 
As new source for the download url https://code.visualstudio.com/sha could be an option.

cinnamon-coder-hub:
As some of the problems mentioned before may be related to Linux or just have to be settled at some point, 
I could tackle them, or a subset, after the issue #941 progressed enough. Maybe this would even be advantageous for writing methods handling the differing download urls for the issue #941.

scripts/src/main/resources/scripts/command/vscode
cinnamon-coder-hub: For the most recent stable version of VSCode on debian-based linux the url currently looks like this: 
https://code.visualstudio.com/sha/download?build=stable&os=linux-deb-x64

scripts/src/main/resources/scripts/command/intellij
hohwille: See here:
https://github.com/devonfw/ide-mirrors/blob/master/intellij/urls

It seems that these two mirrors dropped idea so we need to remove them.

scripts/src/main/resources/scripts/command/ionic
hohwille: > So the proper path in my case is '/home/dr/projects/PR934/software/node/bin/node_modules/npm/bin/npm-cli.js'.

Did you mean '/home/dr/projects/PR934/software/node/bin/node_modules/npm/npm-cli.js'?

scripts/src/main/resources/scripts/command/cobigen
hohwille: I could not test this locally as I get this error in both cases:
https://github.com/devonfw/cobigen/issues/1591

However, I changed it as you suggested.

scripts/src/main/resources/scripts/command/eclipse
cinnamon-coder-hub: Yeah, didn't know the meaning of the M before our meeting today. So, the ide should just take the last stable version instead :+1:

scripts/src/main/resources/scripts/command/ionic
alfeilex: Full log of this error can be found here: https://github.com/alfeilex/ide/actions/runs/3368418467/jobs/5586938443#step:4:11593

cinnamon-coder-hub: No, but I actually took away the wrong bin folder. In my case the npm-cli.js file has the following path
/home/dr/projects/PR934/software/node/node_modules/npm/bin/npm-cli.js .

alfeilex:
On MacOS, Python is not working. See here:

```
Warning: openssl@3 3.0.5 is already installed and up-to-date.
To reinstall 3.0.5, run:
  brew reinstall openssl@3
Success: run command brew
sh: /Users/m1/Documents/Alex_dev/test_doInstall/workspaces/main/test/scripts/target/integration-test/test-setup/software/python/configure: No such file or directory
make: *** No targets specified and no makefile found.  Stop.
make: *** No rule to make target `altinstall'.  Stop.
make: *** No rule to make target `clean'.  Stop.
make: *** No rule to make target `distclean'.  Stop.
m1@3dc6a48e-5e10-4e97-a481-1c8483f2748d test-setup % which python
/usr/bin/python
m1@3dc6a48e-5e10-4e97-a481-1c8483f2748d test-setup % where python
/usr/bin/python
m1@3dc6a48e-5e10-4e97-a481-1c8483f2748d test-setup % devon python
Version 3.10.8 of python is already installed.
Python 2.7.18
Success: verify installation of python (python)
```

When Python is installed globally, `devon python` appears to run the global Python executable.

Full log can be found here: https://github.com/alfeilex/ide/actions/runs/3368406824/jobs/5586914231#step:5:831

scripts/src/main/resources/scripts/command/ionic
alfeilex: https://github.com/devonfw/ide/blob/11db4ef9862f5f7baa8b89e9f3c14d15bd0380dd/scripts/src/main/resources/scripts/command/ionic#L29-L33

It seems that the errors occurs because the `npm_command` uses the path from line 29, but it must be the path of line 32. Probably the condition in line 30 needs to be changed to (e.g. `! doIsWindows`) or something like that

alfeilex:
More MacOS errors related to this MacOS workaround: https://github.com/devonfw/ide/blob/11db4ef9862f5f7baa8b89e9f3c14d15bd0380dd/scripts/src/main/resources/scripts/functions#L1358-L1368

**Possible error**
This workaround also runs for commands like `terraform`, `oc` or `dotnet` and probably should not be run. I was not able to fix this error. This is the error output:
```
Creating symlink as workaround for terraform on MacOS
Success: run command rm
Success: run command mv
Success: run command ln
cp: /Users/m1/Documents/Alex_Dev/test_doInstall/workspaces/main/test/scripts/target/integration-test/test-setup/software/terraform/.devon.software.version and /Users/m1/Documents/Alex_Dev/test_doInstall/workspaces/main/test/scripts/target/integration-test/test-setup/software/macos/terraform/.devon.software.version are identical (not copied).
```

Full log: https://github.com/alfeilex/ide/actions/runs/3368406824/jobs/5586914231#step:5:841

scripts/src/main/resources/scripts/command/ionic
cinnamon-coder-hub: @alfeilex that's just the code part that needed to be fixed :+1: 
I tried it out with  `[ -x "${npm_command}" ]` instead of `[ ! -x "${npm_command}" ]` and the setup of ionic works now. 
If macs needs the same path, then `! doIsWindows`  seems to be a better readable way. 
Maybe @hohwille knows more about this?

scripts/src/main/resources/scripts/command/npm
```diff
@@ -10,32 +10,20 @@ then
...
+TOOL_VERSION_COMMAND="npm -v"
+# shellcheck source=scripts/commandlet-cli
+source "$(dirname "${0}")"/../commandlet-cli
 
 # $1: optional setup
 function doSetup() {
-  local software_version=${NPM_VERSION}
-  local installed_version
-  if [ -x "${NPM_CMD}" ]
+  doDevonCommand node setup silent
```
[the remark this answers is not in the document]
cinnamon-coder-hub: The problem doesn't look like being directly related to the code for npm. Compare with [this conversation](https://github.com/devonfw/ide/pull/934/files#r1012872348).

scripts/src/main/resources/scripts/command/ionic
cinnamon-coder-hub: From speaking with @alfeilex now we should most probably use `! doIsWindows` instead, as otherwise the npm_command variable would falsely also be changed for windows.

hohwille: This code pattern is actually bad. We have several npm based commandlets that would need such C&P logic.
Instead we should simply change to `doDevonCommand npm install ...`.
Then only `npm` commandlet should implement the logic how and where to find `npm`.
I can do that.

For the initial comment removing the `!` does not make any sense.
The idea is that we have two options where `npm` can be located. If the first one does not apply because it does not point to an executable we use the second one as fallback.
Maybe you have been facing problems because `npm` was not installed when this decision was made and after install `${npm_command}` was used without doing the check again. We have solved this problem in `aws` with an own `findAws` function that can be called again.

hohwille:
> So there is a bug in the MacOS Workaround detection.

I found and fixed it. The function `doGetFirstExistingPath` does nothing if no existing path found was found. Hence `${contents}` is empty but this was not checked and obviously `[ -d "${target_path}/${contents}" ]` applies then.
Problem resolved. Thanks for finding this!

hohwille:
BTW: Starting from here I stop accepting more errors in this PR that are not related to my changes at all. Otherwise, if we try to find and fix all bugs of the last two years, we will never be able to finish this "monster pr". Do not get me wrong: Your feedback is extremely valuable and you found important bugs I made in this PR. However, we need to focus on getting this PR merged and with #963 fixed so that we can reach a working state. It is fine and more than welcome to file to file more issues for bugs we might have in devonfw-ide but if they are not related to the changes of this PR, we should keep them out of here.

scripts/src/main/resources/scripts/command/ionic
hohwille: I created bug #966 for this. IMHO we need to simply switch the two paths as the second should be used in favor if it exists. However, this has nothing to do we this PR so we can fix it in a separate PR.

scripts/src/main/resources/scripts/functions
hohwille: I updated the documentation and hope that my guess was correct. Therefore I resolve this conversation.
If my guess was incorrect, please file a new bug issue for this problem that is IMHO not related to this PR.

hohwille:
I tested this PR again with the latest code and after #963 being fixed.
Now everything looks good to me except python (what is most probably not related to this PR at all):
```
Successfully installed python
The software python has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
Freeing Python from embedded Mode...
Traceback (most recent call last):
  File "runpy.py", line 196, in _run_module_as_main
  File "runpy.py", line 86, in _run_code
  File "D:\projects\devonfw\software\pip\bin\pip.exe\__main__.py", line 4, in <module>
ModuleNotFoundError: No module named 'pip'

******** ATTENTION ********
Failed to pip install azure-cli (pip)
Exit code was 1
```

Issue #912, opened 2022-09-29
Stability issue of nightly tests


hohwille:
As PO of devonfw-ide, I want automated integration tests so that we can deliver quality and get feedback if some PR merge breaks our product by causing a bug.
Therefore we have implemented #691 and related stories.
The problem now is that we have added all available commands and that takes too long.
Github gives a lot of resources for free to support open-source projects.
Obviously there are limits and Microsoft can not give every project unlimited computing time (otherwise people would also mine bitcoins with GHA, etc.).
As it turns out, our tests run slow and after 6:00 hours github simply cancels the action what seems fair.

Questions:
1. Why does it take more than 6:00 h to run our tests? Are also the resources (CPU/bandwith/...) throttled so hard? I did not notice this so far. Maybe there is a problem on our side or one special commandlet is taking too long and test should be removed.
2. Should we focus on the most important commandlets if we have limits?
3. Could we find a way to run selected commandlets (e.g. by day of the week) so every commandlet is covered but not all in the same nightly-build?
4. Can we improve our logging so we can see easier how long the test took per command and what the status is? My initial idea was already that we have a final summary with all the tests we have run including their status (OK/NOK, duration). If we get cancelled by github should we print out this summary after each test so we spam the log more but also can see intermediate status?

ahmedagdmoun:
After a small analysis and investigation I came with the following results:
The GHA resources are limited. As stated in Github Documentation (https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration): "Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete." So the workflows were canceled because the execution time of some jobs took more than 6 hours.
1. Unfortunately, there are different reasons why the tests took 6 hours depending on the OS.
For Ubuntu and MacOS, some commands (different commands) failed to run and the question "Do you want to continue?" was prompted out and waited infinitely for an answer till the 6 hours were reached. So we just have to find a way to automate skipping this question every time it's prompted out.
![ubuntu-test-failure](https://user-images.githubusercontent.com/98025282/194036877-1799139b-c42e-401a-92af-6a48baece879.PNG)
![MacOS-test-failure](https://user-images.githubusercontent.com/98025282/194036922-b60292ac-3739-4279-a539-a3976eb9abed.PNG)
For Windows, There was a full thread dump while installing Rancher Desktop. As this was already expected, I would suggest to simply exclude Rancher from the tests.
[...]

Issue #935, opened 2022-10-10
doDevonCommand reconsider doAskToContinue


hohwille:
This issue is about reconsidering this line of code:
https://github.com/devonfw/ide/blob/d72d893239cacf7b67b233b2b18880511ffe9c26/scripts/src/main/resources/scripts/functions#L328

If `doDevonCommand` is called and fails it will ask the user to continue.
This currently made our build stalled as the automated tests ran into this situation and the build was waiting for user-input that could never happen.

We should question why this behavior has been introduced.
Maybe it makes sense if `devon release` calls other functions and the user might want to continue in some cases.
However, IMHO this is not a good default behavior and should only apply in very special situations.

Finally, if we keep interactive questions like this, we need to run all our tests strictly in batch mode.
Even better devonfw-ide could check if the environment can actually never provide user input (no tty present, `${CI}=true`, etc.) and enable batch mode automatically in such case.

hohwille:
For the record:
These "special situations" are already using `doDevonCommandAndReturn":
https://github.com/devonfw/ide/blob/d72d893239cacf7b67b233b2b18880511ffe9c26/scripts/src/main/resources/scripts/command/release#L110

So IMHO there is no need left for this `doAskToContinue`.
PR #936 created.

PR #937, opened 2022-10-11
#912: Stability issue of nightly tests

alfeilex:
I had to remove all commands with Docker dependencies. Now the workflow runs to the end without interuption. But there are some errors for each OS which are detected.

Log for Windows: https://github.com/alfeilex/ide/actions/runs/3225608568/jobs/5278124299#step:5:396

Log for Ubuntu: https://github.com/alfeilex/ide/actions/runs/3225608567/jobs/5278124311#step:4:396

Log for MacOs: https://github.com/alfeilex/ide/actions/runs/3225472600/jobs/5277822159#step:5:396

I will test the commands and create some issues for each error.

Further, all errors are now summarized in the end of all-tests. I used some colors as in functions to highlight success and error outputs.


scripts/src/test/bash/functions-test
```diff
@@ -3,6 +3,21 @@
 function doCommandTest()  {
   cd ../test-setup
   CLI="${PWD}/scripts/devon"
-  "${CLI}" "${1}" setup
+  echo yes | "${CLI}" "${1}" setup
```
hohwille, line 6: can you simply run in batch mode (`--batch` or `-b`) instead?
FYI: I have meanwhile created #935 and implemented it already.

scripts/src/test/bash/functions-test
```diff
@@ -3,6 +3,21 @@
...
   exit "${?}"
-}
\ No newline at end of file
+}
+
+# $@: success message
+function doSuccess() {
+  echo -e "\033[92m${*}\033[39m"
```
hohwille, line 12: Cool! I was not aware that this also works in github logs. Awesome!

scripts/src/test/bash/all-tests
```diff
@@ -12,15 +14,17 @@ do
   mkdir -p "${testcase}/conf"
   cd "${testcase}"
   echo "export M2_REPO=~/.m2/repository" > "conf/devon.properties"
-  "${testpath}"
+  "${testpath}" &> errorLog
```
hohwille, line 17: so on success we still get the log. Could cause some spam and make it harder to read but with the coloring should be fine now.

hohwille (changes requested):
@alfeilex cool. Thanks for this great improvement.
I can also edit directly to incorporate the batch mode and other comments are just for notice so I can immediately set them to resolved...

scripts/src/test/bash/functions-test
hohwille: I just updated this so I can approve now...

hohwille (approved):
Ready for merge now.

PR #938, opened 2022-10-11
#847: #826: fixes

hohwille:
Small fixes and improvements to changes from #826 and part of #933:
* when extracting `doUpdateMirrors` I also extracted the `mirrors_dir` variable what caused some confusing errors discussed in #933 - see PR #934 that will receive an upstream fix with this PR after merge.
* when `devon «tool» version list` is performed or `devon «tool» version set latest` is performed, the mirrors git needs to be cloned and up-to-date what was missing in #826 
* after `devon «tool» version set «version»` I now output a hint to the user to run setup to actually do the installation

### DIFF
--- a/documentation/functions.asciidoc
+++ b/documentation/functions.asciidoc
@@ -187,7 +187,7 @@ If the URL is not provided as first argument, it will use the `mirrors` config t
 
 |=======================
 |*Param*|*Name*  |*Description*
-|`$1`   |URL     |The explicit URL to download from or `-` to compute in xref:doDownload[].
+|`$1`   |URL     |The explicit URL to download from or `-` to compute from mirrors.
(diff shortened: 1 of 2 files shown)


hohwille:
How come that the build of this PR is green?
The function `doUpdateMirrors` will be added in PR #934 and is missing here.
Therefore `test-function-doc` should fail.
Shouldn't shellcheck also notice this and break the build?

hohwille:
> Documentation of functions and function declarations are in sync.
Awesome!
[SUCCESS] Succeeded running test #0: test-functions-doc

OK. It has been vice versa:
`doUpdateMirrors` was called but not declared. As it was not declared the test did not fail as no declared function was missing in the documentation.

The question is if we can detect if some `do[A-Z].*` is called but never declared.

Issue #944, opened 2022-10-13
Remove cancelled runs from all workflows


alfeilex:
Cancelled runs prevent to show the regular status of integration-test. The badges in the README.asciidoc use the status of the last runs. Unfortunately, canceled runs are taken into account.

This problem could be fixed by removing all cancelled runs from workflows

PR #945, opened 2022-10-13
#944: Remove cancelled runs from all workflows

alfeilex:
To delete all cancelled workflows, I use [this](https://github.com/Mattraks/delete-workflow-runs) github action in d580866 . The question is: Can we trust this github action or should we implement this action ourselves?

### DIFF
--- a/.github/workflows/clean-runs.yml
+++ b/.github/workflows/clean-runs.yml
@@ -1,0 +1,17 @@
+name: Delete cancelled workflow runs
+on:
+  schedule:
+    - cron: '0 6 * * *'
+
+jobs:
+  del_runs:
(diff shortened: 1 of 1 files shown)


hohwille (approved):
@alfeilex great that you found this solution. 👍 
So cancelled workflow runs will be deleted automatically via cron once per day.
For me it is OK to reuse existing workflows on github.
Surely Mattraks could potentially change the workflow in `v2` and send out `github.token` via internet to some evil usage. See [here](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#stealing-the-jobs-github_token) for details.
However, we are already using some other actions from external github users and Mattraks does not look suspicious to me.

Issue #901, opened 2022-09-26
Remove CI/CD generator (cicdgen)


baumeister25:
The cicdgen command is based on the https://github.com/devonfw/cicdgen repository which is not maintained actively since more than 2 years and will be deprecated and archived soon. Therefore, the devon-ide should provide the functionality further via the CLI.

Please remove
- [ ] any usage of the repository https://github.com/devonfw/cicdgen 
- [ ] the dependent CLI command(s)
- [ ] documentation.

PR #951, opened 2022-10-17
#901: remove cicdgen

Amueller36:
Removed all occurences of CICDGEN to implement #901.


hohwille:
Great work 👍 Merging now...

PR #952, opened 2022-10-17
#903: eclipse plugin removal

Akuhana:
Implemented a way to remove plugins


hohwille (changes requested):
Nice work. Thanks for the PR and finding this feature in Eclipse API. 👍 
I have the same issues as for #911:
* `eclipsec` only exists on windows so we need to use `${ECLIPSE}` instead of `${ECLIPSE_HOME}/eclipsec`
* Instead of writing the entire log and then filtering it with `sed`, we should directly use a pipe for that as already done in line 67 (IMHO we should even create a function for the invocation of eclipse with the `-application` option, etc. including to write a log file and grepping for an expected message to check if this was successful or not instead of copy&paste)
* If the `plugin_id` property is undefined nothing happens to actually uninstall the plugin. Shouldn't we raise an error instead?
* To install a plugin one can even provide a `plugin_id` followed by Update-Site-URL. If I later want to uninstall this, it would fail. Shouldn't we try to use the given parameter of the plugin to uninstall as `plugin_id` directly if no according plugin properties file exists?
* IMHO we should first uninstall the plugin via eclipse and then remove the `.devon.*.installed` file. This would make the process adding/removing plugins symmetric and the more fragile part is the eclipse invocation and not deleting an empty file.

For the record: I will either catch up with this PR or someone from the new team can pick this to do the rework as @Akuhana is not available anymore for devonfw-ide.

Amueller36:
Only thing I havent done yet is 
"Instead of writing the entire log and then filtering it with sed, we should directly use a pipe for that as already done in line 67 (IMHO we should even create a function for the invocation of eclipse with the -application option, etc. including to write a log file and grepping for an expected message to check if this was successful or not instead of copy&paste)" 
this because I am not sure how to do it

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -108,110 +115,110 @@ function doCheckPluginsInstalled() {
...
+    return 1
+  else
+    doEcho "Removing plugin ${1}..."
+
+    local debug_info_file
+    debug_info_file=$(mktemp)
+    "${ECLIPSE}" -nosplash -clean -consoleLog -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.equinox.p2.director -uninstallIU "${plugin_id}" -vmargs "${ECLIPSE_VMARGS}" >"${debug_info_file}" 2>&1
+    local exit_code=$?
+    # Overwrite file with lines containing !MESSAGE
+    sed -i -n '/!MESSAGE/ p' "${debug_info_file}"
```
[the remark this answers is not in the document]
hohwille: You already wrote this:
> Only thing I havent done yet is
"Instead of writing the entire log and then filtering it with sed, we should directly use a pipe for that as already done in line 67 (IMHO we should even create a function for the invocation of eclipse with the -application option, etc. including to write a log file and grepping for an expected message to check if this was successful or not instead of copy&paste)"
this because I am not sure how to do it

hohwille (changes requested):
@Amueller36 thanks for the rework. You have done a very good job 👍 
Maybe we can have a short call to discuss the feedback.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -1,14 +1,11 @@
 #!/usr/bin/env bash
 
 # autocompletion list
-if [ "${1}" = "shortlist" ]
-then
+if [ "${1}" = "shortlist" ]; then
```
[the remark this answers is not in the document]
hohwille: You reverted these changes. Thanks.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -39,45 +33,58 @@ source "$(dirname "${0}")"/../commandlet-cli
...
   local plugin_dependencies
   plugin_dependencies=$(doCheckPluginsInstalled "${1}")
 
   # Install plugins and write debug info into temp file
   local debug_info_file
   debug_info_file=$(mktemp)
-  # get only !MESSAGE lines from the debug info
-  "${ECLIPSE}" -consoleLog -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.equinox.p2.director -repository "${2}" -installIU "${1}" -vmargs "${ECLIPSE_VMARGS}"  | grep '^!MESSAGE' | sed 's/^!MESSAGE //' > "${debug_info_file}" 2>&1
+  # get only !MESSAGE lines from the debug inf
+
```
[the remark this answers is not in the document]
hohwille: I reverted this myself.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -165,47 +227,23 @@ function doSetup() {
...
+  local optional_Plugin_Active="${2}"
+  if [ -f "${plugin_properties_file}" ]
   then
-    doLoadPluginProperties "${1}"
-    local url
-    if [ -n "${plugin_mirror_url}" ]
+    doLoadPluginProperties "${plugin_properties_file}"
+    if [ -n "${plugin_id}" ] && [ -n "$plugin_url" ] && [ "$plugin_active" = "true" ] || [ "${optional_Plugin_Active}" = "true" ] 
     then
-      url="${plugin_mirror_url}"
```
hohwille, line 174: you broke the `plugin_mirror_url` feature with your refactoring.

hohwille (dismissed):
@Amueller36 Thanks for your nice improvements. 👍 This PR is getting a little bit big and tricky to complete. I hope we will be able to complete this without further review rounds.
Can you follow DoD and add the two tickets to the CHANGELOG as well as address my review comments?

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -39,47 +39,65 @@ source "$(dirname "${0}")"/../commandlet-cli
...
+  set -o pipefail
+  "${ECLIPSE}" -consoleLog -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.equinox.p2.director -repository "${pluginUrl}" -installIU "${pluginId}" -vmargs "${ECLIPSE_VMARGS}" | grep '^!MESSAGE' | sed 's/^!MESSAGE //' >"${debug_info_file}" 2>&1
   local exit_code=$?
   if [[ $(grep "Overall install request is satisfiable" "${debug_info_file}") != "" ]] && [ "${exit_code}" == 0 ]
   then
+    #If plugin name was specified, create a properties file for an easier remove plugin process.
+    if [ -n "${3}" ]
+    then
+      doCreateProperties "${pluginId}" "${pluginUrl}" "${pluginName}"
+    fi
```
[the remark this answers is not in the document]
Amueller36: I removed the overwriting of the properties file now, I didnt know about the mirror url feature before which was my bad. I readded the code for the plugin_mirror_url feature!

documentation/eclipse.asciidoc
```diff
@@ -16,6 +16,7 @@ You may also supply additional arguments as `devon eclipse «args»`. These are
 |`--all`                  |if provided as first arg then to command will be invoked for each workspace
 |`setup`                  |setup Eclipse (install or update)
 |`add-plugin «id» [«url»]`|install an additional plugin
+|`remove-plugin «id»`     |remove an installed plugin with given id
```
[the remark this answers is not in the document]
Amueller36: Updated the docu now

scripts/src/main/resources/scripts/command/eclipse
Amueller36: Readded the code for the plugin_mirror_url feature.

Amueller36:
By the way, I tried to connect the other related to issue to this ticket but it didn't work.
image

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -363,10 +397,10 @@ function doLoadPluginProperties() {
   then
     file="${SETTINGS_PATH}/eclipse/plugins/${1}.properties"
   fi
-  if [ -f  "${file}" ]
+  if [ -f "${file}" ] 
   then
     plugin_id=""
-    plugin_active="true"
+    plugin_active=""
```
hohwille, line 403: It may be a good default to leave this undefined. However, with this change we are changing existing behavior. Do we really want to do this?

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -430,18 +464,20 @@ do
   elif [ "${1}" = "add-plugin" ]
   then
     shift
-    doSetup silent
```
hohwille, line 433: if you remove this and eclipse is not installed, then `doAddPlugin` will fail with `command not found: eclipse`.
IMHO this should not be removed.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -430,18 +464,20 @@ do
...
     exit ${?}
-  elif [ "${1}" = "import" ] 
+  elif [ "${1}" = "remove-plugin" ]
+  then
+    shift
+    doRemovePlugin "${@}"
+    exit ${?}
+  elif [ "${1}" = "import" ]
   then
-    doSetup silent
```
hohwille, line 438: if you remove this here, you should call it from the function calling the `${ECLIPSE} ...` command.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -39,47 +39,67 @@ fi
...
     return
   fi
+  echo -e "plugin_url=${pluginUrl}\nplugin_id=${pluginId}\nplugin_active=${pluginStatus}" >>"${propertiesFile}"
+}
+
+
+function doInstallEclipsePlugin() {
+  local doinstall_Plugin_Id="${1}"
+  local doinstall_Plugin_Url="${2}" #Update site
+  local doinstall_Plugin_Name="${3}" # if specified, the plugin will be installed and a properties file will be created with the name specified
```
hohwille, line 70: This comment is IMHO confusing as this function does not create such properties file. 
Feel free to move this hint to `doAddPlugin` as this is where the properties file gets created if `$3` is given.
BTW: why not just naming the variables short like `plugin_id` instead of `doinstall_Plugin_Id`? Also typically either do camlCase or snake_case but do not mix both...

hohwille (approved):
@Amueller36 thanks for your rework. This PR is now functionally ready 👍 
I have left some nice-to-have review comments, but generally this could be merged now.

scripts/src/main/resources/scripts/command/eclipse
hohwille: This is also a documented feature:
https://github.com/devonfw/ide/blob/master/documentation/eclipse.asciidoc#plugins

So changing this and leaving the existing documentation untouched will cause an inconsistency.
I recommend that you revert this change.

Amueller36: I don't understand, why this changes existing behaviour, the plugin_active variable gets overridden 3 lines later anyway. (doLoadProperties "${file}")
so if the plugin was true in the settings folder(which should be the default case), then the plugin_active variable will be true aswell.
Nevertheless I changed it to true in the following commit.

scripts/src/main/resources/scripts/command/eclipse
Amueller36: readded doSetup silent

scripts/src/main/resources/scripts/command/eclipse
Amueller36: readded doSetup silent

scripts/src/main/resources/scripts/command/eclipse
Amueller36: The reason why I chose function specific names is that plugin_id is a global variabel if I am not wrong. So I did not want to cause any undesired behaviour by messing with the global variable.
Should I refactor all variables to snake_case then, because the other functions that were written are already in snake_case

hohwille: > The reason why I chose function specific names is that plugin_id is a global variabel if I am not wrong. So I did not want to cause any undesired behaviour by messing with the global variable.

The local variable would not influence the global one but it could indeed cause confusion naming both identical.

scripts/src/main/resources/scripts/command/eclipse
hohwille: > I don't understand, why this changes existing behaviour, the plugin_active variable gets overridden 3 lines later anyway. (doLoadProperties "${file}")

That only happens if the *.properties File contains a line saying something about `plugin_active`. Otherwise the defaults defined here in the code apply.

Issue #875, opened 2022-09-09
Extend integration-tests with advanced features


hohwille:
Analog to #868 we want to extend integration-tests with additional testing of specific features:
* testing of [custom software repo](https://github.com/devonfw/ide/blob/master/documentation/software.asciidoc#custom)
* testing of [shared tools](https://github.com/devonfw/ide/blob/master/documentation/software.asciidoc#shared) (on linux/mac)
* test of different versions (e.g. java8, java11, java19) and combinations?

hohwille:
Actually all was already written down by me in the epic #663. However, almost nobody noticed and read this. Now stories had been created without linking the epic. Pitty. I assume we can close the epic at least when this story is cloded.

PR #962, opened 2022-10-25
#875:  Extend integration-tests with advanced features

alfeilex:
Nightly tests have been extended:

- add custom tools test
- add shared tools test
- add extra java version test
- add a summary of all errors in the end of the workflow
- change all-test to avoid redundancy of code


scripts/src/test/bash/integration-test-custom-tools
```diff
@@ -0,0 +1,26 @@
...
+"${CLI}" ide update
+
+if [[ -f "${TOOL_PATH_A}" && -f "${TOOL_PATH_B}" ]]
+then
+  rm "${PWD}/settings/devon.properties"
+  exit 0
+else
+  rm "${PWD}/settings/devon.properties"
+  exit 1
+fi
```
hohwille, line 26: can be simplified to:
```
rm "${PWD}/settings/devon.properties"
[[ -f "${TOOL_PATH_A}" && -f "${TOOL_PATH_B}" ]] || exit 1
```
or
```
rm "${PWD}/settings/devon.properties"
[[ -f "${TOOL_PATH_A}" && -f "${TOOL_PATH_B}" ]]
exit ${?}
```

scripts/src/test/bash/integration-test-java-extra
```diff
@@ -0,0 +1,17 @@
...
+"${CLI}" java setup
+
+if [ -x "${PWD}/software/extra/java/bin/java" ]
+then
+  rm "${PWD}/conf/devon.properties"
+  exit 0
+else
+  rm "${PWD}/conf/devon.properties"
+  exit 1
+fi
```
hohwille, line 17: same here.
If we always want to remove `devon.properties` to avoid side-effects between tests, shouldn't we just do this in all-tests?

For the record: `all-tests` is always creating a new folder for each test-case and creates a `conf/devon.properties` in there. It seems that all command-tests and all integration-tests always work in `test-setup`.
If that is our default, we could simply change this and create a central folder in `all-tests` where all tests are executed and that is properly initialized.
Then we could simplify all other tests that first `cd` to `test-setup`, deleting software (`rm -fR "${PWD}/software"`) that should be done in `all-tests`, etc.
Another thing is that most of these tests therefore "depend" on `test-setup` being run before.
However, this is not guaranteed for `test-*` as it depends on the directory order of the tests.

scripts/src/test/bash/integration-test-shared-tools
```diff
@@ -0,0 +1,18 @@
...
+"${CLI}" java setup
+
+if [ -d "${SHARED_DIR}/default/java" ]
+then
+  rm "${PWD}/conf/devon.properties"
+  exit 0
+else
+  rm "${PWD}/conf/devon.properties"
+  exit 1
+fi
```
hohwille, line 18: once again.

scripts/src/test/bash/integration-test-java-extra
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+cd ../test-setup
+CLI="${PWD}/scripts/devon"
+EXTRA_VERSION=$(head -n 1 "${PWD}"/mirrors/java/available-versions)
```
hohwille, line 5: this would be the same as `JAVA_VERSION` that is undefined and would also lead to the same latest version.
In that case, can you also set `JAVA_VERSION` to some 11x version - why not `11*` to test the new feature of version prefix?

hohwille (changes requested):
@alfeilex excellent. Great work 👍 
Can you have a look at my review-comment(s)?

scripts/src/test/bash/integration-test-java-extra
alfeilex: I added this code to avoid redundancy with `cd test-setup` and `rm -fR software` and so on. In my view it is a stable change without changing the whole test structure. However, we can discuss whether to completely redesign the workflow of `all-tests`.

https://github.com/devonfw/ide/blob/bb49b7e547355351df8befe1ac151af36eea8e09/scripts/src/test/bash/all-tests#L14-L19

alfeilex:
I have tested the new extended test on Windows. Next, I will test on MacOS and Linux.

hohwille (approved):
@alfeilex thanks for your rework. Everything looks just perfect 👍 
as PR #932 is now merged, we can now already use the `11*` feature.
To avoid (unlikely) globbing, I did put the `JAVA_VERSION` in single quotes.
Ready to be merged.

scripts/src/test/bash/integration-test-java-extra
hohwille: Excellent solution. I am resolving this conversation.

scripts/src/test/bash/integration-test-java-extra
```diff
@@ -0,0 +1,14 @@
...
+JAVA_VERSION='11*'
+EXTRA_VERSION=$(head -n 1 "${PWD}"/mirrors/java/available-versions)
+
+echo -e "\nJAVA_VERSION=${JAVA_VERSION}" >> "${PWD}/conf/devon.properties"
+echo -e "\nEXTRA_JAVA_VERSION=${EXTRA_VERSION}" >> "${PWD}/conf/devon.properties"
+"${CLI}" java setup
+JAVA_TEST=$(cat "${PWD}"/software/java/.devon.software.version)
+EXTRA_TEST=$(cat "${PWD}"/software/extra/java/.devon.software.version)
+
+[[ $JAVA_TEST == $JAVA_VERSION && $EXTRA_TEST == $EXTRA_VERSION ]]
```
hohwille, line 13: I just stumbled over this one:
`$JAVA_TEST == $JAVA_VERSION` will not match as `JAVA_VERSION` will be `11*` but `JAVA_TEST` will be something like `11.0.17_8`.

hohwille: I just fixed it.

hohwille: OK, I was actually wrong here, as I just learned:
```
JAVA_TEST=11.0.17_8
if [[ ${JAVA_TEST} = 11* ]]; then echo match; fi
```
will print "match" :) Sorry, in case of any confusion...

Issue #966, opened 2022-11-03
npm detection not reliable and redundant


hohwille:
We have the commandlets that call `npm` and have redundant logic to detect it that also seems to be buggy.
The commandlets based on `npm` are:
* `ionic`
* `ng`
* `yarn`

Instead of finding `npm` inside these commandlets to call `doRunCommand "npm install ..."` we should simply delegate to `doDevonCommand npm install ...`.

Copied from PR #934:
During the setup of npm a 'MODULE_NOT_FOUND' blocks the  installation of ionic. 
To be more precise, I've been getting the following error after the installation of node seemed
to be complete:

`node:internal/modules/cjs/loader:998
  throw err;
  ^

Error: Cannot find module '/home/dr/projects/PR934/software/node/bin/node_modules/npm/bin/npm-cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
    at Module._load (node:internal/modules/cjs/loader:841:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:82:12)
    at node:internal/main/run_main_module:23:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v19.0.0
Could not determine Node.js install directory

******** ATTENTION ********
Failed to install ionic-cli ('/home/dr/projects/PR934/software/node/npm')
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
Exit code was 1
`

Full log of this error can be found here: https://github.com/alfeilex/ide/actions/runs/3368418467/jobs/5586938443#step:4:11593

In my case the npm-cli.js file has the following path
```
/home/dr/projects/PR934/software/node/node_modules/npm/bin/npm-cli.js
```
instead of
```
/home/dr/projects/PR934/software/node/bin/node_modules/npm/bin/npm-cli.js
```

It seems that the errors occurs because the npm_command uses the path from here:
https://github.com/devonfw/ide/blob/d9ebcad8bdc0c43bb7a205580b3212c2f9c6ebdd/scripts/src/main/resources/scripts/command/ionic#L28
but it must be the path from here:
https://github.com/devonfw/ide/blob/d9ebcad8bdc0c43bb7a205580b3212c2f9c6ebdd/scripts/src/main/resources/scripts/command/ionic#L31

PR #967, opened 2022-11-07
#966 npm detection not reliable and redundant

Amueller36:
I updated the ionic and npm commands,
the if check that was done before is redundant, because the npm_command path will always be at software/node/npm Tested ionic command and it now works fine

### DIFF
--- a/scripts/src/test/bash/all-tests
+++ b/scripts/src/test/bash/all-tests
@@ -11,9 +11,17 @@ for testpath in "$(dirname $0)"/"${1}"-*
 do
   testcase="${testpath/*\//}"
   echo "Running test #${total}: ${testcase}"
-  mkdir -p "${testcase}/conf"
-  cd "${testcase}"
-  echo "export M2_REPO=~/.m2/repository" > "conf/devon.properties"
+  if [[ $testcase == integration* ]]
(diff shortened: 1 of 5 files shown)


scripts/src/main/resources/scripts/command/ionic
```diff
@@ -14,6 +14,8 @@ if [ -n "${DEVON_IDE_TRACE}" ]; then set -vx; fi
 TOOL_VERSION_COMMAND="ionic -v"
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
+# shellcheck source=scripts/command/npm
+source "$(dirname "${0}")"/npm
```
hohwille, line 18: IMHO you can not source the script as this will execute its code as if it was pasted here what will cause undesired side-effects from the `npm` commandlet into the `ionic` commandlet.

hohwille (changes requested):
@Amueller36 thanks for creating this PR so fast.
However, I noticed that I forgot to copy the implementation suggestion that is hidden in the original PR https://github.com/devonfw/ide/pull/934#discussion_r1013197215 into the issue #966 - sorry, I put it into the review comments.

Issue #954, opened 2022-10-18
First install removes all folders from user path


RichardLinde:
### Expected behavior

When first installing the ide I want to use it to add other programs/tools to my already existing programs.
For example, I already had VSCode installed and wanted to use the ide for the first time to install Rancher Desktop.

### Actual behavior
When running the `start.bat` for the first time while `%USERPROFILE%\scripts` is not added to the user path the installation will instead of adding it to the path replace the path with it. Removing all folders from the path.
In my example the command `code` was removed and had to be added again.

### Steps to reproduce 

1. check content of user path variable (Add a path if there are none)
2. clean system from devon ide
2.1 Uninstall devon ide `devon uninstall ide`
2.2 Remove `%USERPROFILE%\scripts` from user path 
3. Install devon ide again by running `start.bat`
4. Check if the user Path only contains %USERPROFILE%\scripts

### Comments/Hints:
Already talked about the issue with @cinnamon-coder-hub 

### Affected version:
Version: 2022.08.003
OS: Windows 10.0.19044 Build 19044

hohwille:
This is a duplicate of #409.
As you can see on the milestone the issue is already fixed and released.
Simply do an update (download latest version or run `devon ide update scripts`).
With the new version of devonfw-ide it should be fixed.
Otherwise please give details and reopen.

RichardLinde:
![image](https://user-images.githubusercontent.com/112941483/197754020-e93f23d0-1b86-4bb5-a179-f942459a32a6.png)


I updated everything and tested again. The `setup.bat` still removed all folders from the user Path.
- removed the old  devonfw-ide-scripts folder
- removed the folder `%USERPROFILE%\scripts`
- removed the folder `%USERPROFILE%\scripts` from path
- downloaded the latest version 2022.08.003
- unpacked the folder to C:\_dev\tutorials\devonfw-ide-scripts-2022.08.003
- ran the `setup.bat` by clicking it in the explorer directly
- log out / log in
- Path variable only contains `%USERPROFILE%\scripts`



OUTPUT LOGS
```
Setting up your devonfw-ide in C:\_dev\tutorials\devonfw-ide-scripts-2022.08.003
Copying devon CLI script to your home directory...
        1 file(s) copied.
Adding C:\Users\rlinde\scripts to your users system PATH

SUCCESS: Specified value was saved.
←[93m"ATTENTION:"←[0m
"Your user PATH environment variable has not been previously set."
[...]

PR #975, opened 2022-11-11
#954: Fixed devon.bat with delay expansion

alfeilex:
Added delay expansion in `devon.bat` to fix a bug reported by a user. Therefore some variables were changed from `%VAR%` to `!VAR!`. 

Issue: https://github.com/devonfw/ide/issues/954

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -6,7 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/ide[devonf
 
 WIP: Release with small but important bugfixes:
 
-* TODO: replace me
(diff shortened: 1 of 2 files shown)


hohwille (approved):
@alfeilex thanks for the fix. Looks good to me 👍 
We should wait for the user reporting #954 to test and verify this fix.
I was first a little frightened about the mixture of `%VAR%` and `!VAR!` syntax but it seems to be correct.
As you also have tested this it should be fine to merge after the user confirmed the fix.

alfeilex:
> Any feedback yet? Can you follow DoD and add CHANGELOG entry?

DoD and CHANGELOG are done. No feedback from the user at the moment.

Issue #974, opened 2022-11-10
Fix dotnet command for nightly test


alfeilex:
### Actual behavior
The nightly tests in all OS's detect an error with `dotnet` command. 

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Check latest nightly tests for [Windows](https://github.com/devonfw/ide/actions/runs/3415590306/jobs/5684879737#step:5:883), [MacOS](https://github.com/devonfw/ide/actions/runs/3415908475/jobs/5685524287#step:5:11215) and [Ubuntu](https://github.com/devonfw/ide/actions/runs/3416122554/jobs/5685946063#step:4:10658) 

### Comments/Hints:
I have done the tests locally and no such an error occurs. A keyword of the error is `Segmentation fault (core dumped)`.

mustafaokyay:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IDE-61dbfc20045ead0022812582/issues/devonfw/ide/974?planning-poker) with Zenhub @alfeilex @hohwille @tobka777 @Amueller36

mustafaokyay:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IDE-61dbfc20045ead0022812582/issues/devonfw/ide/974?planning-poker) with Zenhub @CREITZ25 @cinnamon-coder-hub

mustafaokyay:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IDE-61dbfc20045ead0022812582/issues/devonfw/ide/974?planning-poker) with Zenhub @Ttring

alfeilex:
It seems that this line https://github.com/devonfw/ide/blob/249f5f680c86ca778bc5b97a21c23f93bffc8afa/scripts/src/main/resources/scripts/command/dotnet#L35  leads to an error in github actions runners. At the moment, I can't figure out why another installation of dotnet is started after the installation of dotnet. And why git commands are not found? 

https://github.com/devonfw/ide/actions/runs/3475607093/jobs/5810015947#step:5:361

alfeilex:
I suppose that the github hosts don't the typical directory structure. 

https://github.com/devonfw/ide/blob/249f5f680c86ca778bc5b97a21c23f93bffc8afa/scripts/src/main/resources/scripts/command/dotnet#L35
[...]

PR #980, opened 2022-11-16
#974: Fixed dotnet command for nightly test 

alfeilex:
The nightly test showed us an error after the dotnet installation. While analysing the error, I realised that there is something like a loop so that the github runner comes to `segmentation fault` error.  This PR fixes this bug.

More details can be found in related issue: https://github.com/devonfw/ide/issues/974

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -6,7 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/ide[devonf
 
 WIP: Release with small but important bugfixes:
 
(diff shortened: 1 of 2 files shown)


hohwille (approved):
@alfeilex Thanks for the fix 👍 
Might be worth to add to CHANGELOG but ready for merge...

alfeilex:
> @alfeilex Thanks for the fix 👍 Might be worth to add to CHANGELOG but ready for merge...

Wasn't sure if it belonged in the changelog. Can be merged

Issue #982, opened 2022-11-16
Add new feature to vscode - recommended plugin installation


alfeilex:
In the past, when installing vscode, the devonfw-extentions-pack was automatically installed. This plugin installs ~60 additional extensions and takes some time to complete the installation of all of extensions. This causes errors for the nightly tests and perhaps, more importantly, not every user wants to have this plugin installed by default. So the idea is to add an option in the vscode command to install vscode without devofw-exetnsions-pack. 

The new option should be to recommend plugins to the user after opening it. One possible way would be to add an entry like `Recommend=true` in the properties file. [Here](https://dev.to/askrishnapravin/recommend-vs-code-extensions-to-your-future-teammates-4gkb) is the description of how to implement the whole thing for vscode.

**However, the default should be that the plugin is installed automatically, the recommendation is just an option.**

- [x] DoD
- [x] Adding the option to recommend plugins to the user
- [x] Adjusting the vscode documentation

alfeilex:
While implementing this feature, I have the following idea. When vscode is set up the `doAddPlugins` function https://github.com/devonfw/ide/blob/689d2df14041f2207a4d7af4ef266d3e6228c788/scripts/src/main/resources/scripts/command/vscode#L48 will install all plugins configured in the ide settings. Maybe it makes more sense to add deactivated plugins automatically as "recommended". As a result we don't need an additional property. 

Furthermore, we could add a new option `devon vscode clean-setup` which installs vscode without any plugins and recommendations. This feature can be used by the user and we can use this for nightly testing.

The function is already implemented in PR https://github.com/devonfw/ide/pull/990. I need feedback to add this feature to the documenation and complete the PR.

alfeilex:
The documentation has been updated, but we still need to dicuss how detailed we need to explain this feature in the docs.

PR #993, opened 2022-11-24
#982: Add plugin recommendation feature to vscode

alfeilex:
This feature adds deactivated plugins as "recommended" plugins for vscode, after vscode is installed. See [here](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions).

Link to related issue: https://github.com/devonfw/ide/issues/982


alfeilex:
Please check the comments and my train of thought in the related issue: https://github.com/devonfw/ide/issues/982

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -126,6 +138,18 @@ function doStartVsCode() {
   doRunVsCode "${WORKSPACE_PATH}"
 }
 
+function doAddRecommendations() {
+  doEcho "add recommendations..."
+  local extensions="${WORKSPACE_PATH}/.vscode/extensions.json"
+  if [ -f "${extensions}" ]
+  then
+    rm "${extensions}"
```
hohwille, line 146: IMHO this is a dangerous behavior and most users will very much dislike it.
What if some plugins are already installed (active=true, or manually installed) and then (maybe weeks later) a recommended plugin shall be added? So you are simply kicking out all plugins of the user?
That seems odd or maybe I do not understand this approach and am missing something...

hohwille (changes requested):
@alfeilex thanks for your idea with the recommendation feature and creating this PR. 👍 
While I like the idea very much, I do not yet agree that this PR's implementation is ready for merge. Can you have a look? Feel free to get in touch so we can discuss how to address this finally.

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -180,6 +205,11 @@ do
     shift
     doAddPlugin "${@}"
     exit ${?}
+  elif [ "${1}" = "clean-setup" ]
+  then
+    doCleanSetup "${2}"
+    doStartVsCode
```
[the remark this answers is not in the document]
alfeilex: I added this feature while I was doing the integration-test for `vscode`. I had to do a workaround there, because the `devonfw-extension-package` broke the whole workflow. So I thought it would be nice to have an option to install vscode without plugins in "clean" way.

I am not sure, but probably user with poor internet connection will also have the problem that the plugin takes a lot of time to install. There is already an issue for the extension pack https://github.com/devonfw/extension-pack-vscode/issues/12. If there will be an update, the clean installation is not so relevant anymore or we disable the extension plugin as default.

scripts/src/main/resources/scripts/command/vscode
alfeilex: https://github.com/devonfw/ide/blob/3e4e62f7b1dbf6c079415d279d769c213f43e680/scripts/src/main/resources/scripts/command/vscode#L52-L55

The `doAddRecommendation()` function is only executed during the execution of `vscode setup`. So the `extensions.json` can not be deleted anyway. So we don't need this query there either. We can assume that the `extensions.json` does not exist. 

Furthermore, the `extensions.json` for installed plugins is in `software/vscode-extensions`. Unfortunately, I can't use this file because adding the `recommendations` key there has no effect on vscode. The recommendations have to be added to `${WORKSPACE_PATH}/.vscode/extensions.json`

scripts/src/main/resources/scripts/command/vscode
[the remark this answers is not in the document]
hohwille: > Maybe it is better to add an option like -no-plugins instead, if we need this at all.

If you still think it would be good to offer a way to skip plugin installation, that was my recommendation.
Otherwise we can entirely remove the "clean-setup".

I am also happy to have the devonfw extension pack as recommendation by default instead of a fully automated installation for every user by force as https://github.com/devonfw/extension-pack-vscode/issues/12 does not seem to make any progress.
So once we have this PR completed and merged, my understanding is that VSCode plugins with `plugin_active=false` will not be installed but added as "recommendation". So after merge, we can then change our settings accordingly:
https://github.com/devonfw/ide-settings/blob/d98a888be2e02922cc5e81f7c186e9c8db40f89f/vscode/plugins/devonfw-extension-pack.properties

scripts/src/main/resources/scripts/command/vscode
hohwille: Oh, that sounds like bad news and this IMHO means that this feature is not really working as it is.
Further, implementing it properly will be tricky as new workspaces can be added at any time.

> Unfortunately, I can't use this file because adding the recommendations key there has no effect on vscode.

See
https://github.com/devonfw/ide/blob/c53d140ccc881b869a9e082a599f25a9379e7320/scripts/src/main/resources/scripts/command/vscode#L107
and
https://github.com/devonfw/ide/blob/c53d140ccc881b869a9e082a599f25a9379e7320/scripts/src/main/resources/scripts/command/vscode#L120

I am confused that `extensions.json` are read from `user-data-dir` rather than `extensions-dir`. Bug in VSCode?

alfeilex: I have implemented a new solution to add the recommendations. Now after each run of `devon vscode` the plugin recommendation is added to the current workspace. If there are is already an entry with the same values, nothings happens. But when the user adds  new plugins in to the properties file, the `extensions.json` is updated with the new plugins set to `false`.

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -122,10 +122,55 @@ function doRunVsCode() {
...
 }
 
+function doAddRecmdt() {
+  local recmdtJSON="${WORKSPACE_PATH}/.vscode/extensions.json"
+  local file
+  local recmdtKey
+  local recmdtValue
+  local pluginsToRecmdt=()
+  for file in "${SETTINGS_PATH}"/vscode/plugins/*.properties
+  do
```
alfeilex, line 137: This code block could be moved in to a function, as it is already used in the `doAddPlugins` function

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -122,10 +122,55 @@ function doRunVsCode() {
...
+  done
+  recmdtValue=${pluginsToRecmdt[*]}
+  recmdtKey="\"recommendations\": [$recmdtValue]"
+  if [ -f "${recmdtJSON}" ] && [ ! -s "${recmdtJSON}" ]
+  then 
+    if grep -q "\"recommendations\":" "${recmdtJSON}"
+    then
+      if ! grep -Fq "${recmdtKey}" "${recmdtJSON}"
+      then
+        sed -i "s/\"recommendations\": \[[^]]*\]/${recmdtKey}/g" "${recmdtJSON}"
```
alfeilex, line 161: I had some difficulties adding entries in a JSON property. There is a tool `jq`, but since this is not a default tool, I chose some `sed` commands. If there are better solutions, I am happy to hear that.

hohwille: Correct. I also used `jq` once in a different context but as you already wrote we can not expect it to be installed.
We would need to add a commandlet `jq` that download and installs it in order to use it here. Seems overkill as long as you can solve it with `sed` hacking :)

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -122,10 +122,55 @@ function doRunVsCode() {
 
 function doStartVsCode() {
   doConfigureVsCode -u
+  doAddRecmdt
```
hohwille, line 125: why are we calling this on every run rather than in `doSetup` analogue to the plugins?

scripts/src/main/resources/scripts/command/vscode
hohwille: Sounds good. Maybe you can even directly integrate into `addPlugin` with a specific parameter...
For maintenance it is IMHO not so good to spare some single characters over readability (`doAddRecmdt` vs. `doAddPluginRecommendation`).
Bash and perl are designed with this philosophy of very short and compact syntax and the readability of such code is pain as you might have experienced ;)
After all our function names never need to be typed by end-users but only by our internal developers and we can use IDEs with auto-completion :)

scripts/src/main/resources/scripts/command/vscode
```diff
@@ -126,6 +137,32 @@ function doStartVsCode() {
...
+  local recommendations="${*}"
+  local extensionsJSON="${WORKSPACE_PATH}/.vscode/extensions.json"
+  local key="\"recommendations\": [$recommendations]"
+  if [ -f "${extensionsJSON}" ] && [ ! -s "${extensionsJSON}" ]
+  then 
+    if grep -q "\"recommendations\":" "${extensionsJSON}"
+    then
+      if ! grep -Fq "${key}" "${extensionsJSON}"
+      then
+        # if recommendations key is already in extensions.JSON
```
alfeilex, line 150: Since the IDE only adds the recommendation during installation, it may not be necessary to distinguish between all three cases here. But if, for example, a user reinstalls `vscode`, this could lead to errors without distinction. I would keep it.

hohwille: Thanks for explanation. Sounds reasonable. 👍

hohwille (approved):
@alfeilex thanks for your rework and completing this nice feature 👍 
Ready for merge now.

scripts/src/main/resources/scripts/command/vscode
hohwille: Seems fine now.

scripts/src/main/resources/scripts/command/vscode
hohwille: all done. Thanks.

scripts/src/main/resources/scripts/command/vscode
hohwille: Done.

Issue #941, opened 2022-10-11
New approach for ide-mirrors


hohwille:
This epic is about replacing [ide-mirrors](https://github.com/devonfw/ide-mirrors) with [ide-urls](https://github.com/devonfw/ide-urls). The latter has been created from scratch for this epic.
The rationale is that ide-mirrors had several design flaws in its structure that prevented us from automatic updating and required manual intervention. The idea of both mirrors and urls is to provide the meta-data of all the various tools we support with their available versions and URLs where to download which version for what operating system and architecture.
With the new approach of ide-urls we want to make a clean cut and redesign the entire structure of the metadata such that we get rid of manual maintenance and our end-users benefit from getting always the latest meta-data what especially means the latest available version information. The benefit of creating a new ide-urls repository is that we can keep the existing ide-mirrors untouched so existing releases of devonfw-ide out there that still rely on it do not get broken by our changes. Starting with release `2023.07.001` the new ide-urls will be used and ide-mirrors becomes legacy. For downward compatibility we will leave ide-mirrors online even for a long time after the new release is published. However, after a year or so it could also be deleted.

So far we have implemented an approach in [get-versions](https://github.com/devonfw/ide-mirrors/blob/master/get-versions) script that computes the `available-versions` from scratch every time via `curl` download and pipe magic to extract the versions from formats such as JSON, do specific replacements, filter unstable/undesired versions, sort versions, etc.

The actual details can be found in the comments below.

I marked this complex issue as an Epic and therefore created sub-tasks for the actual stories to break down the epic:
* #1007 
* #1008 
* #1009
* #1035
* #1084
* #1085
* #1086
* #1087
* #1088
* #1176
* #1142
* #1180
* #1181
* #1182
* Not required anymore: #1170 (the blocker part has been resolved, the rest can also be fixed later)
* #1200
* #1132
* #1210
* #1213
* #1218
* #1229

hohwille:
For the record we could even change more radically:
Create a new repo as replacement for `ide-mirrors` that is used by newer versions of `devonfw-ide` and has a new structure.
```
mirrors/«tool»/«version»/«os»[_«arch»].urls
```
So each version would be a folder containing config files. That would directly contain the complete download URL(s) for the exact version, OS, architecture. If no matching `«os»_«arch».urls` file is found, `«os».urls` is used, if again not found `urls` is used.
[...]

PR #994, opened 2022-11-27
#941:Added (folders + code) frame for discussion

cinnamon-coder-hub:
Related to #941. 
This PR is mainly supposed to give us a basic mutual folder structure to work on the issue #941 and further issues related to the new ide-mirror approach. Locally I was able to successfully execute the command `mvn compile`. Still, this may not be enough and some files may need to be added to complete the folder structure for propert usage with maven. 
As template I used the subfolders of the 'configurator' folder, but left out some files inside 'bin/src/test/resources/templates' as I'm not sure if they are needed. 
To discuss the generation and editing process of the urls-folders I included most of the related code I worked on recently. There are four classes that relate to the folder structures hierarchy, starting with the UrlRepository class, followed by UrlTool, UrlEdition and UrlVersion (for a better understanding I recommend reading Jörgs comments inside the related issue #941). A last class (I'm currently working on locally ) should be UrlFile. An object of this class will be related to an file, with download urls of a specific tool (with an specific edition), inside a version folder and there will be methods for reading, manipulating and finally saving the the files (potentially modified) content. 
To have a relation between the before mentioned hierarchies, a child-parent structure was introduced. Recently I tried to overcome some smaller unpleasant issues by writing the three subclasses UrlHasChildArtifact, UrlHasParentChildArtifact and UrlHasParentArtifact, as an result of talks with @hohwille. 
Those subclasses are inherited from UrlArtifact. Now I'll be happy to get some feedback and also welcome @Amueller36  to this project.


url-updater/bin/pom.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hohwille, line 1: This file was generated by Eclipse and should not be comitted.
You can add `bin` to `.gitignore` to avoid this in the future.

url-updater/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifest>
+              <addClasspath>true</addClasspath>
+              <mainClass>com.devonfw.tools.ide.url-updater.Url-updater</mainClass>
```
hohwille, line 40: A Java class shall never have dashes in its name. Java-packages can not have dashes (will cause compile error AFAIK - did this change in newer Java version?).

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlVersion.java
```diff
@@ -0,0 +1,43 @@
+package com.devonfw.tools.ide.url.folderhandling;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class UrlVersion extends UrlHasChildParentArtifact<UrlEdition, UrlFile> {
```
hohwille, line 7: UrlFile is missing. IMHO you forgot to commit this.
This code can not compile.

url-updater/pom.xml
cinnamon-coder-hub: In the last commit I now used camel case resulting in the folder name "urlUpdater" instead of "url-updater". Still, I would be open to other ideas or further proposals.

cinnamon-coder-hub: That's completely right. The mainClass was nonexistent anyway. Now I temporarily choose UrlRepository as main class, but maybe I should just leave this out for now, as we don't have a main class to start code execution yet?

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlVersion.java
cinnamon-coder-hub: This I left out by intention, but I guess this is an unpleasant way of making an early stage PR. So I added this class in an a kind of empty state for now. This week I should be able to commit this class with some proper methods.

url-updater/pom.xml
hohwille: In general better leave such things out until they are really needed. However, not preventing a merge to get collaboration started...

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlVersion.java
hohwille: Great. Yes, code should always compile without errors and JUnits if present should also pass.

hohwille:
Build failed with this error that seems unrelated to your PR:
```
curl: (22) The requested URL returned error: 503 
Failed to download https://github.com/adoptium/temurin19-binaries/releases/download/jdk-19%2B36/OpenJDK19U-jdk_x64_linux_hotspot_19_36.tar.gz with exit code 22

******** ATTENTION ********
Download of java in version 19_36 failed after trying 1 mirror(s).
```
It is great to have automated tests. However, in our case things are not as stable as desired and falsely breaking tests are a problem...

hohwille (approved):
@cinnamon-coder-hub thanks for the rework. 👍 
I will merge as a starting point for collaboration.
We still have to establish code-style rules like having JavaDoc and Junit-tests but fine to add this on the go.

hohwille:
This PR is obsolete and was replaced by #1000.

Issue #939, opened 2022-10-11
Consider extending test-functions-doc with invocation of undeclared functions


hohwille:
As a devonfw-ide PO, I want automated tests on the code so that bugs are detected early.
Developing bash is tedious. Where other programming languages like Java or C# would raise compile errors, we get successful build and tests even though the code is obviously broken.

As we have a strict convention to name all functions following the pattern `do[A-Z][a-zA-Z]*` we could easily extend our `test-functions-doc` with a check that greps all such patterns from functions and for each unique of them (`sort -u`) check if has been declared.
Otherwise raise and error and break the build.

So if `functions` contains e.g. `doUndefinedFunction` but not `function doUndefinedFunction()` then the test should report that `doUndefinedFunction` is undefined and fail.

mustafaokyay:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IDE-61dbfc20045ead0022812582/issues/devonfw/ide/939?planning-poker) with Zenhub @alfeilex @Amueller36 @cinnamon-coder-hub @CREITZ25 @hohwille @tobka777

PR #995, opened 2022-11-28
#939: extending test functions doc to check for invocation of undeclared functions

Amueller36:
Functions in commandlets are now checked to have a decleration either in the same file or in one of the devon function files!
Tested it and it already found one function that missed the "function" keyword. I didn't even know before that you could define functions without the function keyword 😄 

image
image

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -1,11 +1,12 @@
 = Changelog of devonfw-ide
 
 This file documents all notable changes to https://github.com/devonfw/ide[devonfw-ide].
-
+*
 == 2022.11.002
 
 Release with small but important bugfixes:
 
+* https://github.com/devonfw/ide/issues/939[#939]: Consider extending test-functions-doc with invocation of undeclared functions
 * https://github.com/devonfw/ide/issues/966[#966]: npm detection not reliable and redundant
 * https://github.com/devonfw/ide/issues/954[#954]: First install removes all folders from user path
 * https://github.com/devonfw/ide/issues/956[#956]: no matches found error if software folder missing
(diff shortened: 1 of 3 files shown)


CHANGELOG.asciidoc
```diff
@@ -14,6 +14,7 @@ The full list of changes for this release can be found in https://github.com/dev
 
 Release with small but important bugfixes:
 
+* https://github.com/devonfw/ide/issues/939[#939]: Consider extending test-functions-doc with invocation of undeclared functions
```
hohwille, line 17: 2022.11.002 is already released. Can you move this to 2022.12.001. Thanks.

scripts/src/test/bash/test-functions-doc
```diff
@@ -1,16 +1,38 @@
...
   echo "Documentation of functions does not match function declarations!"
   echo "Please keep function documentation and declarations in sync!"
   exit 1
 fi
+
+#For every file in SCRIPTS_PATH/command, check for functions that follow the pattern do[A-Z][a-zA-Z]*. 
+#Then check if there is a corresponding function declared in the same file or in functions.list file
+#which contains every function declared in functions and functions-core files.
+
+for file in "$SCRIPTS_PATH"/command/*; do
```
hohwille, line 28: great, we should also check the `functions` file itself as suggested in the issue.

scripts/src/test/bash/test-functions-doc
```diff
@@ -1,16 +1,38 @@
...
+  (
+    set -o pipefail
+    grep -o -E "do[A-Z][a-zA-Z]*" "$file" | sort | uniq |
+      while read -r line; do
```
hohwille, line 32: if I get it right, `line` is actually not the source-code line but only the called function name.
In that case renaming the variable to `functionName` (`function` seems to be a reserved keyword) or `funct` would make it more explicit.

hohwille (changes requested):
@Amueller36 thanks for this PR. excellent work. 👍 
I have very small review comments. Can you please check them and rework. After that I can merge this.

hohwille (approved):
@Amueller36 thanks for the rework. 👍 
Ready for merge now. Nice Work!

Issue #978, opened 2022-11-14
Support for MSI


hohwille:
In `doExtract` (in `functions`) we should add support for MSI files that can be extracted.
https://github.com/devonfw/ide/blob/605aa475d5a0103ccc22f20525ecdb3d6876094f/scripts/src/main/resources/scripts/functions#L673
Also remove `msi` from here:
https://github.com/devonfw/ide/blob/605aa475d5a0103ccc22f20525ecdb3d6876094f/scripts/src/main/resources/scripts/functions#L747

@CREITZ25 already found out how to extract MSI files so this could be done easily.

After that we can revisit `aws` commandlet and see if we can switch to a local installation by extracting MSI.
Also #874 can afterwards be revisited.

However, the scope of this issue is only to create a PR that extends `doExtract` so `*.msi` files are automatically extracted.
If that breaks any existing behavior (will happen for `aws` and maybe for for `docker`), simply add `noUnpack` option to keep previous behavior.
Then after being merged, we can change commandlets (and mirrors/urls config) to use MSI.

mustafaokyay:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IDE-61dbfc20045ead0022812582/issues/devonfw/ide/978?planning-poker) with Zenhub @alfeilex @Amueller36 @cinnamon-coder-hub @CREITZ25 @hohwille @tobka777

CREITZ25:
This issue will be implemented together with #964 and #992

CREITZ25:
How to see the configuration of a msi Archive:
https://learn.microsoft.com/en-us/windows/win32/msi/orca-exe

Issue #992, opened 2022-11-24
aws does not install a different version 


CREITZ25:
### Expected behavior

if aws release x is already installed I want to configure release y in devon.properties and start devon aws setup to install the release y instead of release x. 

### Actual behavior

if aws is already installed, the doInstall function won't start

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. install aws release 2.8.12 (devon aws setup)
2. configure devon.properties with AWS_VERSION="2.8.11"
3. run devon aws setup again

### Related/Dependent Issues
This has to implement with or after issues #978 and #964

CREITZ25:
This issue will be implemented together with #978 and #964

Issue #964, opened 2022-11-01
make aws config local


hohwille:
devonfw-ide started supporting `aws` CLI tool (see issue #711).
However, I noticed that the configuration remains global in the users home directory (in `~/.aws/` it creates files `config` and `credentials`).
Assuming you have two different projects for different customers with two different installations of devonfw-ide and their own `aws` CLI installation.
Now, these two should not interfere with each other so I should be able to do `aws configure` for both of them with totally different AWS accounts.
However, currently configuring one of them also configures the other and changes affect both/all installations.

Instead, we should tell `aws` to read the configuration from somewhere under `${DEVON_IDE_HOME}/conf/` (e.g. in `aws` or `.aws` subfolder).
Hint: We have archived the same already for tools like `mvn` (maven) or `az` (azure CLI).

cinnamon-coder-hub:
While looking at this issue, I stumbled upon a setup problem. On my computer the msi file won't be installed. 
It turned out, that `powershell.exe -Command "Start-Process msiexec.exe -verb runas -Wait -ArgumentList '/I ${windows_path_to_package}\aws-*-windows.msi /quiet'"` won't work on my computer, because the asterix sign  `*`  in `aws-*-windows.msi` won't be replaced by the version number. So I tried to fix this with [the PR 989](https://github.com/devonfw/ide/pull/989). During another free moment I'll be looking at the actual problem mentioned by @hohwille.

CREITZ25:
This issue will be implemented together with #978 and #992

PR #1002, opened 2022-12-08
#978 support for msi and adjustment of aws commandlet

CREITZ25:
#978 support for msi and adjustment of aws commandlet

### DIFF
--- a/scripts/src/main/resources/scripts/command/aws
+++ b/scripts/src/main/resources/scripts/command/aws
@@ -13,7 +13,7 @@ fi
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
 AWS_HOME="${DEVON_IDE_HOME}/software/aws"
-TOOL_VERSION_COMMAND="${DEVON_IDE_HOME}/software/aws/aws --version"
+TOOL_VERSION_COMMAND="${AWS_HOME}/aws --version"
 # shellcheck source=scripts/commandlet-cli
 source "$(dirname "${0}")"/../commandlet-cli
 


CREITZ25:
This PR implements Issues #978, #992 and #964

scripts/src/main/resources/scripts/functions
```diff
@@ -701,8 +701,17 @@ function doExtract() {
...
     extracted="no"
+  elif [ "${ext}" = ".msi" ]
+  then
+    local windows_download_file
+    windows_download_file=$(cygpath -w "${1}")
+    local windows_target_dir
+    windows_target_dir=$(cygpath -w "${target_dir}")
+    echo "Start-Process powershell.exe -ArgumentList \"msiexec /a ${windows_download_file} /qn TARGETDIR=${windows_target_dir}\" -Wait -NoNewWindow" >"${TMP}"/test.ps1
+    < "${TMP}"/test.ps1 powershell.exe
+    [ -f "${TMP}"/test.ps1 ] && rm "${TMP}"/test.ps1
```
hohwille, line 714: Instead of creating a script to run in powershell another powershell that runs our actual command and then deleting the script again, we can simply call the command directly.
The only special case to consider is that we are in bash here and have to use double slashes for windows command options:
```
msiexec //a "${windows_download_file}" //qn "TARGETDIR=${windows_target_dir}"
```

hohwille (commented):
@CREITZ25 thanks for this PR - great solution 👍 
Your approach with `AWS_CONFIG_FILE` looks perfect to me. Also I love the simplification of the setup and the finally local installation.
BTW: In `${target_dir}` I do not only find the extracted MSI content but also the MSI file again. Is there some option for `msiexec` to prevent this or should we simply delete the MSI copy ourselves after calling `msiexec`?

hohwille:
I linked the issues implemented by this PR. Can you also add them to the CHANGELOG?

scripts/src/main/resources/scripts/functions
```diff
@@ -744,7 +753,7 @@ function doInstallWithPackageManager() {
 # $1: filename (e.g. ${filename})
 function doIsInstallerExtension() {
   local myext="${1/*\./.}"
-  if [ "${myext}" = ".msi" ] || [ "${myext}" = ".exe" ] || [ "${myext}" = ".pkg" ] || [ "${myext}" = ".bat" ] || [ "${myext}" = ".py" ]
+  if [ "${myext}" = ".exe" ] || [ "${myext}" = ".pkg" ] || [ "${myext}" = ".bat" ] || [ "${myext}" = ".py" ]
```
hohwille, line 756: See #978 this will break at least `az` commandlet, maybe also docker/rancher?
https://github.com/devonfw/ide-mirrors/blob/88a8f841b353913166c3809bfb593276841fdb67/az/ext-mappings
Can you add `noUnpack` to affected `commandlets` to keep their behavior unchanged.
Maybe this flag has to be set OS dependent...

scripts/src/main/resources/scripts/functions
CREITZ25: A great solution, I will implement this.

scripts/src/main/resources/scripts/functions
CREITZ25: az do not use doInstall, it uses "doDevonCommand pip install azure-cli az".
No changes are necessary.
docker and rancher are not affected, because they use exe files for installation.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -12,85 +12,72 @@ fi
...
   else
-    doEcho "Installating AWS for Linux..."
-    if [ -L "/usr/local/bin/aws" ]
+    if [ "${ERR}" = 0 ]
     then
-      doRunCommand "sudo ${path_to_package}/install -u"
-    else
-      doRunCommand "sudo ${path_to_package}/install"
+     doRunCommand "${AWS_HOME}/install -i ${AWS_HOME} -b ${AWS_HOME}"
+     rm -rf "${AWS_HOME}"/README.md "${AWS_HOME}"/THIRD_PARTY_LICENSES "${AWS_HOME}"/install "${AWS_HOME}"/dist
```
hohwille, line 70: This seems problematic according to legal compliance.
Removing license information this way is typically "illegal" (infringement of license).
Either we have to ensure (only on linux!) that we first let `doInstall` do the installation to a temporary directory and then we run `install` with `-b ${AWS_HOME}`.

As a quick fix, you can only `rm` the `dist` folder here.

scripts/src/main/resources/scripts/command/aws
```diff
@@ -12,85 +12,72 @@ fi
...
-    sudo installer -pkg "${path_to_package}"/aws-"${version}"-mac.pkg -target /
+    if [ "${ERR}" = 0 ]
+    then
+      xar -C "${AWS_HOME}" -xf "${AWS_HOME}"/aws-"${version_to_install}"-mac.pkg
+      tar -C "${AWS_HOME}" -xzvf "${AWS_HOME}"/aws-cli.pkg/Payload
+      rm -rf "${AWS_HOME}"/Distribution "${AWS_HOME}"/Resources "${AWS_HOME}"/aws-cli.pkg
+      [ ! -h "${AWS_HOME}/aws" ] && ln -s "${AWS_HOME}"/aws-cli/aws "${AWS_HOME}"/aws
+      [ ! -h "${AWS_HOME}/aws_completer" ] && ln -s "${AWS_HOME}"/aws-cli/aws_completer "${AWS_HOME}"/aws_completer
+      [ -f "${AWS_HOME}"/aws-"${version_to_install}"-mac.pkg ] && rm "${AWS_HOME}"/aws-"${version_to_install}"-mac.pkg
+    fi
```
hohwille, line 65: I would love to see all such magic in do install but in a generic way (autodelete the pointless MSI copy that msiexec is creating for whatever reason that nobody can understand, also add extraction of pkg files to doExtract, etc.).
However, to keep it simple, I will create a new story for this.

hohwille (changes requested):
@CREITZ25 Thanks for this PR. You did a really great job 👍 Thanks also for testing and implementing on all OS/platforms what is quite some work.
Can you address the legal issue with removing license information?
Then we could merge this and I will create a new issue for the rest of my suggestions for further improvement?

scripts/src/main/resources/scripts/command/aws
CREITZ25: I cancel deleting the files README.md and THIRD_PARTY_LICENSES.

scripts/src/main/resources/scripts/command/aws
hohwille: I created #1010 for this

hohwille (approved):
@CREITZ25 Thanks for the quick rework.
I approve and will merge.
For the other discussion I created #1010

PR #1013, opened 2022-12-14
#1001: Adding wsl1 check for dockerD and rancherD

### LINKED ISSUE #1001: Problem using Rancher Desktop for users with wsl 1 on Windows
While setting up rancher desktop on windows computers, cases occured in which the rancher-desktop launcher failed with an rather cryptic error output. If I remember it correctly, those cases included the error code 4294967295. Some solutions relating to this error code are given on [stackexchange](https://stackoverflow.com/questions/70953842/error-wsl-exe-exited-with-code-4294967295-on-installing-rancher-desktop).

### Expected behavior

No error code should be thrown out by the rancher-desktop launcher.

### Affected version:

OS: Windows

### Solution
The problem lies in the usage of wsl 1 instead of wsl 2, but can also relate to disabled virtualization (see comment following)
Suggested steps for solving the bug: 

- Recognize that wsl 1 is used
- If so, execute `wsl --update` as admin
- Only start the installation of rancher desktop after these steps

cinnamon-coder-hub:
fixes #1001.

### DIFF
--- a/scripts/src/main/resources/scripts/command/docker
+++ b/scripts/src/main/resources/scripts/command/docker
@@ -36,6 +36,10 @@ function doResolveDownloadUrl() {
 
 # $1: optional setup
 function doSetup() {  
+  if doIsWsl1Used; then
+    #update from wsl1 to wsl2, else the following docker editions won't work
+    wsl --update
+  fi
   if [ "${DOCKER_EDITION}" != "DOCKERDESKTOP" ]
   then
     echo "Rancher Desktop set as option"
@@ -116,6 +120,26 @@ function doRun() {
   doSetup silent
   doEcho "Running: docker ${*}"
   docker "${@}"
+}
+
+function doIsWsl1Used() {
+{
+  echo '@echo off'
+  echo 'ver > tmp-kernel-versiontest-result.txt'
+} > tmp-kernel-versiontest.bat
+
+cmd "/C tmp-kernel-versiontest.bat"
+
+groupVersion=$(cat tmp-kernel-versiontest-result.txt | sed 's/^.*[^0-9][0-9]*\.[0-9][0-9]*\.\([0-9][0-9]*\)\.[0-9][0-9]*.*$/\1/')
+
+rm tmp-kernel-versiontest.bat
+rm tmp-kernel-versiontest-result.txt
+
+if [ "${groupVersion}" -lt 18917 ]; then
+  return 0
+else 
+  return 1
+fi
 }
 
 # OS independent check if docker is installed.
--- a/scripts/src/main/resources/scripts/command/docker
+++ b/scripts/src/main/resources/scripts/command/docker
@@ -36,10 +36,11 @@ function doResolveDownloadUrl() {
 
 # $1: optional setup
 function doSetup() {  
-  if doIsWsl1Used; then
-    #update from wsl1 to wsl2, else the following docker editions won't work
+  if ! doIsWsl2Used; then
+    #update wsl kernel and later on from wsl1 to wsl2, else the following docker editions may not work properly
     wsl --update
   fi
+  wsl --set-default-version 2
   if [ "${DOCKER_EDITION}" != "DOCKERDESKTOP" ]
   then
     echo "Rancher Desktop set as option"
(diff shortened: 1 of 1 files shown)


hohwille:
Build failed:
```
Error:  Failed running test #3: test-shellcheck - exit code 1

In scripts/command/docker line 133:
groupVersion=$(cat tmp-kernel-versiontest-result.txt | sed 's/^.*[^0-9][0-9]*\.[0-9][0-9]*\.\([0-9][0-9]*\)\.[0-9][0-9]*.*$/\1/')
                   ^-- SC2002 (style): Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
```

scripts/src/main/resources/scripts/command/docker
```diff
@@ -118,6 +122,26 @@ function doRun() {
   docker "${@}"
 }
 
+function doIsWsl1Used() {
+{
+  echo '@echo off'
+  echo 'ver > tmp-kernel-versiontest-result.txt'
+} > tmp-kernel-versiontest.bat
+
+cmd "/C tmp-kernel-versiontest.bat"
```
hohwille, line 131: We should stop doing such hacks.
Why creating a script containing one command instead of executing the command directly?
Why even writing a temporary file at all here?
Example:
```
echo $(wmic os get caption)
 Microsoft Windows 10 Pro
```
Or a generic way for calling via CMD:
```
$ MSYS_NO_PATHCONV=1 cmd /c ver

Microsoft Windows [Version 10.0.19044.2251]

hohwille@CE49454 MSYS /d/projects/mmm
$
```

BTW: I just had a similar discussion with @CREITZ25 in another PR review where a powershell script was created temporary and then called via powershell. The `docker` comandlet is full of such things and most probably none of them are required...

scripts/src/main/resources/scripts/command/docker
```diff
@@ -118,6 +122,26 @@ function doRun() {
...
+
+groupVersion=$(cat tmp-kernel-versiontest-result.txt | sed 's/^.*[^0-9][0-9]*\.[0-9][0-9]*\.\([0-9][0-9]*\)\.[0-9][0-9]*.*$/\1/')
+
+rm tmp-kernel-versiontest.bat
+rm tmp-kernel-versiontest-result.txt
+
+if [ "${groupVersion}" -lt 18917 ]; then
```
hohwille, line 138: IMHO we can not determine from the windows version what the `wsl` version actually is.
When I have WSL1 and do `wsl --update` I should have WSL2 then but this does not change my windows version, or does it?

scripts/src/main/resources/scripts/command/docker
```diff
@@ -36,6 +36,10 @@ function doResolveDownloadUrl() {
 
 # $1: optional setup
 function doSetup() {  
+  if doIsWsl1Used; then
+    #update from wsl1 to wsl2, else the following docker editions won't work
+    wsl --update
```
hohwille, line 41: Don`t you need admin persmissions to run this?

cinnamon-coder-hub: No, at least not if there are no applyable updates available. 
So I can just execute `wsl --update` in a git-bash prompt without admin rights.

hohwille: This is because you already have wsl2 so this command is doing pretty much nothing.
We need to properly test this otherwise we are publishing and untested feature that will most probably never work.

cinnamon-coder-hub:
Now, with [commit 5609197...](https://github.com/cinnamon-coder-hub/ide/commit/5609197f2b12c1869c6658513c31992e5ebb8935) I went the way I actually wanted to take by testing for wsl 2, instead of wsl 1,
as there may be different ways to check for wsl 1, but I'm not sure if they work only in some cases. 
This is because I cannot test this on my computer. Even when trying to downgrade to wsl 1 in various ways, e.g. by deleting all wsl distributions before, followed by `wsl --set-default-version 1` and rebooting, it still seems as if I'm treated differently then a wsl1 only user (as a previous wsl2 user). 
When installing rancher desktop, I'll still be able to install it, after downgrading as described above. After launching rancher desktop I can also see its two wsl distributions running under wsl 2. Only if I disable necessary virtualization features, then I get into trouble during the installation of rancher desktop (as described above). 

Also I included the command `wsl --set-default-version 2` as the command `wsl --update` wouldn't change my default wsl version from 1 to 2. I guess this switch only happens for users that switch from wsl 1 to wsl 2 for the first time,
but I can't test it. 

The most efficient alternative I can think of for now is to get an user that has always used wsl1, to give us feedback on what some commands do. 
E.g. I've read that `wsl -l -v` should return an error or nothing for "pure" wsl 1 users. Based on this information we could implement the wsl1 test, but for now I can only treat this as a rumor. 
Another way would be to try out setting up a VM with an older Windows 10. Here I'm afraid that the virtualization inside the VM may not work properly. Also this would cost more time. 

IMHO we should go with the wsl 2 check for now, as long as there is no less time excessive way (as for example trying this out on a virtual machine). If needed, this could be tackled in a few weeks, after the epic 941 is in a more advanced state, or at least my issue 1007 (with a higher priority).
Also this way should work for all users, only with the compromise that users without installed wsl 2 distributions will need to wait a few seconds more to run `wsl --update`.

hohwille:
I am finally closing this PR as it is not likely to become ready for merge.
I have added the problem to the documentation and described the workaround.
If we find a way to properly and reliably determine the `wsl` version and detect if WSL1 is present and no WSL2 is available, we can try to create a new PR to automate the workaround. However, most likely this is an edge-case and it is loosing its importance over time. Who will have WSL1 in 2023 and who in 5 years?

cinnamon-coder-hub:
> I am finally closing this PR as it is not likely to become ready for merge. I have added the problem to the documentation and described the workaround. If we find a way to properly and reliably determine the `wsl` version and detect if WSL1 is present and no WSL2 is available, we can try to create a new PR to automate the workaround. However, most likely this is an edge-case and it is loosing its importance over time. **Who will have WSL1 in 2023 and who in 5 years?**

Yes, I agree, especially to your last sentence. The effort of finding a way to simulate this scenario seems to exceed the benefit to heavily, but the effort for a user, with this nowadays rather rare version to update, seems to be  more acceptable, as there should already be few wsl 1 users.

Issue #1005, opened 2022-12-09
Proper error message if bash is not found


hohwille:
In our [setup](https://github.com/devonfw/ide/blob/master/documentation/setup.asciidoc) we have described that we need `git` installed as a pre-requesite. On windows this means git-for-windows that includes git-bash.
However, our user-base is growing and we happen to have users that do not read this properly.
As a result the users get strange errors with not very helpful messages like this:
```
""\bin\bash.exe"" kann syntaktisch an dieser Stelle nicht verarbeitet werden.
```

Bash detection starts here:
https://github.com/devonfw/ide/blob/d867f9a956b0eddc2c01f9b3a0e494edbf12e2ba/scripts/src/main/resources/scripts/devon.bat#L61

IMHO falling back to `cygwin` or whatever is rather pointless.
For KISS we could simply fail if we do not find `git-bash` with a proper error message if git[-bash] was not found like this:
```
ERROR: Could not find bash. It seems git for windows is not installed on your machine (GIT_HOME=)
Please download and install git for windows from the following URL and after that rerun devonfw-ide setup:
https://git-scm.com/download/win
```

hohwille:
Cool, there also is an actual bug in the current code:
https://github.com/devonfw/ide/blob/d867f9a956b0eddc2c01f9b3a0e494edbf12e2ba/scripts/src/main/resources/scripts/devon.bat#L78

Note that in this line unlike the others it says `exists` instead of `exist` so most likely this evaluates to true for some reason and so the actual error message never shows:
https://github.com/devonfw/ide/blob/d867f9a956b0eddc2c01f9b3a0e494edbf12e2ba/scripts/src/main/resources/scripts/devon.bat#L85

mustafaokyay:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IDE-61dbfc20045ead0022812582/issues/devonfw/ide/1005?planning-poker) with Zenhub @alfeilex @Amueller36 @cinnamon-coder-hub @CREITZ25 @hohwille @tobka777 @Ttring

PR #1014, opened 2022-12-15
#1005: update devon.bat to show bash error

Ttring:
- add pause before goto :eof to prevent terminal from closing immediately (if git bash is not installed)
- add new subroutine called "git_installation". This allows git for windows installation conditions to be executed/checked.


hohwille (approved):
@Ttring thanks for this PR. 👍 
Nice that you also found and fixed typo besides fixing the `exist` keyword and improving the error message.
I simply added an extra echo to get the GIT_HOME so we can better do support and debugging if in the future users run again into this problem. Now ready for merge.

hohwille:
Oh, now after this being merged, I see several changes that I did not see in the diff before.
Either github played a trick on me or I did not scroll to the end.
It is great that you added `pause` after the error but you also added various other `pause` statements that are more for debugging and now have been merged. I need to remove this.

hohwille:
I fixed this with PR #1019

Issue #1015, opened 2022-12-15
Fix jenkins for integration-test


alfeilex:
Currently, there is an error with `jenkins` in the integration-tests. 

This is the output:
```
Running with Java 19 from D:\a\ide\ide\scripts\target\integration-test\test-setup\software\java, which is not yet fully supported.
Run the command again with the --enable-future-java flag to enable preview support for future Java versions.
Supported Java versions are: [11, 17]
See https://jenkins.io/redirect/java-support/ for more information.
```

It seems that `jenkins` only runs on Java 11 or Java 17. We should update the integration-test for `jenkins`. For example to try the `--enable-future-java` flag or by setting the Java version to 17 and reverting to the default version after the integration-test for `jenkins` is finished.

Issue #1026, opened 2022-12-19
Focus on LTS releases by default


maybeec:
As a user first accessing the devonfw ide, I want proper setup, which is as best compatible with my project as possible so that I don't need to reconfigure too much before getting first successful build.

For sure, this is a problem without a solution. Nonetheless, currently especially for software packages like java or node, you will get the latest version, which might be incompatible with quite some frameworks you might use by default. Therefore, it would be recommendable if we will focus by the default settings to the LTS versions, which could give a better "first-shot" compatibility and thus does not stop people immediately to further investigate devonfw IDE.

maybeec:
This idea came basically out of the experience made with #1024

hohwille:
We shall consider this in #941 but it might be tricky to automatically determine what version is an LTS version as the tool maintainers do not give us an API telling such information. Maybe we have to maintain that in a config file (status.json? - IMHO better an optional dedicated file `latest-version` that can simply contain e.g. `17*` for Java so the latest 17.x version would be installed and we simply maintain such files by hand) and honor it.

maybeec:
I feel, it could even be part of the manually configured standard ide-settings repo. This would give transparency for configuration to people who want to adapt to their projects as some values are already declared in the devon.properties file. Furthermore, I don't see a value to have again another concept here. What would be the value?

hohwille:
@maybeec thanks for your feedback and suggestion. As I like KISS I fully agree. So we can easily address this [here](https://github.com/devonfw/ide-settings/blob/master/devon.properties). Further, we do not even have to wait for anything related to story #941 - great!

alfeilex:
This PR https://github.com/devonfw/ide-settings/pull/28 adds LTS versions for some tools. 

- dotnet
- java
- nodejs
- python
- sonartype

Have I missed something here? Because most of the tools does not follow the LTS strategy. In most cases, the latest version is installed immediately.

PR #1027, opened 2022-12-21
#1015: Fix jenkins for integration-test (set Java 17 as default version)

alfeilex:
Currently, there is a warning when testing `jenkins` informing that running `jenkins` with Java 19 version is not fully supported. In this PR, the Java 17 version is set as default for the integration-tests. This also makes more sense since Java 19 is not LTS version. 

Related issue: #1015


hohwille (approved):
@alfeilex thanks for analyzing and fixing the problem 👍 
Ready to be merged.
See also #1026

Issue #1004, opened 2022-12-08
Support for gcloud CLI


hohwille:
As a devonfw-ide user, I want to automate setup of `gcloud` (google cloud CLI) so that I can also devonfw-ide seamlessly in my google cloud project.

https://cloud.google.com/sdk/gcloud
https://cloud.google.com/sdk/docs/install

alfeilex:
The documentation suggests installing `gcloud` via`.exe` file. Fortunately, Google provides versioned archives. We can use these to install gcloud via script on Windows, Linux and MacOS.

PR #1028, opened 2022-12-29
#1004: GCloud CLI integration for windows

alfeilex:
GCloud CLI is added to the ide as a commandlet. It only works for Windows systems as it requires Python to run. After Python is fixed for MacOS and Linux, we can update this command. 

Related issue: #1004

### DIFF
--- a/scripts/src/main/resources/scripts/command/gcloud
+++ b/scripts/src/main/resources/scripts/command/gcloud
@@ -10,11 +10,10 @@ then
   exit
 fi
 
-GCLOUD_HOME="${DEVON_IDE_HOME}/software/gcloud"
-TOOL_VERSION_COMMAND="${GCLOUD_HOME}/bin/gcloud.cmd --version"
-
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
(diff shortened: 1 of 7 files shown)


scripts/src/main/resources/scripts/command/gcloud
```diff
@@ -0,0 +1,79 @@
...
+  then
+    doRunCommand "${gcloud_config_export}"
+    echo -e "\n${gcloud_config_export}" >> "${DEVON_IDE_HOME}/conf/devon.properties"
+    doEcho "Location of GClouds's configuration files is set to ${gcloud_config_files}"
+  fi
+}
+
+# the path of gcloud modules must be added to python's sys.path via .pth file
+function doSetPth() {
+  local wsl_path="${GCLOUD_HOME}/lib"
```
hohwille, line 33: just a question: what does this path have to do with WSL (Windows Subsystem for Linux)?

hohwille (approved):
@alfeilex thanks for the PR. Looks good to me and great that you could already solve it for windows. 👍

scripts/src/main/resources/scripts/command/gcloud
hohwille: Should not block PR. Simply renamed it. Done.

Issue #999, opened 2022-12-01
Ability to require minimum devonfw-ide version


hohwille:
As a ide admin, I want to configure a minimum devonfw-ide version so that I can use a new feature and ensure that users with too old version of devonfw-ide will get notified and can update automatically.

Idea:
Add this do my `settings/devon.properties`: 
```
DEVON_IDE_VERSION=2023.01.001
```
(Better name it `DEVON_IDE_MIN_VERSION`?)
Users with e.g. `2022.12.001` containing this new feature already will then get this result when they setup devonfw-ide for the project:
```
Your version of devonfw-ide is currently 2022.12.001.
However, your project requires at least version `2023.01.001`.
Newer features will be required for your project so you need to update your devonfw-ide via the following command:
devon ide update scripts
Do you want to run the update now, so you can rerun "devon ide setup" after that? (yes/no)
```

hohwille:
Note: Actually only on windows this has to abort the setup due to file locking issues `devon ide update` will therefore `exit` on windows terminating the further process of our `ide` commandlet calling the functionality.

hohwille:
A good place where to add this check would be here:
https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/ide#L235

hohwille:
Retested with latest release successful:
```
******** ATTENTION ********
Your version of devonfw-ide is currently 2023.03.001
However this is too old as your project requires at latest version 2023.03.002
Please run the following command to update to the latest version of devonfw-ide and fix the problem before rerunning setup:

devon ide update scripts


We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
```

PR #1034, opened 2023-01-08
#999: add feature to check minimum require version

Ttring:
To test it, `DEVON_IDE_MIN_VERSION=2023.01.001` needs to be added in settings\properties. 

If the current Devon version is smaller than the minimum required version, it is updated to the minimum required version. 

If the current Devon version is greater than the minimum required version, it remains at the current version. 

If the given minimum version is not available, the operation is terminated with code 22.

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -46,7 +46,14 @@ function doUpdateMetaFile() {
 
 function doUpdateScripts() {
   # shellcheck disable=SC2007,SC2154
-  doUpgradeMavenArtifact "${DEVON_IDE_HOME}" "${DEVON_IDE_REPO_URL}" "devonfw-ide-scripts" "${target_version}" ".tar.gz" "$[devon_ide_version]"
+  local update_version
+  if [ -z "${1}" ]
+  then 
+    update_version="LATEST"
+  else 
+    update_version=${1}
+  fi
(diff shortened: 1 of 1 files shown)


scripts/src/main/resources/scripts/command/ide
```diff
@@ -236,10 +243,74 @@ function doSetup() {
...
+  local current_devon_version_month
   if [ ! -f "${SETTINGS_PATH}/devon.properties" ]
   then
     doEcho "Your settings are missing the 'devon.properties' file. Please ask your technical lead to properly merge and update your settings."
     doDownload "https://raw.githubusercontent.com/devonfw/ide-settings/master/devon.properties" "${SETTINGS_PATH}"
+  else
+    current_devon_version="$("${DEVON_HOME_DIR}/scripts/devon.bat" -v)"
+    min_devon_ide_version_month=${DEVON_IDE_MIN_VERSION##${DEVON_IDE_MIN_VERSION%%.*}.} 
+    current_devon_version_month=${current_devon_version##${current_devon_version%%.*}.} 
+    if [[ ${DEVON_IDE_MIN_VERSION%%.*} -gt ${current_devon_version%%.*} ]]
```
hohwille, line 257: we should use `doVersionCompare` for this as `-gt` is not suitable for comparing versions. In the worst case, we later have to change versioning schema again (see devonfw Teams discussion) and this will break and stop working.

hohwille (changes requested):
@Ttring thanks for this PR.
I assume we will take over and do the rework from the new team.

hohwille:
This PR is replaced by #1056 so the work is not lost and the story will be completed. Hence, I can now close this PR.

Issue #971, opened 2022-11-08
Ability to run "migration script" on update


hohwille:
When we publish a new release, our users can install it to existing devonfw-ide installations via `devon ide update scripts`.
Sometimes we do some structural changes, here is a brief history with just some examples:
* we have changed the eclipse configuration structure in `settings/eclipse/workspace`
* we have moved the eclipse plugin configuration from hard-coded settings to `settings/eclipse/plugins`
* we are planning to migrate from `ide-mirrors` to `ide-urls`

Due to such changes we sometimes "pollute" our scripts/commandlets with legacy handling code (see #904 or see `ide` commandlet with blocks of `cp` and `mv` commands, etc.).

Instead, we should create a `bash` script `scripts/migration` where we can place logic that would be executed automatically by `devon ide update scripts` but at the end on the new release. 
To avoid mistakes and make things simple for developers, I would propose that the `migration` script gets called with the previous version of devonfw-ide that just has been updated. We create a subfolder `scripts/migrations` where we can create scripts for specific migrations named by the source version. So e.g. when we have version `2022.11.001` installed and call `devon ide update scripts` then `migrate` script would find all scripts in `migrations` subfolder that are named by a version greater or equal to `2022.11.001` so he may find `migrations/2022.12.001` and execute it. In case someone knows flyway or liquibase this is the same pattern that we are trying to implement here...

We can still discuss if we need to track the version of devonfw-ide itself for reliability (e.g. in `.devon.software.version` file in `DEVON_IDE_HOME`).

Issue #1032, opened 2023-01-03
devonfw ide does not show correct version after updating to 2022.12.001


maybeec:
### Expected behavior

As a user, I want to get the actual version by `devonfw -v` command so that I can report the version properly to my bug report.

### Actual behavior

`devon -v` does not show the correct version number after upgrading to `2022.12.001` release:

![image](https://user-images.githubusercontent.com/1427255/210313216-76060c4a-a500-4a95-b0dd-b8a84ee219ae.png)

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Install 2022.08.002 devonfw ide
2. Update to 2022.12.001 version by `devon ide update all`

### Affected version:

* OS: Windows 10
* Console: Git Bash

hohwille:
At least to explain what you are observing:
You need to distinguish between the `devonfw-ide` installation and the `devon` command wrapper script.
You `devonfw-ide` installation has properly been updated and to get the "real" version you can simply do `devon help` and you will see the expected version.
However, the `devon[.bat]` wrapper script is not automatically overwritten on every update as usually this script does not change at all and it is "global" so reasonable changes here could cause compatibility problems so that we are also motivated to avoid them.

To address your actual concern:
IMHO the end-users expectation would be that `devon -v` would show the actual version of your devonfw-ide installation rather than the devon wrapper script. However, this can only work if you are inside a devonfw-ide installation (with your `CWD`) or have at least setup `devonfw-ide` from a devonfw-ide installation before calling it.
If that later is the case we could print the proper version. Maybe we can try to implement a trade-off and print both versions as available with reasonable context information so an end-user can understand everything.

maybeec:
Sounds like a good tradeoff. I feel the explanation makes sense but is completely non transparent to users. Especially for debugging or bug reporting purposes, people like to know the current version they are working with.

alfeilex:
While working on #971 I found out that after running `devon ide update scripts` the `devon.bat` of the previous version is copied to `USERPROFILE/scripts`. So if the updated IDE had version `2022.11.001` and I update, the copied `devon.bat` will contain the previous version and `devon -v` will return this previous version. Is this the right behavior? Maybe the `cp` commands should be executed after the update is complete?
https://github.com/devonfw/ide/blob/70baf93ecffb8967f34be86c3ac6f4465b911af3/scripts/src/main/resources/scripts/command/ide#L54-L60

hohwille:
[...]

PR #1038, opened 2023-01-16
#971: Ability to run "migration script" on update

alfeilex:
This PR implements the `migration` command which is executed after `devon ide update scripts`, for example. The script runs various migration scripts located in the folder  `DEVON_HOME/scripts/migrations`.

Example: 
The current IDE version is `2022.03.001` and the user updates the IDE to the latest version. So all migrations scripts >= `2022.03.001` will be executed for the migration. 

Related issue: #971


scripts/src/main/resources/scripts/command/migration
```diff
@@ -0,0 +1,15 @@
...
+
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+for migration in "$(dirname "$0")/../migrations"/*
+do
+  from_version="${1}"
+  to_version="${migration/*\//}"
+  doVersionCompare "${from_version}" "${to_version}"
+  result="${?}"
+  if [ "${result}" -eq 0 ] || [ "${result}" -eq 2 ]
```
hohwille, line 11: The filesystem can give us files (migrations) in random order.
Is the an option for `sort` so we can compare such that the segment numbers are sorted numeric and not alphabetical (e.g. `1`, `2`, ..., `10` and not `1`, `10`, ... - also `1`, `02`, `003`). Otherwise we would need to implement sorting ourselves using `doVersionCompare` what will be very slow...

hohwille (changes requested):
@alfeilex Thanks for your PR. Looks already good to me. 👍 
The spot where you integrated the change is fine.
We should discuss if `migration` should be a commandlet. The impact is that end-users will get this via code-completion and `devon help` while IMHO it is something internal not intended for end-users. So I would rather move this to the `scripts` folder out of `command` folder.
IMHO we should consider creating a `documentation/migration.asciidoc` file as well.

scripts/src/main/resources/scripts/command/ide
```diff
@@ -63,6 +63,10 @@ function doUpdateScripts() {
       exit 0
     fi
   fi
+  local current_version
+  current_version="$("${DEVON_IDE_HOME}/scripts/devon" -v)"
+  doEcho "\nRun migration scripts to version ${target_version} ..."
+  doDevonCommand migration "${current_version}"
```
hohwille, line 69: One more thing. In the refinement meeting when introducing the issue story, I showed that on windows due to file locking we have to do the update asynchronous.
So this will most probably not work on windows. See here:
https://github.com/devonfw/ide/blob/c53d140ccc881b869a9e082a599f25a9379e7320/scripts/src/main/resources/scripts/functions#L779

alfeilex: Could we also run the the migrate script for windows like this? 
 `(sleep 10;doBackup "${target}" "${3}";doRunCommand "mv '${1}' '${target}'";doRunCommand "${DEVON_IDE_HOME}/scripts/migration" "${current_version}") & `
 
 I don't see another option as we have to run the migration scripts after the `backup` function.

scripts/src/main/resources/scripts/command/migration
alfeilex: I have added a `sort -n` command to order the migrations script numerically.

Check commit b548765c7441a3edba09d716db9d91f3c307bbaf

scripts/src/main/resources/scripts/migrations/2022.12.001
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+echo "test migration to 2022.12.001"
```
hohwille, line 3: Great to have a test, but we should create something real instead.
For the moment we could simply remove this having no migrations at all.
Once the feature is released (for a while), we could add a migration to address [tool excuses](https://github.com/devonfw/ide/issues/826#issuecomment-1271311469). So we should have `software/mvn` instead of `software/maven` as the commandlet is also named `mvn` and the same also for `node` instead of `nodejs`.

scripts/src/main/resources/scripts/migration
```diff
@@ -0,0 +1,18 @@
...
+
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/functions
+
+doEcho "\nRun migration scripts to version ${1}."
+migrations="$(dirname "${0}")/migrations/*"
+for migration in $(ls ${migrations} | sort -n)
+do
+  from_version="${1}"
+  to_version="${migration/*\//}"
```
hohwille, line 11: could be simplified, if you just `cd` to the migrations folder (first check if the folder exists and otherwise log a warning what will also help if something goes wrong) and then simply use globbing with `sort -n`. Then you will only get the plain version and do not have to extract the filename from the entire path.
However, just a nice-to-have comment - nothing wrong with your code.

scripts/src/main/resources/scripts/command/migration
hohwille: Seems to be the perfect solution (simple and efficient). As long as we do not change our versioning scheme to something odd it should work once and forever. Thanks.

scripts/src/main/resources/scripts/command/ide
hohwille: Yes, you found the perfect solution and implemented it already. Thanks. So I am resolving this.

hohwille:
I found a potential conflict of requirements:
#1032 suggests to change the output of `devon -v` and this PR is using it in order to determine the current version installed.
Maybe we can easily solve this when we address my review suggestion to maintain the version in `${DEVON_IDE_HOME}/.devon.software.version`.

scripts/src/main/resources/scripts/functions
```diff
@@ -780,21 +780,28 @@ function doReplaceExtractedFile() {
   local filename="${1/*\//}"
   local target_dir="${DEVON_IDE_HOME}"
   local target="${DEVON_IDE_HOME}/${filename}"
+  local migration="${DEVON_IDE_HOME}/scripts/migration"
+  local curr_version
+  curr_version="$("${target_dir}/scripts/devon" -v)"
```
alfeilex, line 785: Should we remove the option to specify the current version of the IDE with `devon -v` as first parameter at all? 

Further, should we have the newly created `.devon.software.version` automatically updated after each release?

scripts/src/main/resources/scripts/migration
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/functions
+
+from_version=2023.0.001
```
hohwille, line 6: typo. Should be `2023.02.001`.

hohwille: I quickly resolved this.

scripts/src/main/resources/scripts/migration
```diff
@@ -0,0 +1,34 @@
...
+    result="${?}"
+    if [ "${result}" -eq 0 ] || [ "${result}" -eq 2 ]
+    then
+      doEcho "Run migration to version ${migrate_to_version}"
+      bash "${migrate_to_version}"
+      migration_result="${?}"
+      if [ "${migration_result}" -ne 0 ]
+      then
+        doAskToContinue "The migration to version ${migrate_to_version} failed. Would you like to continue?"
+      fi
```
hohwille, line 29: if no error happened, we need to update `${DEVON_IDE_HOME}/.devon.software.version)` to `${migrate_to_version}`.

scripts/src/main/resources/scripts/migration
```diff
@@ -0,0 +1,34 @@
...
+MIGRATIONS_DIR="${DEVON_IDE_HOME}/scripts/migrations"
+
+if [ -d "${MIGRATIONS_DIR}" ]
+then
+  cd "${MIGRATIONS_DIR}"
+  for migrate_to_version in  $(ls | sort -n)
+  do
+    doVersionCompare "${from_version}" "${migrate_to_version}"
+    result="${?}"
+    if [ "${result}" -eq 0 ] || [ "${result}" -eq 2 ]
```
hohwille, line 21: after rethinking this, we should only migrate if `${from_version}` is less `${migrate_to_version}` and NOT if equal (see previous review comment).
If we agree and change this accordingly, we need to change `from_version=2023.02.001` to `from_version=2023.01.001`.
If I am not missing something and thought everything through correctly, all should be working perfectly then.

hohwille (changes requested):
@alfeilex Thanks for this PR. Looks quite good now 👍 
Sorry, but I still have some last review comments for rework to make this fully functional.

scripts/src/main/resources/scripts/migration
alfeilex: Should work now

scripts/src/main/resources/scripts/migration
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/functions
+
+from_version=2023.01.001
+if [ -f "${DEVON_IDE_HOME}/.devon.software.version" ]  
+then
+  devon_version="${DEVON_IDE_HOME}/.devon.software.version"
```
hohwille, line 9: should be defined outside the if (and used for if condition). Otherwise line 31 can fail.

hohwille (approved):
@alfeilex Thanks for the update 👍 
Ready for merge now.

scripts/src/main/resources/scripts/migration
hohwille: Done. Thanks.

scripts/src/main/resources/scripts/migration
hohwille: Thanks.

scripts/src/main/resources/scripts/migration
hohwille: Done.

Issue #1010, opened 2022-12-13
further improvements for extraction of MSI and PKG


hohwille:
As discussed in PR #1002 I want to further improve our `doInstall` function to avoid quirks in the commandlets:
* [ ] with #978 we have added support for MSI extraction. However, `msiexec` does not only extract the MSI file but also creates a copy of the MSI in the target folder. This is undesired. As I did not find an option for `msiexec` to prevent this nonsense, we should simply delete the copy after calling `msiexec`.
* [ ] the same way we support `MSI` now (#978), we should also support `pkg` packages on macos and automatically extract them rather than doing this manually in our commandlets. Again everything can be found in #1002 (see my review comments)
* [ ] we already do support special cases where an extracted package only contains a single folder that then contains the actual content. If so, we simply take that content and therefore omit the single folder automatically. For e.g. AWS we would need to extend this so it works in a loop so that also a 2nd level single folder can be omitted. However, besides `*.app` folders we would also need to exclude `bin` folders to avoid accidental buggy behavior:
https://github.com/devonfw/ide/blob/478794db3388f7cd80e11f937facb7d03086eb38/scripts/src/main/resources/scripts/functions#L809
* [ ] at the end we should remove the "workarounds" from the AWS commandlet that are now already handled by our updated functions.

Examples to make the last point clear (the asterisk * marks the folder containing the content that should appear inside the target folder such as `software/«tool»`):
```
«package1 root»
└── Amazon
   └── AWSCLIV2 (*)
      ├── _asyncio.pyd
      ├── ...
      ├── aws.exe
      ├── ...
      └── wheel-0.37.0.dist-info

«package2 root» (*)
└── bin
  └── my-tool

«package3 root»
└── Foo (*)
  └── bin
     └── my-tool

«package4 root»
└── Foo (*)
  └── MacOs.app
      └── Content
          ├── bin
          ├── ...
          └── README.txt

```

CREITZ25:
How to see the configuration of a msi Archive:
https://learn.microsoft.com/en-us/windows/win32/msi/orca-exe

PR #1041, opened 2023-01-23
Feature/support for msi

CREITZ25:


CREITZ25:
#1010

scripts/src/main/resources/scripts/functions
```diff
@@ -817,13 +828,31 @@ function doReplaceExtractedSkipSingleFolder() {
...
+      if [ -d "${dir}/${files}" ] && [ "${files}" != "bin" ] && [ "${files%.app}" = "${files}" ]
+      then
+        dir="${dir}/${files}"
+      else
+        break
+      fi
+      files="$(find "${dir}" -maxdepth 1 -mindepth 1 -exec basename {} \;)"
+      anz_files="$(echo "${files}" | wc -w)"
+    done
+    source_dir="${dir}"
```
hohwille, line 852: In case you changed `source_dir` the above debug message would be incorrect.
Can you simply move it down here?

hohwille (approved):
@CREITZ25 this now looks good and seems functionally correct. Thanks for your work 👍 
Can you still have a look at my last 2 review comments. IMHO you can address both in a few minutes.

scripts/src/main/resources/scripts/functions
CREITZ25: Yes, of course. I move it to the end and adapt the variables

CREITZ25:
One correction:

[ -f "${downloadfile}" ] && rm -rf "${downloadfile:?}"

is no longer necessary. A few commits ago the expression was

[ -f "${var1}/${var2}" ] && rm -rf "${var1}/${var2}" 

which could result in a "/" which results in rm -rf "/"
Changing the expression to ${downloadfile} fix this error

Issue #1030, opened 2023-01-03
Rancher Desktop Installation not detected - ide is trying to install


maybeec:
### Expected behavior

As a user, I want to use docker so that I can work on my current project task. Running `devonfw docker` I want to get the setup for docker runtime OR start the available installation OR being asked to update, if my existing rancher desktop installation is outdated.

### Actual behavior

The devonfw ide does not detect the rancher desktop installation on windows and thus staring a setup resulting in an error (in my case) during installation, that "Rancher Desktop cannot be closed. Please close it manually and retry the installation".

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Have Rancher Desktop installed (in my case v1.2.1)
2. Run `devon docker` on windows
3. You get a setup dialog although an installation is already available (which you might forgot :D).

![image](https://user-images.githubusercontent.com/1427255/210313439-ae90d5e1-cf07-4a89-993b-71c8e2f0b820.png)

### Affected version:

* OS: Windows 10
* Console: Git Bash

maybeec:
I have to admit, that I installed Rancher Desktop manually some time ago, which then was not detected by devonfw ide.
I uninstalled it ran the installation again by devonfw ide, which is not working even accross devonfw ide installations on my machine. Nonetheless, that means, that we might need another detection mechanism to be clear, that we don't try to reinstall rancher or even worth install rancher over docker desktop

hohwille:
Checked here:
https://github.com/devonfw/ide/blob/70baf93ecffb8967f34be86c3ac6f4465b911af3/scripts/src/main/resources/scripts/command/docker#L146

Long time ago we were simply checking if `docker` command is available. We should read this to `doSetup` so if I already have docker installed (whatever way Rancher-Desktop, Docker-Desktop or an alternative we might not even know of) all should be fine as installing yet another docker does not seem to make sense. So lets just read-add this check again.

Ttring:
I could not reproduce the problem as shown in the picture. What I did was: install Rancher Desktop 1.7.0 manually and then run docker desktop.

However, I noticed that devon docker does not recognize that Rancher is already installed if the user tries to run devon docker in the same git bash terminal after installing Rancher Desktop. The possible solution would be to reinitialize the git bash terminal after installing rancher desktop.

Issue #100, opened 2019-04-16
bug if path containts whitespace


Volkan61:
The following error message occurred when starting the installation.

```
Welcome to devon-ide!
This product and its 3rd party components is open-source software.
However, before using it you need to read and accept the license agreement and t
erms of use.
You will be able to find it in one of the following locations:
/c/Users/****/Desktop/New folder/devonide/TERMS_OF_USE.adoc
https://github.com/devonfw/devon-ide/wiki/license

Do you accept these terms of use and all license agreements? (yes/no) yes

*** Software Update of devon-ide-settings ***
Updating /c/Users/****/Desktop/New folder/devonide/settings from https://rep
o.maven.apache.org/maven2/com/devonfw/tools/ide/devon-ide-settings
Trying to determine the latest available version from https://repo.maven.apache.
org/maven2/com/devonfw/tools/ide/devon-ide-settings/maven-metadata.xml
LATEST = 3.0.0-beta5
Starting installation from https://repo.maven.apache.org/maven2/com/devonfw/tool
s/ide/devon-ide-settings/3.0.0-beta5/devon-ide-settings-3.0.0-beta5.zip to /c/Us
ers/****/Desktop/New folder/devonide/settings
Downloading devon-ide-settings-3.0.0-beta5.zip from https://repo.maven.apache.or
g/maven2/com/devonfw/tools/ide/devon-ide-settings/3.0.0-beta5/devon-ide-settings
-3.0.0-beta5.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 68045  100 68045    0     0   474k      0 --:--:-- --:--:-- --:--:--  474k
Running command: unzip -q '/c/Users/****/Downloads/devon-ide-settings-3.0.0-
beta5.zip' -d '/c/Users/****/Desktop/New folder/devonide/updates/extracted'
Succeeded to run command unzip
Successfully extracted archive devon-ide-settings-3.0.0-beta5.zip to updates/ext
racted
Running command: mv '/c/Users/****/Desktop/New folder/devonide/updates/extra
cted/settings' '/c/Users/****/Desktop/New folder/devonide/settings'
Succeeded to run command mv
Successfully installed /c/Users/****/Desktop/New folder/devonide/settings
Initialized empty Git repository in C:/Users/****/Desktop/New folder/devonid
e/settings/.git/
[...]

PR #1045, opened 2023-01-30
#1030: Rancher Desktop Installation not detected.

Amueller36:
Issue #1030 
Added check right at the beginning of doSetup if ANY docker installation exists, the doSetup returns, because then no setup is required. In case of Rancher the Vpn Kit installation will be ensured aswell.


scripts/src/main/resources/scripts/command/docker
```diff
@@ -67,13 +69,12 @@ function doSetup() {
...
           # ensures installation of vpnkit if user directly restarts computer for enabling wsl
-          doRunCommand "${rancher_install_dir}/rancher-*-windows.exe" 
+          doRunCommand "${rancher_install_dir}/rancher-*-windows.exe"
           #check if vpnkit is already installed
           doEnsureVpnKitInstallation
         elif doIsMacOs
         then
-          # shellcheck disable=SC2016
-          doRunCommand 'cp -a "${rancher_install_dir}"/*.app /Applications/'
+          doRunCommand "cp -a ${rancher_install_dir}/*.app /Applications/"
```
hohwille, line 77: IMHO you introduced a bug here in case the path (`${rancher_install_dir}`) would contain a whitespace.
Can you put single quotes around it due to #100

hohwille: As constructive review, I simply added the quotes directly so we can already merge your great work.

hohwille (approved):
@Amueller36 thanks for this PR. Excellent work 👍 
Ready for merge

hohwille:
CLA check is haning, waiting for next change on master to merge and trigger rebuild/-check.
If the error persists, I will merge as admin since CLA has already been signed by @Amueller36

hohwille:
license/cla now green but the build will turn red now due to #1062. As soon as #1062 is fixed I will merge into this PR to fix build and finally merge.

Issue #998, opened 2022-11-30
release info messed with installation output


hohwille:
```
$ devon release -Dmmm.flatten.updatePom=false
 You are inside a top-level maven project
Trying to determine current version of your project...
Current version: Success: run command mkdir
Artifact already exists at /c/Users/hohwille/Downloads/devonfw-ide/maven-3.8.6.tar.gz
To force update please delete the file and run again.
Success: run command tar
Successfully extracted archive maven-3.8.6.tar.gz to /d/projects/mmm/updates/extracted/maven
Success: run command mv
Successfully installed maven
The software maven has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
17-SNAPSHOT
Release version: Success: run command mkdir
Artifact already exists at /c/Users/hohwille/Downloads/devonfw-ide/maven-3.8.6.tar.gz
To force update please delete the file and run again.
Success: run command tar
Successfully extracted archive maven-3.8.6.tar.gz to /d/projects/mmm/updates/extracted/maven
Success: run command mv
Successfully installed maven
The software maven has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
17
Next version: Success: run command mkdir
Artifact already exists at /c/Users/hohwille/Downloads/devonfw-ide/maven-3.8.6.tar.gz
To force update please delete the file and run again.
Success: run command tar
Successfully extracted archive maven-3.8.6.tar.gz to /d/projects/mmm/updates/extracted/maven
Success: run command mv
Successfully installed maven
The software maven has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
18-SNAPSHOT

Is next version correct?
(yes/no):
```

Expected output would be:
```
$ devon release -Dmmm.flatten.updatePom=false
 You are inside a top-level maven project
Trying to determine current version of your project...
Current version: 17-SNAPSHOT
Release version: 17
Next version: 18-SNAPSHOT

Is next version correct?
(yes/no):
```

Maybe we can simply first call some `doDevonCommand mvn setup silent` before invoking the maven commands.

PR #1046, opened 2023-01-30
#998: fixed release-info messed with installation output

Amueller36:
Added mvn setup call at the beginning of the command as stated by @hohwille. I couldn't reproduce the error so I assume that might be it.


Amueller36:
#998 related issue

.mvn/maven.config
```diff
@@ -1 +1 @@
--Drevision=2023.01.001-SNAPSHOT
+-Drevision=2023.01.001
```
hohwille, line 1: you should not commit changes to this file.

hohwille (changes requested):
@Amueller36 Thanks for this PR. Your fix is fine👍 
However, can you revert the change to `maven.config` please?

hohwille (approved):
@Amueller36 thanks for your PR. Workaround should fix the problem though it will install maven also if the release build will not use maven...

Issue #991, opened 2022-11-23
Consider integration of openrewrite


hohwille:
The tool [openrewrite](https://docs.openrewrite.org/) allows to define "recepies" that are a description of a migration and supports applying them to a project.
This allows to automate migrations in code projects.
The ecosystem of openrewrite already offers all kind of migrations for spring and spring boot but also for maven, kuberenetes, etc.

This could entirely replace our custom build migrations (see `devon java migrate`).
See also issue #226 (PR #597).

In the future we could offer migrations like e.g. `devon rewrite «recepie»`.

alfeilex:
My first assumption that we need to change the pom.xml to run the rewrite-maven-plugin was wrong. We can use a single command to run built-in recepies. 

For example:
`./mvnw org.openrewrite.maven:rewrite-maven-plugin:4.38.0:run -Drewrite.activeRecipes=org.openrewrite.java.format.AutoFormat`

alfeilex:
After testing some ideas for integratiing `openrewrite` into ide, I found that we can define the location of the `pom.xml`, which contains the dependencies for the recipe-plugin with `-f`.

Finally, we can use `devon rewrite <recipe-name>` as a wrapper for `devon maven org.openrewrite.maven:rewrite-maven-plugin:4.38.0:run -Drewrite.activeRecipes=<recipe-name> -f <path-pom>` and run external recipes.

alfeilex:
Some recipes are more complex and need a yaml file with configuration. See [here](https://docs.openrewrite.org/running-recipes/getting-started#step-5-run-a-recipe-with-yaml-configuration). A possible structure for all files needed for rewrite plugin could look like this:

```
conf/
--rewrite/
----pom.xml
----rewrite.yaml
```

alfeilex:
Questions left open after researching how to integrate rewrite into the devon ide:

- Do we offer all the extended recipes offered by open-rewrite (See [here](https://docs.openrewrite.org/reference/recipes))?
- Do we add a new command `rewrite` ? It is not a CLI and has not setup like previous commandlets.
- Or do we extend the `maven` commandlet?

alfeilex:
I don't think that we can integrate open-rewrite in the current manner of devon-ide. While built-in recipes can be run with
`devon maven org.openrewrite.maven:rewrite-maven-plugin:4.38.0:run -Drewrite.activeRecipes=<recipe-name>`
there are recipes for which plugin dependencies must be set in `pom.xml` file of the target project. 

[...]

PR #1047, opened 2023-01-31
#991: Integrate openrewrite feature

alfeilex:
This PR integrates [openrewrite-maven-plugin](https://github.com/openrewrite/rewrite-maven-plugin) into the IDE. With command `devon rewrite <recipe> [<catalog>]`, the user can run multiple recipes that are provided by openrewrite.

`recipe` stands for the name of the recipe like `org.openrewrite.java.format.AutoFormat` and the `catalog` parameter is optional in case that the recipe is provide by [external modules](https://docs.openrewrite.org/running-recipes/getting-started#step-6-running-recipes-from-external-modules).

Catalogue of external modules
https://docs.openrewrite.org/reference/recipes


alfeilex:
If the implementation of this "wrapper" is suitable for the IDE, i will write documentation on how to use it properly.

scripts/src/main/resources/scripts/command/rewrite
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+function doRun() {
+    local recipe="${1}"
+    local tool="${2}"
+
+    doRunCommand "mvn org.openrewrite.maven:rewrite-maven-plugin:4.39.0:run \
```
timtebeek, line 20: Optionally you can always pick up the latest version through
```suggestion
    doRunCommand "mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
```

hohwille: @timtebeek awesome. Thanks for your constructive hint. I am very happy to get unexpected feedback to ease my review work :)

scripts/src/main/resources/scripts/command/rewrite
```diff
@@ -0,0 +1,37 @@
...
+
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+function doRun() {
+    local recipe="${1}"
+    local tool="${2}"
+
+    doRunCommand "mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
+      -Drewrite.recipeArtifactCoordinates=${tool} \
```
hohwille, line 21: According to the help, the `tool` is optional.
Shouldn't we omit this entire line if `tool` is not present?

Can one of the arguments contain spaces? I assume not, otherwise you would need to put single-quotes around each arg-line.

hohwille (changes requested):
@alfeilex Thanks for your PR. Great that you have been able to figure out how to solve #991 without manipulating the local pom file. Looks good and functional already.

However, I have left two review comments. Can you please have a look and consider according rework.

scripts/src/main/resources/scripts/command/rewrite
alfeilex: The arguments can not contain spaces since the syntax is always something like `groupId:artifactId:version`. 

We can omit the line with `tool` but there is no error when the second parameter with `tool` is empty.

documentation/rewrite.asciidoc
```diff
@@ -0,0 +1,35 @@
+:toc:
+toc::[]
+
+= rewrite
```
alfeilex, line 4: Should we explain this tool more detailed? The Getting Started page gives already a good introduction to the use of openrewrite

scripts/src/main/resources/scripts/command/rewrite
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+# shellcheck source=scripts/functions
+source "$(dirname "${0}")"/../functions
+
+function doRun() {
+    local recipe="${1}"
+    local tool="${2}"
+
+    doRunCommand "mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
```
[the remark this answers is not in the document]
hohwille: Nicely solved now. Thanks 👍

scripts/src/main/resources/scripts/command/rewrite
hohwille: OK, fine. Thanks for clarification.

documentation/rewrite.asciidoc
hohwille: Your link to the getting started page is fine and should be sufficient. Documentation can always be improved but this is a great starting point.

hohwille (approved):
@alfeilex thanks for your rework. Looks perfect to me 👍 
Ready for merge.

documentation/rewrite.asciidoc
```diff
@@ -0,0 +1,35 @@
...
+
+The rewrite commandlet allows the use of https://docs.openrewrite.org/[Openrewrite], a Maven based tool that can be used to refactor code with recipes for multiple languages and tools. Before using this tool, we recommend that you read the https://docs.openrewrite.org/running-recipes/getting-started[Getting-Started] page.
+
+== Usage
+The commandlet excepts the name of a recipe such as
```
timtebeek, line 9:
```suggestion
The commandlet accepts the name of a recipe such as
```

hohwille: @timtebeek thanks for your feedback to improve. I merged roughly at the same moment.
However, I just included your suggestion so it is not lost.

Issue #1022, opened 2022-12-16
suppress confusing windows process result messages like The operation completed successfully.


hohwille:
As discussed in #1020:
We get magic result messages from windows for every (sub-)process like here:
```
Completed
The operation completed successfully.
The operation completed successfully.
ERROR: Error opening the file. There may be a disk or file system error.
ERROR: Error opening the file. There may be a disk or file system error.
ERROR: Error opening the file. There may be a disk or file system error.
Setup of devonfw-ide completed
```

The message `The operation completed successfully.` is nowhere in our code. Windows automatically prints this out if we return from a sub-process. I have no idea how to prevent this yet but this story is about finding a solution for it.
Even more strange there are 5 of such messages and if I run `setup.exe` not via double-clicking but from a `CMD` tab in Windows Terminal the last three messages turn into `ERROR: Error opening the file. There may be a disk or file system error.` but I have no clue why and what this really means or what is wrong here.
We should continue to dig deeper and find answers and solutions to such phenomena.

PR #1048, opened 2023-01-31
#1022: suppress confusing windows process result messages like The operation completed successfully.

Amueller36:
Added redirection to >nul 2>&1.
#1022


Amueller36:
Tested, in the screenshot you can see that the 5 messages are not appearing anymore.
image
And here you can see that the setup fails properly as intended
image

hohwille (approved):
@Amueller36 thanks for tracing this down and fixing it 👍 
Ready for merge.

Issue #1052, opened 2023-02-02
Fix leading spaces on macOS


alfeilex:
~~There is an error with integration-test-custom-tools in the nightly integration-test in Windows and Ubuntu.~~ 

The error did not relies on the integration-test, but on `wc-c` command which produces leading spaces on some macOS systems. 
- [x] Fix leading spaces in doReplaceExtractedSkipSingleFolder()

alfeilex:
The error is related to the changed path after installing the custom tools. If we remove the `/service/` from the path, the condition in L16 is true and the exit code is 0.

https://github.com/devonfw/ide/blob/60055a97c50082c884d85d66b2345e4f21c32f71/scripts/src/test/bash/integration-test-custom-tools#L7-L17

Since the errors only occurred on Windows and Ubuntu, we need to check how the path of the custom tools is for MacOS.

alfeilex:
The error was caused by `wc-c` command, because it produces extra whitespaces on some macos systems.
https://github.com/devonfw/ide/blob/0a438d9157e7109b500240ab2edbfc3de8d25951/scripts/src/main/resources/scripts/functions#L839

Solutions: https://unix.stackexchange.com/questions/205906/extra-space-with-counted-line-number/205910#205910

PR #1053, opened 2023-02-02
#1052: Fix leading spaces on macOS

alfeilex:
This PR fix the behaviour of `doReplaceExtractedSkipSingleFolder` on MacOS system without M1 chips. The reason of the bug was that `wc-w` may produce leading spaces,

Related issue: #1052

### DIFF
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -836,7 +836,7 @@ function doReplaceExtractedSkipSingleFolder() {
   if [ "${#*}" = 1 ] && [ "${1%.app}" = "${1}" ] && [ -d "${1}" ]
   then
     files="$(find "${dir}" -maxdepth 1 -mindepth 1 -exec basename {} \;)"
-    anz_files="$(echo "${files}" | wc -w)"
+    anz_files="$(echo "${files}" | awk 'END{print NR}')"
     while [ "${anz_files}" == "1" ]
     do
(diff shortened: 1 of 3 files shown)


scripts/src/main/resources/scripts/functions
```diff
@@ -836,7 +836,7 @@ function doReplaceExtractedSkipSingleFolder() {
   if [ "${#*}" = 1 ] && [ "${1%.app}" = "${1}" ] && [ -d "${1}" ]
   then
     files="$(find "${dir}" -maxdepth 1 -mindepth 1 -exec basename {} \;)"
-    anz_files="$(echo "${files}" | wc -w)"
```
alfeilex, line 839: It seems that the wc implementation on some MacOS systems produces leading spaces. 
Example: https://github.com/alfeilex/ide/actions/runs/4108659467/jobs/7089608184#step:5:436

I have found some solutions to fix this. I chose the `awk` option, but you can see other alternatives here:
https://unix.stackexchange.com/questions/205906/extra-space-with-counted-line-number/205910#205910

hohwille (approved):
@alfeilex thanks for tracing this down and fixing it 👍 
Ready for merge.

scripts/src/main/resources/scripts/functions
hohwille: Thanks for the explanation.
Indeed we should avoid calling external processes like `find`, `awk` or `wc` at all for such things, but I will create another PR for this. I resolve this conversation to allow merging.

Issue #1044, opened 2023-01-26
eclipse commandlet cleanup


hohwille:
Cite from #904 :
> Only thing I havent done yet is
"Instead of writing the entire log and then filtering it with sed, we should directly use a pipe for that as already done in line 67 (IMHO we should even create a function for the invocation of eclipse with the -application option, etc. including to write a log file and grepping for an expected message to check if this was successful or not instead of copy&paste)"
this because I am not sure how to do it

There are other aspects, I did not mention in the review to get this finally merged like wrong indentation, etc.
So as a nice to have this story can do some polishing, reduce redundancies, etc.

PR #1058, opened 2023-02-14
#1044: eclipse commandlet cleanup

Amueller36:
#1044 
Switched all variables to snake_case
Created Functions for the invocation of eclipse with specified arguments and extracting debug info.
Wrote functions for evaluating the debug info and log accordingly


scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -124,72 +126,104 @@ function doCheckPluginsInstalled() {
...
-  # Overwrite file with lines containing !MESSAGE
-  sed -i -n '/!MESSAGE/ p' "${debug_info_file}"
+  if [ $exit_code -ne 0 ];
+  then
+    doError "Failed to invoke Eclipse with arguments: ${eclipse_args[*]}"
+    echo ""
+    return 1
+  else
+  local debug_info
+  debug_info=$(<"${debug_info_file}")
```
hohwille, line 196: why do we first have to write the output into a file and then read the file into a variable and echo that?
IMHO this does not make sense anymore. Why don't we just remove `> "${debug_info_file}"` etc. and let the output go its regular way?

hohwille (commented):
@Amueller36 thanks for your PR. Nice rework and as it seems you also fixed at least one bug 👍 
Can you check my comment and see if I am not mistaken?

scripts/src/main/resources/scripts/command/eclipse
Amueller36: Done

hohwille (approved):
@Amueller36 thanks for this PR with the improvements. Looks good. I will merge via commandline and also do some testing when I have some time.

hohwille:
Tested successfully and will now merge.

Issue #1035, opened 2023-01-12
Github action to update ide-urls


hohwille:
For our epic #941 we have decided to do create a new process to update the `ide-urls` repository every night automatically.
This story is about implementing the github action that triggers this process.
It needs the following steps:

* [x] clone the default branch of `ide-urls` repository
* [x] build `ide-updater` project of `devonfw-ide` (this repo) using maven.
* [x] run the `ide-updater` on the resulting cloned git repo (may also be done with exec-maven-plugin)
* [x] commit the changes
* [x] push the comit

Blocked by dependencies:
* #1086

PR #1059, opened 2023-02-15
#1035: Github action to update ide-urls

alfeilex:
This PR creates a GitHub Action to update urls in https://github.com/devonfw/ide-urls repository.

Related issue: #1035


.github/workflows/update-urls.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout ide
+        uses: actions/checkout@v3
+      - name: Checkout ide-urls
+        uses: actions/checkout@v3
+        with:
+          repository: devonfw/ide-urls
+          path: ide-urls
+          token: ${{ secrets.ACTION_PUSH_TOKEN }}
```
alfeilex, line 18: This token needs repo scope permissions. If there is already a token in `devonfw/ide` for this, we can update.

alfeilex:
See results in forked ide-urls repo:

https://github.com/alfeilex/ide-urls

hohwille (commented):
@alfeilex thanks for your PR and this workflow. Nice work 👍 
As the Java code executed from the workflow is not yet merged, we have to wait for this before we can merge and use this workflow. Thanks also for the hint about the token.

tobka777 (approved):
Looks good. We should merge it and then test if it works incl. the tokens.

.github/workflows/update-urls.yml
```diff
@@ -0,0 +1,31 @@
...
+          mvn install
+          mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.url.UpdateInitiator" -Dexec.args="../ide-urls"
+      - name: Commit and push to ide-urls
+        run: |
+          cd ide-urls
+          git add .
+          git config --global user.name ${{ secrets.BUILD_USER }}
+          git config --global user.email ${{ secrets.BUILD_USER_EMAIL }}
+          git commit -m "Update urls"
+          git push
```
hohwille, line 31: as discussed in PR #1093 we need to solve the push problem as this will also not work as is.

hohwille:
#1086 needs to be implemented and merged before merging this PR to avoid clutter in git history.

.github/workflows/update-urls.yml
hohwille: We do not have such token yet. Do you have a link to any github instructions about this?
I know how to create secrets but should I add my personal github access token there or what value should this have?

alfeilex: https://docs.github.com/de/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#erstellen-eines-personal-access-token-classic

I added my own personal access token there. That was the only way I could find to make it work.

hohwille: OK, I have created such token as fine-grained org token (was more complicated than I expected) and added it as according secret to the devonfw github org (even though so far the token is only limited to ide specific repos). Lets first test this approach. Also the token requires an expiry so the github action will stop working on 2.4.2025.

.github/workflows/update-urls.yml
hohwille: Lets give it a try with the push token...

Issue #1062, opened 2023-02-28
update devon4j to 2023.01.001


hohwille:
Same procedure as after every devon4j release - see #522 
```
Expecting:
 <["<?xml version="1.0" encoding="UTF-8"?>",
    "<archetype-catalog xsi:schemaLocation="http://maven.apache.org/plugins/maven-archetype-plugin/archetype-catalog/1.0.0 http://maven.apache.org/xsd/archetype-catalog-1.0.0.xsd"",
    "    xmlns="http://maven.apache.org/plugins/maven-archetype-plugin/archetype-catalog/1.0.0"",
    "    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">",
    "  <archetypes>",
    "    <archetype>",
    "      <groupId>com.devonfw.java.templates</groupId>",
    "      <artifactId>devon4j-template-server</artifactId>",
    "      <version>2022.12.001</version>",
    "      <repository></repository>",
    "    </archetype>",
    "  </archetypes>",
    "</archetype-catalog>"]>
to contain:
 <["      <version>2023.01.001</version>"]>
but could not find:
 <["      <version>2023.01.001</version>"]>
```

I already updated ide-settings but we need to do the following in this story:
* [x] update settings submodule to latest commit on master
* [x] update migrations

All the above tasks can be checked from the diffs done for #522.

PR #1063, opened 2023-03-02
#1062: Update devon4j to 2023.01.001

alfeilex:
Update devon4j to 2023.01.001 | submodules and migrations

Related issue: https://github.com/devonfw/ide/issues/1062


hohwille (approved):
@alfeilex thanks for the PR.
We could actually include the version updates for 2023.01.001 to the migrations but as this migration feature is rarely used and we will remove it anyhow this is sufficient. 
Ready for merge👍

Issue #1066, opened 2023-03-08
devon.bat stopped adding PATH


hohwille:
On windows when using devon.bat (e.g. typing devon in CMD) it stopped setting the PATH properly.
```
D:\projects\mmm>devon
devonfw-ide environment variables have been set for D:\projects\mmm in workspace main

D:\projects\mmm>mvn -v
'mvn' is not recognized as an internal or external command,
operable program or batch file.

D:\projects\mmm>dir software
 Volume in drive D is DATADRIVE1
 Volume Serial Number is 120F-6414

 Directory of D:\projects\mmm\software

07.03.2023  12:43    <DIR>          .
07.03.2023  12:43    <DIR>          ..
06.02.2023  18:35    <DIR>          eclipse
30.11.2022  22:33    <DIR>          java
30.11.2022  18:36    <DIR>          maven
07.03.2023  12:43    <DIR>          node
09.01.2023  16:35    <DIR>          vscode
09.01.2023  16:38    <DIR>          vscode-extensions
               0 File(s)              0 bytes
               8 Dir(s)  262.994.305.024 bytes free

D:\projects\mmm>
```

This is a severe bug and it breaks the entire *.bat support of devonfw-ide.

hohwille:
The bug was already in `2022.11.002` but on in `2022.11.001`.
OMG, I have been so used to git-bash that I never used CMD since such a long time and never noticed the bug in my daily work!

hohwille:
It seems that PR #975 broke it and we never properly tested that and also our nightly windows tests do not reveal the error.

hohwille:
As a first analysis, I hacked this block in `devon.bat` as following:
```
if exist scripts\environment-project.bat (
  echo %CD%
  echo set DEVON_IDE_HOME=%CD%
  set DEVON_IDE_HOME=%CD%
  echo DEVON_IDE_HOME=%DEVON_IDE_HOME%
  echo DEVON_IDE_HOME=!DEVON_IDE_HOME!
  call scripts\environment-project.bat
  echo devonfw-ide environment variables have been set for %CD% in workspace %WORKSPACE%
  popd
  goto :eof
)
```

Result:
```
D:\projects\devonfw\workspaces>devon
D:\projects\devonfw
set DEVON_IDE_HOME=D:\projects\devonfw
DEVON_IDE_HOME=
DEVON_IDE_HOME=D:\projects\devonfw
devonfw-ide environment variables have been set for D:\projects\devonfw in workspace

[...]

PR #1067, opened 2023-03-09
#1066: devon.bat stopped adding PATH

alfeilex:
This PR fixes the error caused by PR https://github.com/devonfw/ide/pull/975. 

**How the fix works:**
It removes the delayed expansion and uses an external label to fix the error provided by issue https://github.com/devonfw/ide/issues/954

Related issue: https://github.com/devonfw/ide/issues/1066


hohwille (approved):
@alfeilex Awesome. Thanks for your fast analysis and this PR with a fix. 👍 
This is really great. I already moved this from draft to ready for review and hereby approve it.
I have locally updated my devon.bat with your fixed version and tested it with several combinations. It looks good to me. I hope that #954 remains fixed.

Issue #1074, opened 2023-03-16
VS Code Plugins won't install


patrickvogel:
### Expected behavior

As a devonfw-IDE user, I want the active plugins of my ide-settings for VS code to be installed and the inactive plugins to be recommandations.

### Actual behavior

The active plugins are not installed. The inactive plugins are correctly configured as recommendations.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Configure active & inactive VS code plugins and call "devon ide setup" or "devon ide update"

### Related/Dependent Issues

https://github.com/devonfw/ide/issues/982

### Comments/Hints:

`local IFS=, ` in line 67 breaks the later `doAddPlugin` call.
https://github.com/devonfw/ide/commit/f7633e0573324917631955c3efa6b2e7aac80859#diff-49491b40b86d2ea85000a079a50079575d1435214f97f3b6b7be00fc2a42ff31R67
Idea for fix: It should be sufficient to call `local IFS=, ` just right before `doAddRecommendations`

### Affected version:

Probably since release/2023.02.001

alfeilex:
Thank you @patrickvogel for this bug report and the great solution to fix this issue. I created this PR for this https://github.com/devonfw/ide/pull/1075

PR #1075, opened 2023-03-16
#1074 :  Fix vscode plugin installation

alfeilex:
This PR fix the bug report for #1074.


hohwille (approved):
@alfeilex thanks for the fix. Looks good. 👍

Issue #1054, opened 2023-02-03
automated security warnings


hohwille:
Already devonfw-ide does a great job to automated tool installations and meanwhile supports automatic updates to recent versions.
However, IT security gets more and more important. As recent vulnerabilities like [CVE-2022-23521](https://nvd.nist.gov/vuln/detail/CVE-2022-23521) and [CVE-2022-41903](https://nvd.nist.gov/vuln/detail/CVE-2022-41903) show, it would be great if devonfw-ide could automatically warn users if he using outdated and vulnerable software. In case of the specified CVEs it would be git that is about the only tool we do not install and manage. However, during setup we could check the version and raise a fat and red warning if the version is vulnerable. Also projects can and for stability also want to pin the versions of tools to get a reliable and reproduce environment for their team. So to spin this further, it would be great if in the context of #941 we could introduce some kind of configuration file that marks versions that are vulnerable. So we could add the tool and edition for git as `git/git/security` with something like:
```
>2.39.0
```

So when we setup a tool (or for git in case we setup the ide itself) we can check the installed version and test if that version is marked as vulnerable. In that case we would render a fat and red warning to the end-user notifying him to take action:
```
******** ATTENTION ********
You are using the tool git in version 2.38.0 that has vulnerabilities. Please update the tool to state safe and secure.
For further details and recent information have a look at the following webpage:
https://github.com/devonfw/ide/blob/master/documentation/vulnerabilities.asciidoc
```

Obviously we would also create this asciidoc documentation page and explain how to update git and how to update tools maintained by devonfw-ide and how to proceed depending on your team role (developer should inform ide admin or provide PR for settings while settings admin should directly push the change and inform the team to call `devon ide update`).

To explain an advanced use-case:
Maybe a tool maintains several major versions and vulnerabilities only apply to `1.x` versions up to `1.11.0` and to `2.x` versions up to `2.0.23` we can write this to the `security` file:
```
1.0>1.11.0
2.0>2.0.23
```
This would mean that the versions `1.0`-`1.11.0` and `2.0`-`2.0.23` are considered vulnerable.
If we instead write this:
```
>1.11.0
2.0>2.0.23
```
This would mean that the versions `0`-`1.11.0` and `2.0`-`2.0.23` are considered vulnerable - so also 0.x versions are included.

hohwille:
I created issue #1106 for the remaining todos so we can close this issue now.

PR #1078, opened 2023-03-19
#1054 automated security warnings

CREITZ25:
#1054 automated security warnings

### DIFF
--- a/documentation/functions.asciidoc
+++ b/documentation/functions.asciidoc
@@ -46,6 +46,9 @@ In xref:doIsForce[force] or xref:doIsBatch[batch] mode, it will continue automat
 Passes all arguments to xref:doWarning[] to print as warning.
 Then it calls xref:doAskToContinue[].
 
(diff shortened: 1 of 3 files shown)


CREITZ25:
File https://github.com/devonfw/ide/blob/master/documentation/vulnerabilities.asciidoc is not created yet!

hohwille:
This PR is replaced by #1083
However, some changes here shall not be lost and may be recovered for future improvements after #1083 has been merged.

hohwille:
I am finally closing this PR as all information will still be preserved whenever we need them.
To summarize: Here we started to solve the problem that an important error or warning is raised during a complex process but further log-messages will "spam" it away. If the user took his promised coffee and comes back the message might already be scrolled out of sight. In our new Java based product it will be much easier to solve this problem cleanly so I am unsure if we will ever solve it in the bash world.

However, this is not going to be merged and while a PR remains open it continuously gains my attention if I have something left to do...

Issue #1076, opened 2023-03-16
Fix extraction error MacOS


alfeilex:
The integration-test for macOS throws out an error for cobigen

See error log here: https://github.com/devonfw/ide/actions/runs/4422124811/jobs/7753619946#step:5:418

A.C:

- [x] Analyse the error
- [x] Fix the error 

****UPDATE****
The error is caused by this PR https://github.com/devonfw/ide/issues/1052 
The function: https://github.com/devonfw/ide/blob/0a438d9157e7109b500240ab2edbfc3de8d25951/scripts/src/main/resources/scripts/functions#L827

does not extract directories correctly (e.g. JAVA)

alfeilex:
I can reproduce the error on remote Mac with m1

Probably something is wrong with the JAVA_HOME variable

alfeilex:
I found the source of the error: 

https://github.com/devonfw/ide/pull/1053

The change to `awk 'END{print NR}'` cause that the installed software is not extracted correctly.

alfeilex:
We updated the doReplaceExtractedSkipSingleFolder with a recursion function.

PR #1091, opened 2023-03-31
#1076: Fix doReplaceExtractedSkipSingleFolder() with recursion

alfeilex:
This PR fixes the error while extracting software on MacOS. Additional variable names of the function were changed because they did not fit to their function.

Related issues: https://github.com/devonfw/ide/issues/1076


alfeilex:
I have also changed the variables names used in the function as they don't match their usage. Maybe, we should create a new issue to check the variable names in `functions` script.

scripts/src/main/resources/scripts/functions
```diff
@@ -898,34 +898,19 @@ function doBackup() {
...
 # $4-$N: contents of $1 ($1/* so all files in the folder of $1 after globbing)
 function doReplaceExtractedSkipSingleFolder() {
-  local source_dir="${1}"
-  local target_dir="${2}"
-  local backup_dir="${3}"
-  local root_dir="${1}"
+  local source_path="${1}"
+  local target_path="${2}"
+  local backup_path="${3}"
+  local root_path="${1}"
```
hohwille, line 904: I do not see any difference between `source_path` and `root_path`.

scripts/src/main/resources/scripts/functions
```diff
@@ -898,34 +898,19 @@ function doBackup() {
...
   shift 3
-  local files
-  local anz_files
-  local dir="${1}"
-  if [ "${#*}" = 1 ] && [ "${1%.app}" = "${1}" ] && [ -d "${1}" ]
+  local path="${1}"
```
hohwille, line 906: Also `path` is the same as `source_path` so one variable instead of 3 would be sufficient and easier to read and understand.

hohwille (approved):
@alfeilex thanks for this PR with your fix. Looks functionally fine and should fix all problems👍 
I left a small review feedback regarding variables. In case you have time to quickly improve, you are happily invited to address it. I will merge this soon also as is and we can still address unimportant nice-to-have improvement later.

scripts/src/main/resources/scripts/functions
hohwille: done.

scripts/src/main/resources/scripts/functions
hohwille: done

Issue #1065, opened 2023-03-07
checksum verification of binary downloads


hohwille:
With epic #941 we are redesigning our downloads and introducing [ide-urls](https://github.com/devonfw/ide-urls/) repo.
The idea of this story is to improve security and prevent man-in-the-middle or other attacks that may lead to downloads of evil content and therefore remote-code-execution.
To archive this, we will introduce checksum files (e.g. sha or md5) for all our downloads in ide-urls.
So in addition to downloading the actual package, we would also download the corresponding checksum file.
Then we would compute the checksum locally using an according tool (e.g. `sha256sum` or `md5sum`) and compare the result with the content of the downloaded checksum file.
If they match, we continue as before, but otherwise, we would fail with an according error and a link with further instructions (AsciiDoc file with hints and link to report a security issue so we can also check if the vendor has actually changed the binary after releasing on purpose what is an anti-pattern but might also happen in rare cases).

hohwille:
For the record: Maven is already following this best practice approach as you can see [here](https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-scripts/2023.01.001/):
* There is [devonfw-ide-scripts-2023.01.001.tar.gz](https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-scripts/2023.01.001/devonfw-ide-scripts-2023.01.001.tar.gz)
* But also there are the checksum files [devonfw-ide-scripts-2023.01.001.pom.md5](https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-scripts/2023.01.001/devonfw-ide-scripts-2023.01.001.pom.md5) and [devonfw-ide-scripts-2023.01.001.pom.sha1](https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-scripts/2023.01.001/devonfw-ide-scripts-2023.01.001.pom.sha1).

I would consider `sha1` as obsolete and insecure. `sha256` is most probably a perfect choice from security point of view but we would have to consider the following:
* is the checksum tool present on every target environment (MacOS, Windows + GitBash, Linux, Github-Actions aka Linux)? I would assume for `md5sum` this is given, but we need to check this for `sha256sum`.
* how long will the signature be and how big will our repository grow just due to bigger signatures? If it is just 1kB+ we should be fine, but if we add megabytes, we should be a little bit more concerned and consider a trade-off between security and sustainability.

[...]

Issue #1086, opened 2023-03-28
Align tool names in ide-urls


hohwille:
Analog to story #1085 we want to change the `ide-urls` structure to strictly follow our conventions that if we have a tool `«tool»` then we use the same name in the following places:
* commandlet: `scripts/command/«tool»`
* software folder: `software/[extra/|macos/]«tool»`
* ide-urls folder: `ide-urls/«tool»/«edition»/«version»/`
* variables: `«TOOL»_VERSION` and `«TOOL»_EDITION`
* When running `devon «tool» «args»` you should get the same effect as when running `devon` followed by `«tool» «args»` except that if the tool is not yet installed in the first case it will be downloaded and installed and in the second case it will fail as the tool is not installed.

Within this story the new crawlers/updaters (#1009) we should align the repository structure of `ide-urls` to strictly follow the above rule.
This means that we eliminate the following excuses/exceptions:
* `mvn` vs. `maven` - as the command to run is given as `mvn` we have to use that instead of `maven` consistently.
* `node` vs. `nodejs` - as the command to run is given as `node` we have to use that instead of `nodejs` consistently.
* `sonar` vs. `sonarqube` - here we do not have any matching given command except `StartSonar.bat` or the like that is not even in the `PATH` after set by `devon` command. For simplicity, I would propose to stay with the shorter name `sonar` and rename `sonarqube` occurances to `sonar`.
* `cobigen` vs. `cobigen-cli` vs. `cli`. The command is actually `cg` and `cobigen`. We can discuss with @jan-vcapgemini but IMHO `cobigen` is more self explanatory than `cg` that could confuse new users that might never get what this is about and therefore miss the according features behind it.

Issue #1057, opened 2023-02-06
Release workflow fails to push changes


hohwille:
After #979 is resolved and fixed, I reran the release workflow and it sill fails now with this error:

```
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: error: You're not authorized to push to this branch. Visit https://docs.github.com/articles/about-protected-branches/ for more information. 2 of 2 required status checks are expected.        
To https://github.com/devonfw/ide
 ! [remote rejected] master -> master (protected branch hook declined)
```

Full log:
https://github.com/devonfw/ide/actions/runs/4102046425/jobs/7074531639

hohwille:
So the actual release itself was build and arrived in maven central:
https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-scripts/2023.01.001/

Unlike to what I expected also the doc artifact was released and staged automatically (I did not manually release it in OSSRH):
https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-doc/2023.01.001/

However, we now to not have the version changes comitted and are lacking the tag for the release.
Also no github release has been crafted, what is obvious as this should happen after the push (https://github.com/devonfw/ide/blob/master/.github/workflows/release.yml)
I can try to manually fake these changes but we still have to investigate how to fix this.
Is the workflow only missing out to configure git properly or are we failing due to a lack of permissions and our branch protection?

hohwille:
I found this one:
https://github.com/stefanzweifel/git-auto-commit-action/issues/71

So we could add `---force` to the `git push` commands and I can allow force pushes to bypass the branch protection.
As we usually do not give out push permissions but only triage permissions that should be no big deal...

maybeec:
@hohwille, should be fixed by making use of https://github.com/devonfw-actions/java-maven-setup or copying the respective section out: https://github.com/devonfw-actions/java-maven-setup/blob/main/action.yml#L117

hohwille:
Thanks @maybeec - even better.
we already have this:
```
        git config --global user.email ${{ secrets.BUILD_USER_EMAIL }}
        git config --global user.name ${{ secrets.BUILD_USER }}
```

So all we need is:
```
git remote set-url origin github.com/devonfw/ide.git
```

hohwille:
Fixed. I am closing in optimistic assumption that it finally works now. Otherwise, we can reopen or create a new issue.

hohwille:
I have to reopen.
[...]

PR #1093, opened 2023-04-03
#1065: Add SHA-256 checksum generator to url-updater

alfeilex:
This PR adds a checksum generator that calculates the checksum of the download files covered by download urls of each tool and its OS. The hash algorithm used is SHA256.

Example:
Windows_x64.urls will have a Windows_x64.urls.sha256


url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
```diff
@@ -0,0 +1,129 @@
...
+        try {
+            HttpRequest request1 = HttpRequest.newBuilder()
+                    .uri(URI.create(url))
+                    .GET()
+                    .build();
+            return client.send(request1, HttpResponse.BodyHandlers.ofInputStream()).body();
+        } catch (IOException | InterruptedException exception) {
+            throw new IllegalStateException("Failed to retrieve response body from url: " + url, exception);
+        } catch (IllegalArgumentException e) {
+            System.out.println("Error while getting response body from url {}\", url, e");
```
hohwille, line 54: Use Logger and ensure the stacktrace does not get lost

.github/workflows/checksum-test.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  updateURLS:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout ide
+        uses: actions/checkout@v3
+      - name: Checkout ide-urls
+        uses: actions/checkout@v3
+        with:
+          repository: alfeilex/ide-urls-checksum
```
hohwille, line 13: we should complete stories like #1086 and then switch to the real repo

.github/workflows/checksum-test.yml
```diff
@@ -0,0 +1,29 @@
...
+          mvn exec:java -Dexec.mainClass="com.devonfw.tools.ide.url.UpdateInitiator" -Dexec.args="../ide-urls"
+      - name: Commit and push to ide-urls
+        run: |
+          cd ide-urls
+          git checkout -b with-checksum-testing
+          git add .
+          git config --global user.name ${{ secrets.BUILD_USER }}
+          git config --global user.email ${{ secrets.BUILD_USER_EMAIL }}
+          git commit -m "Checksum testing"
+          git push --set-upstream origin with-checksum-testing
```
hohwille, line 29: IMHO we have the same problem with pushing in github actions like in #1057. We need to find a solution for this problem and apply it to both actions.

.github/workflows/checksum-test.yml
```diff
@@ -0,0 +1,29 @@
+name: Checksum Testing
```
hohwille, line 1: why is this called `checksum-test.yml`?
It looks like the regular github action for updating `ide-urls`.
Am I missing something?

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
```diff
@@ -0,0 +1,129 @@
...
+                md.update(buffer, 0, bytesRead);
+            }
+            inputStream.close();
+
+            byte[] digestBytes = md.digest();
+            String checksum = toHexString(digestBytes);
+            System.out.println(checksum);
+            return checksum;
+        } catch (IOException e) {
+            throw new IllegalStateException("Failed to read input stream " + e);
```
hohwille, line 80: You are loosing the real exception with its valueable stacktrace and causes:

https://github.com/devonfw/devon4j/blob/master/documentation/coding-conventions.asciidoc#catching-and-handling-exceptions

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
```diff
@@ -0,0 +1,129 @@
...
+        } catch (NoSuchAlgorithmException e) {
+            System.out.println("No such hash algorithm " + hashAlgorithm);
```
hohwille, line 82: Also throw an `IllegalStateException` instead of returning `null`. In this case the JDK is broken, and we should react and fix the problem rather than trying to continue...

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
```diff
@@ -0,0 +1,129 @@
...
+import java.net.http.HttpResponse;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+public class UrlChecksum extends AbstractUrlFile {
+
+    final HttpClient client = HttpClient.newBuilder().build();
```
hohwille, line 18: should be `private`.
IMHO we should keep logic like `HttpClient` out of the "model" (I would keep it in `AbstractCrawler` that already has an `HttpClient` instance).

hohwille (changes requested):
@alfeilex thanks for your PR. Your way to compute the hashes looks fine now. 👍 
I left some review comments for improvements. Can you have a look?

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
alfeilex: Done

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
```diff
@@ -0,0 +1,129 @@
...
+    final String hashAlgorithm = "SHA-256";
```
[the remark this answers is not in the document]
alfeilex: Done

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
alfeilex: done

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
alfeilex: done

url-updater/src/main/java/com/devonfw/tools/ide/url/folderhandling/UrlChecksum.java
hohwille: I resolved this myself and removed `System.out.println` statement as well. Now resolved.

hohwille (approved):
@alfeilex thanks for rework. 👍 
I addressed the last open review comment so now ready for merge.

Issue #1073, opened 2023-03-14
Improve integration tests on windows to test devonfw.bat properly


hohwille:
We have been facing the following bugs:
* #1066 
* #1043
* #1068
* #1071

All of those are just related to bugs in `devon.bat` and broke the setup of devonfw-ide.
Actually the idea of #157 was to prevent such problems.
As this did not happen, we should use this story, to extend our integration tests on windows to do the following:
* [x] do a full initial setup by executing `setup.bat` on the windows machine. This should have revealed #1068. 
* [x] After the `setup.bat` has completed, we should check for exit code 0 (ERRORLEVEL) and further check the users %PATH% variable from the registry once more to prevent #1071 and #954. Maybe for this test we can read the %PATH% from registry before calling `setup.bat` and after and then ensure that by adding the users scripts directory to the first result should be equal to the second result.
* [x] We should somehow unset the environment variables (maybe by running the entire setup in a "subshell CMD") and then check that according software tools such as "mvn" or "java" are not available. Then we call `devon.bat` command and expect that now these tools are available. This should prevent #1066.
* [x] Also during the latter test, we can check from some environment variable from devon.properties not to be present before calling `devon.bat` and then to be present after calling it. This should prevent #1043

Besides this story, we have to do the following checks in every daily meeting:
* are all nightly tests green? if not define who is analyzing the problem on the same day and also reschedule a release if it was planned for that day (releases should then always prepared one day and only be performed the day after if the nightly tests are green).
* are there new (bug) issues in our issue-tracker (https://github.com/devonfw/ide/issues?q=is%3Aopen+is%3Aissue+label%3Abug)? If so, let us have a look with the entire team.

maybeec:
I feel you could easily reuse what is already implemented in the tutorial compiler and set it up within the devonfw ide repository:

test spec: https://github.com/devonfw-tutorials/tutorials/blob/main/devonfw-ide/index.asciidoc
workflow spec: https://github.com/devonfw-tutorials/tutorials/blob/main/.github/workflows/buildMain.yml
recent execution: https://github.com/devonfw-tutorials/tutorials/actions/runs/4419868086/jobs/7748856469
run-command: https://github.com/devonfw-tutorials/tutorial-compiler/blob/main/runners/console/index.ts#L45
assertion: https://github.com/devonfw-tutorials/tutorial-compiler/blob/main/runners/console/index.ts#L547

maybeec:
[...]

PR #1094, opened 2023-04-08
 #1073: Add testcase for windows setup.bat and devon.bat + fix colored output

alfeilex:
This PR extend the integration-tests and add a test for `setup.bat` and `devon.bat`.

Further, `-Dstyle.color=always` is added to mvn build command in the integration-tests to enable colored output. The tests results are now displayed summarized in the end with related colors.


hohwille (approved):
@alfeilex Thanks for this PR. Great that you found the magic trick to get colored output also on github. Also awesome that you improved the test output and reporting.
IMHO it would have been simpler to use a variable instead of a file to collect the messages but that does not matter. All great and ready for merge...

Issue #1087, opened 2023-03-28
Create migration script to switch from mirrors to urls


hohwille:
For epic #941 we also need to adapt the devonfw-ide scripts itself such that it uses `ide-urls` instead of `ide-mirrors`.
This story is about writing the migration script (see #971) for the upcoming release `2023.04.001` with the following aspects:
* [ ] delete the `mirrors` folder.
* [ ] clone the `ide-urls`repository into folder `urls`.
* [ ] adapt `devon.properties` from the legacy config to the new one as discussed in the epic #941, etc. Details are given below.

For the replacements in `devon.properties` use the following rules:
* Variable `INTELLIJ_EDITION_TYPE` has to be renamed to `INTELLIJ_EDITION`.
* Variable `ECLIPSE_EDITION_TYPE` has to be renamed to `ECLIPSE_EDITION`.
* Variable `MAVEN_VERSION` has to be renamed to `MVN_VERSION`.
* Variable `SONARQUBE_VERSION` has to be renamed to `SONAR_VERSION`.
* Value of intellij edition `C` has to be changed to `intellij` (default edition is community edition).
* Value of intellij edition `U` has to be changed to `ultimate`.
* Value of eclipse edition `java` has to be changed to `eclipse` (default edition is java edition).
* Value of docker edition has to be set to `docker` in case it is `DOCKERDESKTOP ` or `docker` and in any other case to `rancher`.

The above replacements shall happen in all affected `devon.properties` (see [configuration](https://github.com/devonfw/ide/blob/master/documentation/configuration.asciidoc) to get the list of them but you can ignore 8.). In case a replacement is happening in `~/devon.properties` you should raise a end-user message with confirmation that he needs to update all his devonfw-ide installations to the new version to avoid breaking changes.

Please also update the edition documentation in the following places:
* https://github.com/devonfw/ide/blob/master/documentation/intellij.asciidoc
* https://github.com/devonfw/ide/blob/master/documentation/eclipse.asciidoc
* https://github.com/devonfw/ide/blob/master/documentation/docker.asciidoc#docker-desktop
* https://github.com/devonfw/ide/blob/master/documentation/variables.asciidoc - add  `«TOOL»_EDITION` next to  `«TOOL»_VERSION` variable in this documentation.

PR #1123, opened 2023-04-25
#1087 Create migration script to switch from mirrors to urls

CREITZ25:
#1087 Create migration script to switch from mirrors to urls

### DIFF
--- a/scripts/src/main/resources/scripts/migrations/2023.05.001
+++ b/scripts/src/main/resources/scripts/migrations/2023.05.001
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+source "$(dirname "${0}")"/../functions
+
+function doAdaptConfigurationFiles() {
+  configuration_file="${1}"
+  cat "${configuration_file}" | awk '
+    /MAVEN_HOME/                  {gsub("MAVEN_HOME","MVN_HOME")}
+    /SONARQUBE_HOME/              {gsub("SONARQUBE_HOME","SONAR_HOME")}
+    /INTELLIJ_EDITION_TYPE/       {gsub("INTELLIJ_EDITION_TYPE","INTELLIJ_EDITION")}
+    /INTELLIJ_EDITION/, /c/       {gsub("c","INTELLIJ")}
+    /INTELLIJ_EDITION/, /u/       {gsub("u","ULTIMATE")}
+    /ECLIPSE_EDITION_TYPE/        {gsub("ECLIPSE_EDITION_TYPE","ECLIPSE_EDITION")}
+    /ECLIPSE_EDITION/, /java/     {gsub("java","ECLIPSE")}
+    /MAVEN_VERSION/               {gsub("MAVEN_VERSION","MVN_VERSION")}
+    /SONARQUBE_VERSION/           {gsub("SONARQUBE_VERSION","SONAR_VERSION")}
+    /DOCKERDESKTOP/               {gsub("DOCKERDESKTOP","DOCKER")}
+    /RANGER-DESKTOP/              {gsub("RANGER-DESKTOP","RANCHER")}        
+    {print $0}
+  ' > "${configuration_file}".tmp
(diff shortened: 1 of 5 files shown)


scripts/src/main/resources/scripts/migrations/2023.05.001
```diff
@@ -0,0 +1,51 @@
...
+    /ECLIPSE_EDITION_TYPE/        {gsub("ECLIPSE_EDITION_TYPE","ECLIPSE_EDITION")}
+    /ECLIPSE_EDITION/, /java/     {gsub("java","eclipse")}
+    /DOCKERDESKTOP/               {gsub("DOCKERDESKTOP","docker")}
+    {print $0}
+  ' > "${configuration_file}".tmp
+  mv "${configuration_file}".tmp "${configuration_file}"
+  if [ "${configuration_file}" = "${HOME}/devon.properties" ]
+  then
+    doWarning "Due to the changes in the devon.properties file in the user home directory, it is necessary to update all devonfw IDE installations to avoid errors."
+  fi
```
hohwille, line 27: actually the warning should only be raised, if something has changed in ~/devon.properties.

scripts/src/main/resources/scripts/migrations/2023.05.001
```diff
@@ -0,0 +1,51 @@
...
+then
+  git clone https://github.com/devonfw/ide-urls.git urls
+fi
+[ -d "${DEVON_IDE_HOME}"/mirrors ] && rm -rf "${DEVON_IDE_HOME}"/mirrors
+
+doAdaptConfigurationFiles "${HOME}"/devon.properties
+doAdaptConfigurationFiles "${DEVON_IDE_HOME}"/scripts/devon.properties
+doAdaptConfigurationFiles "${DEVON_IDE_HOME}"/settings/devon.properties
+doAdaptConfigurationFiles "${DEVON_IDE_HOME}"/conf/devon.properties
+doAdaptConfigurationFiles "${DEVON_IDE_HOME}"/settings/devon/conf/devon.properties
```
hohwille, line 41: great that you also figured this file that acts as the template.

scripts/src/main/resources/scripts/migrations/2023.05.001
```diff
@@ -0,0 +1,51 @@
...
+
+find "${DEVON_IDE_HOME}"/workspaces/ -mindepth 2 -maxdepth 2 -name devon.properties -print | while read foundfile
+do
+  doAdaptConfigurationFiles "${foundfile}"
+done
+
+
+
+
```
hohwille, line 50: multiple empty lines that can be removed

scripts/src/main/resources/scripts/migrations/2023.05.001
hohwille: Does `awk` report via exit code if it did any changes? Otherwise we can diff the tmp file with the original before overwriting it...

CREITZ25: @hohwille ~/devon.properties is the same as ${HOME}/devon.properties. But I change it to avoid confusions.

Now I start a diff command and move the file in case of changes

hohwille: Not entirely. You need to move this `if` inside the other `if` to ensure that the user does not get the warning for no reason.

CREITZ25: I don't know why I didn't see that. I have to continue with my studies.

CREITZ25 (commented):
All done

CREITZ25 (commented):
done

CREITZ25 (commented):
done

CREITZ25 (commented):
done

CREITZ25 (commented):
done

CREITZ25 (commented):
done

Issue #1118, opened 2023-04-24
ability to configure and setup milestone versions of Eclipse


hohwille:
As a devonfw-ide admin and user, I want to be able to test intellij milestone releases so that I can prepare the official setup for the next upcoming release, test new features as well as test and report bugs to improve the quality of an official eclipse release.

Currently only Release-Versions of Eclipse are supported and this is hard-coded:
https://github.com/devonfw/ide/blob/c3d56da8a90b950be0f9acfecd0e5f6cc651d536/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/eclipse/EclipseCrawler.java

So for proper support, we also need to discover milestone versions in `Updater` and build the URL dependent if the version is a milestone or an official release.

alfeilex:
How should we distinguish between release-types? Should we rename other releases that are not `R`, e.g. `2023-06-<release-type>`

alfeilex:
I was able to get `2023-06 M1` version in https://www.eclipse.org/downloads/packages/release 

If we would like to change the name in the final directory, we have to update the `version` already in object `UrlVersion` .

PR #1126, opened 2023-04-27
#1118: Add milestone versions of Eclipse

alfeilex:
This PR adds milestone versions of eclipse e.g. `2023-06-M1`

Related issue: #1118


hohwille (approved):
@alfeilex thanks for this PR. The way we find the versions still seems odd (from HTML with space in the version name) but that is a different story. Your fix looks good to me 👍 
Ready for merge.

Issue #1114, opened 2023-04-24
Separate eclipse plugins and keep software installation pristine


hohwille:
As a devonfw-ide maintainer and product-owner, I want to keep the installation of the `Eclipse` IDE product pristine and have plugins and their updates in a separate folder so that the `Eclipse` installation itself remains "read-only" and pristine and can be shared across multiple devonfw-ide instances without side-effects between projects.

We already have managed to archive this goal for vscode and intellij so Eclipse is the only tool missing to complete the goal (see also bug #1082 that otherwise cannot be cleanly resolved).

One option could be to set this:
```
-Dorg.eclipse.equinox.p2.reconciler.dropins.directory=«my-plugins-folder»
```

However, this will only cause Eclipse to load (additional) plugins/features from there while still installing new plugins/features into the `Eclipse` product itself that therefore manipulates itself.

Another option to try could be:
```
-configuration «my-plugin-folder»
```

Maybe there are even other options so first step would be to test, evaluate and analyse this before even creating a PR.
The goal would be that plugins installed via commandline AND plugins installed within Eclipse itself (ideally including via Eclipse update functionality of its own internal plugins shipped with the product) would go to the newly configured `«my-plugin-folder»` that could be `${DEVON_IDE_HOME}/software/eclipse-plugins` to follow current conventions. But also we could revisit this and consider moving plugins from `software` to their own top-level `plugins` folder as they are something different to the rest we put in `software`.

alfeilex:
I tried the `-configuration` option and it works to install plugins in an external location, either with command or via a GUI. 

I mentioned that the plugins do not show up in the Eclipse GUI, but after searching for the installed plugin in the marketplace, the plugin shown as "installed" marked.

alfeilex:
I have a problem uninstalling plugins via `devon eclipse remove-plugin <id>`. There is no uninstall option to specify the location of installed plugins.

alfeilex:
> I have a problem uninstalling plugins via `devon eclipse remove-plugin <id>`. There is no uninstall option to specify the location of installed plugins.

I could fix this problem by https://github.com/devonfw/ide/pull/1135/commits/7fcc663cee6085a22b8181b5c17d1b73322238e1

alfeilex:
Some tests need to be done to check shared configuration etc.

Issue #609, opened 2021-10-04
Import of Non-Maven Projects into Eclipse


andreasrichtercg:
**As a** Developer/Architect, 
**I want to** be able to import _non-maven projects_ into eclipse automatically based on projects settings using `devon ide update projects`
**so that** I don't have to import them manually

hohwille:
@andreasrichtercg - I fully agree that this would be a great feature. However, eclipse is a big mess. Making the maven import work was already a very long way and still does not work fully reliable. So just for expectation management - without further contribution you should not expect this feature in the next half year as we have many issues with higher priority...
Anyhow, everybody is very much invited to contribute and make devonfw evolve faster and better :)

PR #1135, opened 2023-04-28
#1114: Separate eclipse plugins and keep software installation pristine

alfeilex:
This PR implements the solution of installing Eclipse plugins in an external location to keep them in a single place. The plugins are installed in `software/eclipse-plugins`. Further, there is  a new wrapper for the start of eclipse to avoid redundancies and keep the cli arguments in one central place.

Related PR #1114


scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -76,7 +76,7 @@ function doInstallEclipsePlugin() {
   local pluginDependencies
   pluginDependencies=$(doCheckPluginsInstalled "${plugin_id}")
   local debugInfo
-  debugInfo=$(doInvokeEclipse org.eclipse.equinox.p2.director -repository "${plugin_url}" -installIU "${plugin_id}")
+  debugInfo=$(doInvokeEclipse org.eclipse.equinox.p2.director -repository "${plugin_url}" -installIU "${plugin_id}" -configuration "${DEVON_IDE_HOME}/software/eclipse-plugins/configuration/")
```
hohwille, line 79: you have already added `-configuration` to `doInvokeEclipse` so here you should not add it twice.

hohwille (changes requested):
@alfeilex thanks for your PR. Great that it generally works and you could analyse, implement and test this. 👍 
```
function doCheckPluginsInstalled() {
  local plugin_directory="${DEVON_IDE_HOME}/software/eclipse/plugins" 
```
This does not work anymore with your change as you also would have to check the new `eclipse-plugins` folder as well.
I also would question why we have this function `doCheckPluginsInstalled` after all.
IMHO there was a problem, that e.g. `tm-terminal` was a plugin that we installed additionally but Eclipse was changed such that later releases already included this plugin and our installation process failed. Actually Eclipse itself is stupid and should instead be idempotent. If I say "Eclipse, please install plugin tm-terminal" and `tm-terminal` is already installed, then Eclipse should say "OK (already installed)" instead of raising errors... However, we have no lobby to change Eclipse itself...
Further, we are writing files `.devon.«plugin-id».installed` files into the `software/eclipse` folder. Also this will change the product that we want to leave pristine. If we later want to make use of this story and use symlinks to share the same eclipse installation folder in multiple devonfw-ide instances and this will break as IDE1 will write according files and IDE2 things that those are already installed. Sorry, that all this was unclear in the story description of #1114 but I also forgot all these "hacks".
If we can easily derive which file or folder in `software/eclipse-plugins` will correspond to «plugin-id» we can skip writing these stupid 0 byte indicator files. Otherwise, we would now have to write them to ´software/eclipse-plugins`.
Later, regular software will turn into symlinks but not the plugins folders what is why I would also like to keep them out of `software` and instead have `plugins/eclipse`, `plugins/intellij`, and `plugins/vscode` anlongside with `software/eclipse`, `software/intellij`, and `software/vscode` - but that is a different story. I am only explaining it so you can catch a glimpse of the big picture how all these small stories should form a vision for a future devonfw-ide that is much better and makes sense :)

hohwille:
For the record:
> curl: (92) HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
Failed to download https://nodejs.org/dist/v18.12.1/node-v18.12.1-linux-x64.tar.gz with exit code 92

This build error seems to be a technical problem on github and not related to your change changes.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -113,6 +113,7 @@ function doCheckSuccessOfPluginInstall(){
 # return: dependencies that are already installed
 function doCheckPluginsInstalled() {
```
alfeilex, line 114: Does this check really work? I can continuously add the same plugin via `add-plugin` and get the same output that the plugin is successfully installed. How does the uninstallation process of eclipse work? It seems that the plugin is not removed in `plugins` directory when using 

`debugInfo=$(doInvokeEclipse org.eclipse.equinox.p2.director -uninstallIU "${plugin_id}")`

instead it will be removed in `configuration/org.eclipse.equinox.simpleconfigurator/bundles.info` file.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -396,7 +413,7 @@ function doImportEclipse() {
   importWorkingSets="${2}"
   echo "Starting eclipse import for ${import_path} into workspace ${WORKSPACE} at ${WORKSPACE_PATH}"
   # shellcheck disable=SC2086
-  "${ECLIPSE}" -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.ant.core.antRunner -buildfile "${DEVON_IDE_HOME}/scripts/lib_script/import.xml" -data "${WORKSPACE_PATH}" -keyring ~/.eclipse/.keyring -vm "${ECLIPSE_JAVA_HOME}/bin/java" -vmargs -DdevonImportPath="${import_path}" -DdevonImportWorkingSet="${importWorkingSets}" ${ECLIPSE_VMARGS}
+  doStartEclipse -nosplash -application org.eclipse.ant.core.antRunner -buildfile "${DEVON_IDE_HOME}/scripts/lib_script/import.xml" -data "${WORKSPACE_PATH}" -vmargs -DdevonImportPath="${import_path}" -DdevonImportWorkingSet="${importWorkingSets}" ${ECLIPSE_VMARGS}
```
alfeilex, line 416: Could not test this change, because the import did not work.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -357,16 +358,32 @@ function doAddPlugin() {
...
   else
     # shellcheck disable=SC2086
-    "${ECLIPSE}" -clean -data "${WORKSPACE_PATH}" -keyring ~/.eclipse/.keyring -vm "${ECLIPSE_JAVA_HOME}/bin/java" -showlocation "${DEVON_IDE_HOME/*\//}/${WORKSPACE}" -vmargs ${ECLIPSE_VMARGS} &
+    doStartEclipse -clean -data "${WORKSPACE_PATH}" -showlocation "${DEVON_IDE_HOME/*\//}/${WORKSPACE}" &
+  fi
+}
+
+function doStartEclipse() {
+  local start_eclipse
+  start_eclipse=$("${ECLIPSE}" -keyring ~/.eclipse/.keyring -configuration "${DEVON_IDE_HOME}/software/eclipse-plugins/configuration/" -vm "${ECLIPSE_JAVA_HOME}/bin/java" -vmargs "${ECLIPSE_VMARGS}")
```
alfeilex, line 377: What is `ECLIPSE_VMARS` ? This variable is never declared

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -357,16 +358,32 @@ function doAddPlugin() {
...
-function doStartEclipse() {
+
+function doLaunchEclipse() {
   doConfigureEclipse -u
   echo "launching Eclipse in workspace ${WORKSPACE} at ${WORKSPACE_PATH}"
   if doIsWindows
   then
     # shellcheck disable=SC2086
-    start "eclipse" /B "${ECLIPSE}" -clean -data "${WORKSPACE_PATH}" -keyring ~/.eclipse/.keyring -vm "${ECLIPSE_JAVA_HOME}/bin/javaw" -showlocation "${WORKSPACE}" -vmargs ${ECLIPSE_VMARGS}
+    start "eclipse" /B doStartEclipse -clean -data "${WORKSPACE_PATH}" -showlocation "${WORKSPACE}"
```
alfeilex, line 368: This is not working! When we use `doStartEclipse`, the application starts without the custom plugins... the wrapper of eclipse with parameters like `keyring`, `configuration`, `vm` and `vmargs` can not be used easily, because in all places where it is used, further parameters have to be added... does it make sense to still add this wrapper?

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -183,7 +184,7 @@ function doInvokeEclipse() {
   shift
   local eclipse_args=("$@")
   local debug_info
-  debug_info=$("${ECLIPSE}" -consoleLog -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application "${application}" "${eclipse_args[@]}" -vmargs "${ECLIPSE_VMARGS}" | grep '^!MESSAGE' | sed 's/^!MESSAGE //')
+  debug_info=$(doStartEclipse log -application "${application}" "${eclipse_args[@]}" -vmargs "${ECLIPSE_VMARGS}" | grep '^!MESSAGE' | sed 's/^!MESSAGE //')
```
alfeilex, line 187: Why does the function not work here.... If I add `${ECLIPSE}" -consoleLog -nosplash -configuration "${DEVON_IDE_HOME}/software/eclipse-plugins/configuration/" -vm "${ECLIPSE_JAVA_HOME}/bin/java"` here manually the installation of all plugins works fine... but if I use the function there is a freeze after installing `checkstyle`. Where is the difference?

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -357,16 +357,36 @@ function doAddPlugin() {
...
+      # shellcheck disable=SC2086
+      ${default_arguments} -vm "${ECLIPSE_JAVA_HOME}/bin/java" "${eclipse_args[@]}"
+    fi
+  elif [ "${start_mode}" = "log" ]
+    then
+      "${ECLIPSE}" -consoleLog -nosplash -configuration "${configuration_path}" -vm "${ECLIPSE_JAVA_HOME}/bin/java" "${eclipse_args[@]}"
   else
-    # shellcheck disable=SC2086
-    "${ECLIPSE}" -clean -data "${WORKSPACE_PATH}" -keyring ~/.eclipse/.keyring -vm "${ECLIPSE_JAVA_HOME}/bin/java" -showlocation "${DEVON_IDE_HOME/*\//}/${WORKSPACE}" -vmargs ${ECLIPSE_VMARGS} &
+    echo "Please specify how you want to start eclipse, either in gui or log mode"
```
hohwille, line 389: Please use `doFail` instead of `echo` as this is clearly an internal error.

scripts/src/main/resources/scripts/command/eclipse
```diff
@@ -396,7 +416,7 @@ function doImportEclipse() {
   importWorkingSets="${2}"
   echo "Starting eclipse import for ${import_path} into workspace ${WORKSPACE} at ${WORKSPACE_PATH}"
   # shellcheck disable=SC2086
-  "${ECLIPSE}" -nosplash -vm "${ECLIPSE_JAVA_HOME}/bin/java" -application org.eclipse.ant.core.antRunner -buildfile "${DEVON_IDE_HOME}/scripts/lib_script/import.xml" -data "${WORKSPACE_PATH}" -keyring ~/.eclipse/.keyring -vm "${ECLIPSE_JAVA_HOME}/bin/java" -vmargs -DdevonImportPath="${import_path}" -DdevonImportWorkingSet="${importWorkingSets}" ${ECLIPSE_VMARGS}
+  doStartEclipse log -application org.eclipse.ant.core.antRunner -buildfile "${DEVON_IDE_HOME}/scripts/lib_script/import.xml" -data "${WORKSPACE_PATH}" -vmargs -DdevonImportPath="${import_path}" -DdevonImportWorkingSet="${importWorkingSets}" ${ECLIPSE_VMARGS}
```
hohwille, line 419: would it be possible to move the ` -vmargs ${ECLIPSE_VMARGS}` also to `doStartEclipse` function and here just do:
```
ECLIPSE_VMARGS="${ECLIPSE_VMARGS} -DdevonImportPath=\"${import_path}\" -DdevonImportWorkingSet=\"${importWorkingSets}\""
doStartEclipse log ...
```

hohwille (commented):
@alfeilex thanks for your update. Looks good to me 👍

scripts/src/main/resources/scripts/command/eclipse
alfeilex: done

scripts/src/main/resources/scripts/command/eclipse
alfeilex: I can't move this to the `doStartEclipse()` function because I get VM creation errors when I install a new plugin, for example. But I use this idea to get a more clear code.

hohwille: IMHO the error is because `-vmargs` was given twice.

scripts/src/main/resources/scripts/command/eclipse
hohwille: Looks fine anyway as you added the `-vm` and `-keyring` options to `doStartEclipse`.
For the eclipse import we have an ancient issues #609 and #377 that nobody dared to work on for a long time :)

hohwille (approved):
now ready for merge.

Issue #1149, opened 2023-05-08
Version handling for Issue #1084 Adapt scripts to use ide-urls


CREITZ25:
As part of issue #1084, point 2 of the issue is being outsourced and developed separately:

[ ] also functions to list versions and determine the latest version have to be rewritten (available-versions does not exist anymore and a new challenge is that we have to be able to sort the versions properly that we get from the list of version folders and for some tools like Java there are some specific sorting challenges).

create function doUpdateUrls
adapt functions doSetSoftwareVersion and doListSoftwareVersions

Issue #1085, opened 2023-03-28
Align tool names in scripts


hohwille:
For epic #941 and as special addition for story #1084 we want to use the migration from `mirrors` to `urls` also as cleanup:
In general we follow the best practice and convention that if we have a tool `«tool»` then we use the same name in the following places:
* commandlet: `scripts/command/«tool»`
* software folder: `software/[extra/|macos/]«tool»`
* ide-urls folder: `ide-urls/«tool»/«edition»/«version»/`
* variables: `«TOOL»_VERSION` and `«TOOL»_EDITION`
* When running `devon «tool» «args»` you should get the same effect as when running `devon` followed by `«tool» «args»` except that if the tool is not yet installed in the first case it will be downloaded and installed and in the second case it will fail as the tool is not installed.

However, there are currently the following excuses/exceptions to that rule:
* `mvn` vs. `maven` - as the command to run is given as `mvn` we have to use that instead of `maven` consistently.
* `node` vs. `nodejs` - as the command to run is given as `node` we have to use that instead of `nodejs` consistently.
* `sonar` vs. `sonarqube` - here we do not have any matching given command except `StartSonar.bat` or the like that is not even in the `PATH` after set by `devon` command. For simplicity, I would propose to stay with the shorter name `sonar` and rename `sonarqube` occurances to `sonar`.
* `cobigen` vs. `cobigen-cli` vs. `cli`. The command is actually `cg` and `cobigen`. We can discuss with @jan-vcapgemini but IMHO `cobigen` is more self explanatory than `cg` that could confuse new users that might never get what this is about and therefore miss the according features behind it.

The acceptance criteria of this story is that devonfw-ide strictly follows this rules.
Therefore according scripts (see details above) need to be adapted accordingly.
Further we also have to change our `devon.properties`:
https://github.com/devonfw/ide/blob/117cc4684142405103ff64c0f48518c24ee8db15/scripts/src/main/resources/scripts/devon.properties#L18
Please also do a cleanup of internal variables such as `COBIGEN_CLI_HOME` vs. `COBIGEN_HOME`, etc.

This story does NOT care about backward compatibility as this is addressed by #1087.

Blocked by dependencies:
* #1086 
* #1087
* #1035 
* #1084

CREITZ25:
@hohwille 
[...]

PR #1150, opened 2023-05-08
#1149 Version handling for Issue 1084

CREITZ25:
#1149 Version handling for Issue 1084

### DIFF
--- a/documentation/functions.asciidoc
+++ b/documentation/functions.asciidoc
@@ -298,12 +298,15 @@ The first three parameters are taken directly into variables and removed from th
 If the last parameter is a single directory and, on MacOS, not a directory with a name like *.app, it is taken as the source directory.
 Then the function doReplaceExtractedFile is called with the saved parameters, with source directory, target directory and backup directory.
 
-=== doUpdateMirrors
-Uses xref:doGitPullOrClone[] to clone or pull `mirrors` ensuring that everything is up-to-date.
-
 === doUnzip
 Checks if an unzip program is installed, and if not, installs the program.
 The file specified in the first parameter is then unpacked into the directory specified in the second parameter.
+
+=== doUpdateMirrors
+Uses xref:doGitPullOrClone[] to clone or pull `mirrors` ensuring that everything is up-to-date.
+
+=== doUpdateUrls
+Uses xref:doGitPullOrClone[] to clone or pull `urls` ensuring that everything is up-to-date.
 
 == Functions on system PATH, files, or folders
 
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -449,6 +449,12 @@ function doUpdateMirrors() {
   doGitPullOrClone "${mirrors_dir}" "${DEVON_MIRRORS:-https://github.com/devonfw/ide-mirrors.git}"
 }
 
+function doUpdateUrls() {
+  local urls_dir="${DEVON_IDE_HOME}/urls"
+  doDebug "Cloning or updating download urls"
+  doGitPullOrClone "${urls_dir}" "${DEVON_URLS:-https://github.com/devonfw/ide-urls.git}"
+}
+
 # $1: the URL to download
 # $2: the optional, target directory to save to
 # $3: the software to download (e.g. 'java', 'mvn', 'node')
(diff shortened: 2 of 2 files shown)


scripts/src/main/resources/scripts/functions
```diff
@@ -1824,53 +1830,142 @@ function doGetSoftwareVersion() {
...
 
+# $1: software
+function doGetSoftwareEdition() {
+  local software="${1}"
+  local edition
+  echo "${software}" | grep '-' >/dev/null
+  ERG="$?"
+  if [ "${ERG}" = "0" ]
+  then
+    echo "${software}"
```
hohwille, line 1841: I do not really get this? If the software contains a hyphen, it can not have an edition?
So if this is the case for tools like `kotlin-native` you should instead simply use `${software/-/_}` for creating the variable name for the edition.

scripts/src/main/resources/scripts/functions
```diff
@@ -1824,53 +1830,142 @@ function doGetSoftwareVersion() {
...
+  else
+    edition="$(echo "${software}" | awk '{print toupper($0)}')_EDITION"
```
hohwille, line 1843: this is not the edition but the name of the variable potentially holding the edition...

scripts/src/main/resources/scripts/functions
```diff
@@ -1824,53 +1830,142 @@ function doGetSoftwareVersion() {
...
+    if [ "$(printf '%s\n' "${!edition}")" == "" ]
+    then
+      echo "${software}"
+    else
+      echo "${!edition}"
+    fi
```
hohwille, line 1849: I am also confused by this complication. Is there some real rationale behind the `printf` magic and if you append a newline how can it ever be empty?

How about this:
```
function doGetSoftwareEdition() {
  local software="${1}"
  local editionVar
  editionVar="$(echo "${software/-/_}" | awk '{print toupper($0)}')_EDITION"
  local edition="${!editionVar}"
  if [ -z "${edition}" ]
  then
    edition="${software}"
  fi
  echo "${software}"
}
```

scripts/src/main/resources/scripts/functions
```diff
@@ -1824,53 +1830,142 @@ function doGetSoftwareVersion() {
...
-    if [ "${version}" = "latest" ]
+    if [ -n "${version}" ]
     then
-      doUpdateMirrors
-      version="$(head -n 1 "${DEVON_IDE_HOME}/mirrors/${software}/available-versions")"
+      if [ "${version}" = "latest" ]
+      then
+        doUpdateUrls
+        edition=$(doGetSoftwareEdition "${software}")
+        version=doGetLatestSoftwareVersion "${software}"
```
hohwille, line 1888: how can this work? A bash function can only return an exit code.
Did you forget `$(...)`?

scripts/src/main/resources/scripts/functions
```diff
@@ -1824,53 +1830,142 @@ function doGetSoftwareVersion() {
...
+      if [ "${software_version_variable}" = "nodejs" ]
+      then
+        software_version_variable="node"
+      fi
+      software_version_variable="$(echo "${software_version_variable}" | awk '{print toupper($0)}')_VERSION"
+      local devon_properties="${DEVON_IDE_HOME}/settings/devon.properties"
+      grep -v "${software_version_variable}" "${devon_properties}" > "${devon_properties}.tmp"
+      mv "${devon_properties}.tmp" "${devon_properties}"
+      echo "${software_version_variable}=${version}" >> "${devon_properties}"
+      doEchoInteraction "${software_version_variable}=${version} has been set in ${devon_properties}\nTo install that version call the following command:\ndevon ${commandlet} setup"
```
hohwille, line 1924: why duplicating all this code?

scripts/src/main/resources/scripts/functions
```diff
@@ -1824,53 +1830,142 @@ function doGetSoftwareVersion() {
...
+    local anz_versions
+    anz_versions="$(find "${versions_path}" -mindepth 1 -maxdepth 1 -printf "%f\n" | wc -l)"
+    if [ ! -d "${versions_path}" ]
+    then
+      doError "No versions available at ${versions_path}"
+    elif [ "${anz_versions}" -lt 20 ]
+    then
+      find "${versions_path}" -mindepth 1 -maxdepth 1 -printf "%f\n" | sort -rV | pr -4 -t -s | column -t
+    else
+      find "${versions_path}" -mindepth 1 -maxdepth 1 -printf "%f\n" | sort -rV | head -20 | pr -4 -t -s | column -t
```
hohwille, line 1949: Why do we call `find` twice wasting the results of the first call?
also quite some duplications and already before your change things like `pr -4 -t -s` and `column -t` had been duplicated. Now we have it 4 times.

hohwille (changes requested):
@CREITZ25 thanks for your PR. IMHO some things are not correct and need to be reworked.
Will we have the redundancies only for a short time as we later remove the `if`s and only keep the `ide-urls` branch?

scripts/src/main/resources/scripts/functions
hohwille: my code is not tested - just a best guess

scripts/src/main/resources/scripts/functions
hohwille: For the record: you determine the edition here via `doGetSoftwareEditon` and the very next call `doGetLatestSoftwareVersion` does not get the edition as argument but again has to compute it again via the same `doGetSoftwareEditon` function. With every day I am getting more keen on our migration to Java...

scripts/src/main/resources/scripts/functions
CREITZ25: With the agreement that a "-" in the command name leads to a "_" in the variables, the code can be simplified accordingly and is now valid without restrictions.

scripts/src/main/resources/scripts/functions
CREITZ25: Same as above, and the variable name was chosen more meaningfully

scripts/src/main/resources/scripts/functions
CREITZ25: I didn't question the sense of the code, I just replaced "cat file" 1:1 with "find directories". Now, the code is optimized and looks easier.

scripts/src/main/resources/scripts/functions
CREITZ25: This part is also in the else part - the mirror part. If we remove the mirror part, it's easier to delete the else part entirely, keep only the then part and remove the if construction. Then the code is there once. That was the idea.

CREITZ25 (commented):
Changes done, please review again

scripts/src/main/resources/scripts/functions
```diff
@@ -694,7 +700,7 @@ function doDownloadInternal() {
 # $1: filename (potentiall including absolute or relative path)
 # echos the file extension
 function doGetExtension() {
-  local filename="${1/*\//}"
+  local filename="${1/*\//}"     usr/test/file.txt   file.txt
```
hohwille, line 703: this looks like a Copy&Paste mistake

documentation/functions.asciidoc
```diff
@@ -298,13 +298,16 @@ The first three parameters are taken directly into variables and removed from th
...
-
 === doUnzip
 Checks if an unzip program is installed, and if not, installs the program.
 The file specified in the first parameter is then unpacked into the directory specified in the second parameter.
 
+=== doUpdateMirrors
+Uses xref:doGitPullOrClone[] to clone or pull `mirrors` ensuring that everything is up-to-date.
+
+=== doUpdateUrls
+Uses xref:doGitPullOrClone[] to clone or pull `urls` ensuring that everything is up-to-date.
```
hohwille, line 309: Great that you reworked the functions avoiding side effects via global variables.
Can you also update the documentation now?

documentation/functions.asciidoc
```diff
@@ -357,9 +360,15 @@ Determines whether the actual version is contained in the security file for the
...
 
+=== doGetLatestSoftwareVersion
+Determines the latest available version of a software and write it in the variable "latest_version".
+
 === doGetNextVersion
 A version number is passed to the function doGetNextVersion as an argument and the next version number is generated from this by incrementing the last digit by one and outputs it.
 
+=== doGetSoftwareEdition
+Determines the currently in the environment configured edition of a software and write it in the variable "edition".
+
```
hohwille, line 371: Great that you reworked the functions avoiding side effects via global variables.
Can you also update the documentation now?

scripts/src/main/resources/scripts/functions
CREITZ25: Yes, an accident. It is corrected

documentation/functions.asciidoc
CREITZ25: Documentation is updated

CREITZ25 (commented):
Changes done

scripts/src/main/resources/scripts/functions
```diff
@@ -1810,53 +1816,138 @@ function doGetSoftwareVersion() {
...
+        software_version_variable="node"
+      fi
+      software_version_variable="$(echo "${software_version_variable}" | awk '{print toupper($0)}')_VERSION"
+      local devon_properties="${DEVON_IDE_HOME}/settings/devon.properties"
+      grep -v "${software_version_variable}" "${devon_properties}" > "${devon_properties}.tmp"
+      mv "${devon_properties}.tmp" "${devon_properties}"
+      echo "${software_version_variable}=${version}" >> "${devon_properties}"
+      doEchoInteraction "${software_version_variable}=${version} has been set in ${devon_properties}\nTo install that version call the following command:\ndevon ${commandlet} setup"
+    else
+      doError "You have to specify the version you want to set."
```
hohwille, line 1883: This will cause compatibility issues or you have to adapt all commandlets.
So far the empty string for version is treated the same as `latest` to make it work if $«TOOL»_VERSION is unset.
I would strongly recommend to keep this behavior and just join the ifs so if version is empty or latest, you simply first set `version` to the most recent one.

hohwille (changes requested):

scripts/src/main/resources/scripts/functions
```diff
@@ -1810,53 +1816,138 @@ function doGetSoftwareVersion() {
...
+  if [ -d "${DEVON_IDE_HOME}/urls" ]
   then
-    if [ "${version}" = "latest" ]
+    if [ -n "${version}" ]
     then
-      doUpdateMirrors
-      version="$(head -n 1 "${DEVON_IDE_HOME}/mirrors/${software}/available-versions")"
+      if [ "${version}" = "latest" ]
+      then
+        doUpdateUrls
```
hohwille, line 1867: this already happens in `doGetLatestSoftwareVersion`. Doing `git pull` twice is pointless waste. Please remove this line.

scripts/src/main/resources/scripts/functions
```diff
@@ -1810,53 +1816,138 @@ function doGetSoftwareVersion() {
...
+      then
+        doUpdateMirrors
+        version="$(head -n 1 "${DEVON_IDE_HOME}/mirrors/${software}/available-versions")"
+      fi
+      local software_version_variable="${software}"
+      # temporary workaround for #1098 should be fixed and removed with #1085
+      if [ "${software_version_variable}" = "nodejs" ]
+      then
+        software_version_variable="node"
+      fi
```
hohwille, line 1898: #1085 is already implemented and merged. Please remove.

scripts/src/main/resources/scripts/functions
CREITZ25: Yes, thanks for the hint. Line is removed.

hohwille (approved):
@CREITZ25 Excellent. 👍 
Ready for merge.

Issue #1131, opened 2023-04-28
treat 404 downloads of url-updater


hohwille:
Our url-updater is hitting many 404 (not found) errors for download URLs.
Sometimes various versions that we find seem not to exist as download or at least not in the expected URL schema.
In other cases only a particular OS+architecture combination seems to be missing leading to 404.

Before we decide to just ignore all these errors, we should do some investigation and analysis if maybe we have to fix something.
Further, my expectation is that for versions that are found in some HTML structure but not as download file entirely (all OS+architecture), we will repeat the according errors every night what is undesired. However, we also do not want to add a folder for this version containing no download as the existence of a version folder implies the existence of the tool edition version for devonfw-ide (at least by the current design).

alfeilex:
`aws` version 2.0.32 does not exist and therefore the url responds with 404 error

alfeilex:
- AWS
 > 2.0.32 does not exist and therefore the url responds with 404 error
- Azure
 > Fix with new regex [910fe25](https://github.com/devonfw/ide/pull/1155/commits/910fe25f2de73d0d78b7f68702b4af8e029f258c)
- Dotnet
 > Fix with new regex [910fe25](https://github.com/devonfw/ide/pull/1155/commits/910fe25f2de73d0d78b7f68702b4af8e029f258c)

alfeilex:
Should we add some functions to filter out versions, which do not provide the required installation files? Or should we just override `mapVersion` function in each updater? 

But sometimes there are versions that do not provide windows installation files up to certain version... how we can control this?

alfeilex:
The most 404 errors occurs because 

- download links for arm architecture are not available for all releases
- versions are crawled in wrong format

alfeilex:
See the changelog for the tools in related PR

PR #1155, opened 2023-05-11
#1131: Fix 404 downloads of url-updater

alfeilex:
Summary of fixes for each tool:

- **Azure**: map versions greater than` 2.18.0` as they provide MSI files
- **Dotnet**: map versions if they match `v\\d+\\.\\d+\\.\\d+`
- **GCViewer**: remove versions like `1.31-RC5`
- **GH**: check versions for mac arm64 files greater than `2.24.0`
- **Helm**: check versions for arm greater than `3.5.0`
- **LazyDocker**: check versions for windows greater than `0.7.5` and versions for arm greater than `0.16.0`
- **Node**: check version greater than `4.0.0` and version for win arm greater than 20.0.0 and darwin arm `16.0.0`
- **Quarkus**: check versions greater than `2.6.0`
- **Terraform**: check versions for mac arm greater than 1.2.0
- **VSCode**: map versions with `\d+\.\d+\.\d+`

Open questions:
- **Kotlin**: Why we use the WebsiteCrawler instead of GithubCrawer there? What is the difference between Kotlin Native and C?
- **Python**: there are releases which do not provide embed files for windows -> 404 errors will be persists
- **Sonartype**: some versions crawled on github do not provide correct download urls --> hard to fix

Related issue: #1131


url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcviewer/GcViewerUrlUpdater.java
```diff
@@ -30,4 +35,14 @@ protected String getGithubRepository() {
 
     return "GCViewer";
   }
+
+  @Override
+  protected void collectVersionsFromJson(GithubTags jsonItem, Collection<String> versions) {
+
+    for (GithubTag item : jsonItem) {
+      String version = item.getRef().replace("refs/tags/", "");
+      if(version.matches("\\d+\\.\\d+(\\.\\d+)?"))
```
alfeilex, line 44: How should the versions be filtered in general? Should we add a condition here in this function or use "mapVersion" in general?

hohwille: Why do you have to duplicate this at all?
https://github.com/devonfw/ide/blob/169da0cc26df54921ee82d466de8a47d52a9e1a9/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/GithubUrlUpdater.java#L29

alfeilex: There are some versions that are crawled, but they dont match the version format as 2.5.6

I could also use the `mapVersion` function, but I thought that we can already filter the versions before we save them. Or am I miss something here?

alfeilex: I use the mapVersion function

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/node/NodeUrlUpdater.java
```diff
@@ -28,13 +34,19 @@ protected String getGithubRepository() {
...
-    String baseUrl = "https://nodejs.org/dist/${version}/node-${version}-";
-    doAddVersion(urlVersion, baseUrl + "win-x64.zip", WINDOWS);
-    doAddVersion(urlVersion, baseUrl + "win-aarch64.zip", WINDOWS, ARM64, "");
-    doAddVersion(urlVersion, baseUrl + "linux-x64.tar.gz", LINUX);
-    doAddVersion(urlVersion, baseUrl + "linux-aarch64.tar.gz", LINUX, ARM64, "");
-    doAddVersion(urlVersion, baseUrl + "darwin-x64.tar.gz", MAC);
-    doAddVersion(urlVersion, baseUrl + "darwin-aarch64.tar.gz", MAC, ARM64, "");
+    String version = urlVersion.getName();
+
+    String baseUrl = "https://nodejs.org/dist/v${version}/node-v${version}-";
```
hohwille, line 39: you are breaking compatibility with this change:
https://github.com/devonfw/ide-mirrors/blob/master/node/urls

I already had discussions with @CREITZ25 on these topics and I said that typically a version number should not contain static artificial prefixes or infixes. There are even cases (for other tools) where in the URL the version is once with such prefix and another time without it. As adding such prefix like `v` via `v${version}` is easy while removing it is not, we have cases where we had already been forced to remove it. However, there has even been a case where for the same tool versions with such prefix as well as others without it existed so it always becomes tricky because of the mess that different tool vendors create and the lack of a commonly accepted best-practice.

I am OK with changing this but then we have to add it to the migration (#1087).

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/helm/HelmUrlUpdater.java
```diff
@@ -23,11 +29,13 @@ protected String getGithubOrganization() {
...
 
-    String baseUrl = "https://get.helm.sh/helm-${version}-";
+    String version = urlVersion.getName();
+    String baseUrl = "https://get.helm.sh/helm-v${version}-";
     doAddVersion(urlVersion, baseUrl + "windows-amd64.zip", WINDOWS);
     doAddVersion(urlVersion, baseUrl + "linux-amd64.tar.gz", LINUX);
     doAddVersion(urlVersion, baseUrl + "darwin-amd64.tar.gz", MAC);
-    doAddVersion(urlVersion, baseUrl + "darwin-arm64.tar.gz", MAC, ARM64, "");
+    if (isVersionGreaterThan(version, "3.4.0"))
+      doAddVersion(urlVersion, baseUrl + "darwin-arm64.tar.gz", MAC, ARM64,"");
```
hohwille, line 38: In Java it is best practice to ALWAYS use curly braces for `if` statements. Why?
Look at the following valid Java example:
```
public void myMethod() {
  if (false)
    System.out.println("Inconsistentcy detected: false is true");
    System.exit(-1);
  System.out.println("All is sane here");
}
```

Due to the indentation the developer might have assumed that the second statement (`System.exit`) still belongs to the `if` block but it does not and the code does not behave as expected. This is a common pitfall when extending existing code. It will always get clear and obvious if your code looks like this (also the wrong indentation):

```
public void myMethod() {
  if (false) {
    System.out.println("Inconsistentcy detected: false is true");
  }
    System.exit(-1);
  System.out.println("All is sane here");
}
```

We have created a formatter for Eclipse in devonfw-ide that automatically aligns and formats your code whenever you save it resolving all such anti-patterns. Are you using IntelliJ oder VisualStudioCode? Maybe we can invest to get the same autoformatter preconfigured for your favorite IDE....

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -575,4 +576,16 @@ protected final void addVersion(String version, Collection<String> versions) {
...
+   * @param compareToVersion the version to compare with (e.g. "2.5.3").
+   * @return boolean if the baseVersion is greater than compareToVersion.
+   */
+  protected boolean isVersionGreaterThan(String baseVersion, String compareToVersion) {
+    VersionIdentifier vid = VersionIdentifier.of(baseVersion);
+    if (vid.compareVersion(VersionIdentifier.of(compareToVersion)).isGreater())
+      return true;
+    else
+      return false;
+  }
```
hohwille, line 590: For the record: I only found the time to implement this `VersionIdentifier` but not yet to "integrate" it.
Your code here is correct and can be merged.
However, it would be much more elegant to:
* add `VersionIdentifier` to `UrlVersion` via `getVersionIdentifier()` method to prevent parsing the same version multiple times and reduce waste of performance/time.
* follow DDD best-practices add methods like `isGreater`, `isLess`, `isGreaterOrEqual`, and `isLessOrEqual` to `VersionObject` instead of adding this method.
* for the static versions you are comparing against you could define constants so they are parsed only once and not for every version that is processed.

hohwille (changes requested):
@alfeilex thanks for your great PR that solves many if not all of our 404 problems 👍 
Can you please have a look at my review comments?

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/node/NodeUrlUpdater.java
alfeilex: I use `mapVersion` function to add `v` to maintain compatibility

hohwille (approved):
@alfeilex thanks for your great rework 👍 
Ready for review.

Issue #1156, opened 2023-05-11
Correct wrong version numbers for helm


CREITZ25:
### Expected behavior

The release numbers of the helm software consist of v and a number, e.g. v3.12.0.
The file name of the software that is then downloaded contains also this release number, e.g. helm-v3.12.0-windows-amd64.zip .

### Actual behavior

In the new IDE-URLS repository, the releases of helm are listed without a leading v..., and should be adjusted according to our conventions.

# steps

1. remove v from https://github.com/devonfw/ide/blob/master/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/helm/HelmUrlUpdater.java#L32
2. Remove methode at this point:
https://github.com/devonfw/ide/blob/fa4de78f2aaca0d48b5e36f3d7b7e279e230cf9e/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/helm/HelmUrlUpdater.java#L24-L27
3. Delete folder https://github.com/devonfw/ide-urls/tree/master/helm/ 
4. Wait one night until everything is refilled

PR #1157, opened 2023-05-11
#1156-Correct-wrong-version-numbers-for-helm

CREITZ25:
feature/#1156-Correct-wrong-version-numbers-for-helm


hohwille (approved):
@CREITZ25 thanks for the PR. So we keep the `v` as part of the version number and have compatibility with mirrors and earlier devonfw-ide releases (devon.properties `HELM_VERSION=vxxx`). Ready for merge. We need to delete `helm` in url-mirrors. To avoid clutter, we might need to purge the git history before we go live with urls.

Issue #1162, opened 2023-05-16
Action update-urls.yml takes too long and fails


hohwille:
### Expected behavior

As a devonfw-ide PO, I want the process updating our ide-urls repo to be smooth and stable so that we do not have to worry about it and always have the latest configs (all versions of all tools) available.

### Actual behavior

The process failed because it exceeded the limit of 6 hours that github defined for actions:
https://github.com/devonfw/ide/actions/runs/4987428010/jobs/8929179869

As a result we cause a big problem:
* once we collected so much work to do for the job the problem is a complete blocker
* every night he will start again, do the same things wasting a lot of resources (bandwith, tempoary storage, 6h CPU, etc.)
* as all is "transactional" with an "all or nothing" principle we never push any progress even though a lot of work had been done

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Nothing (just watch the logs and read this issue...)

### Related/Dependent Issues

#941 

### Comments/Hints:

The design I created actually allows to do partial things and continue on errors within our Java process.
So what we would need is to commit and push our work that has already been made before the process gets terminated when it takes too long. Therefore I see the following options:

1. Call the Java-Programm for each/tool edition separately in a loop from the action and after each run commit and push the changes
2. Change the Java-Program so that it can accept a timeout as configuration parameter and when that is reached it will stop looping over next tools/editions/versions and simply aborts (without error). We could set that timeout to e.g. 5:30 in the github action so we still have enough time to ensure the changes are comitted and pushed.

IMHO the 2. approach is easier to implement and better to maintain.

### Affected version:

* OS: Windows/Linux/Mac?
* Browser: Chrome/Firefox/Safari?

hohwille:
was auto-closed by github due to merge of associated PR.
Still not done:
> Change the Java-Program so that it can accept a timeout as configuration parameter and when that is reached it will stop looping over next tools/editions/versions and simply aborts (without error). We could set that timeout to e.g. 5:30 in the github action so we still have enough time to ensure the changes are comitted and pushed.

PR #1166, opened 2023-05-22
#1162 android studio json urls

jan-vcapgemini:
Addresses/Fixes: #1162 

Implements:
* added first implementation of download URL and checksum extraction from version json
* fixed updateStatusJson (can now handle initial URL generate using update false)
* cleaned up AndroidStudioUrlUpdater and test mock
* added AndroidJsonDownload
* added download to AndroidJsonItem
* removed duplicated doAddVersion method (added check for checksum param instead)
* added new checksum param to doAddVersion (defaults to empty string)
* updated javadocs
* added links to javadocs

### DIFF
--- a/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/docker/DockerDesktopUrlUpdater.java
+++ b/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/docker/DockerDesktopUrlUpdater.java
@@ -33,8 +33,8 @@ public class DockerDesktopUrlUpdater extends WebsiteUrlUpdater {
       if (!success) {
         return;
       }
-      doAddVersion(urlVersion, "https://desktop.docker.com/mac/main/amd64/" + code + "/Docker.dmg", MAC, X64);
-      doAddVersion(urlVersion, "https://desktop.docker.com/mac/main/arm64/" + code + "/Docker.dmg", MAC, ARM64);
+      doAddVersion(urlVersion, "https://desktop.docker.com/mac/main/amd64/" + code + "/Docker.dmg", MAC, X64, "");
+      doAddVersion(urlVersion, "https://desktop.docker.com/mac/main/arm64/" + code + "/Docker.dmg", MAC, ARM64, "");
     } else {
       // For the latest version, there is no code in the url.
       // TODO but that means that the implementation is wrong as the URL will then change later and is therefore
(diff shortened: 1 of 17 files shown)


jan-vcapgemini:
I'm still working on a refactoring of the checkDownloadUrl methods. But we need this now to fix the failing url updater.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/docker/DockerDesktopUrlUpdater.java
```diff
@@ -33,8 +33,8 @@ protected void addVersion(UrlVersion urlVersion) {
       if (!success) {
         return;
       }
-      doAddVersion(urlVersion, "https://desktop.docker.com/mac/main/amd64/" + code + "/Docker.dmg", MAC, X64);
-      doAddVersion(urlVersion, "https://desktop.docker.com/mac/main/arm64/" + code + "/Docker.dmg", MAC, ARM64);
+      doAddVersion(urlVersion, "https://desktop.docker.com/mac/main/amd64/" + code + "/Docker.dmg", MAC, X64, "");
```
hohwille, line 36: in such case it is better to keep a method with the exact old signature that internally delegates to the new one with the added parameter.

hohwille (approved):
@jan-vcapgemini Thanks for the PR. Excellent work 👍 
I really appreciate that you not only fix the problem nicely but you also tidy up code and add missing JavaDoc-links, etc.

For the record: we can see some little diff wars left between Eclipse and Intellij. It would be great, if you could create a story and we further investigate how we can change our ide-settings and workspaces for Eclipse and Intellij to produce the same import order also in IntelliJ and align the JavaDoc indent.

I left a comment that you should notice and that you could address in the further refactoring in a later PR. For that we should also avoid the redundancies between the two `checkDownloadUrl` methods (see e.g. content-type checking and error logging).
However, really great work and I will merge as is so we can see the result of the nightly github action tomorrow.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/docker/DockerDesktopUrlUpdater.java
hohwille: Please address in "refactoring PR". The need to pass an empty checksum for all cases where an architecture is specified seems pointless.

Issue #1142, opened 2023-05-05
wrong checksum in ide-urls


hohwille:
### Expected behavior

Our update process populating `ide-urls` shall compute correct sha256 checksums for each download file so that `devonfw-ide` will be able to verify that checksum. Therefore when I download the according file and compute the checksum with `sha256sum`, I should get the exact same hash.

### Actual behavior

For some tools the checksums are computed wrong. According to my analysis some download URLs lead to HTML websites that themselves contain some JavaScript redirecting to the actual download on a mirror. This especially applies to sourceforge.net.
Tools like `curl` can magically handle this and do the actual download. However, our according `UrlUpdater`s cannot handle this situation. As a result they consider the HTML content as the actual download file. Therefore they compute the SHA256 checksum of the HTML content instead of the binary download file.
I have already changed the code such that the SHA256 checksum computation will fail if the content type header is present and its value starts with `text` (such as `text/html`).

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

$ sha256sum ~/Downloads/devonfw-ide/npm-9.6.0-windows.tgz
628f7650e50a4081746a681c74aeca48f9d1cebb333d4d69fdffaa1a487c3ada *...
 
$ cat urls/npm/npm/9.6.0/windows_x64.urls.sha256
5f1da1ef6ba776449900c931e7acc3b60e47f3080425bf0e6a1c0c5a07dbd6fd

https://github.com/devonfw/ide-urls/blob/master/npm/npm/9.6.0/windows_x64.urls.sha256

### Related/Dependent Issues

#1009

### Affected version:

2023.04.001

hohwille:
For `npm` this is just a simple and stupid bug here:
https://github.com/devonfw/ide/blob/eed1f40356b8114358a74b6ab34c6b2ddb326726/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/npm/NpmUrlUpdater.java#L21

hohwille:
If I am not mistaken then only problem left that I could find is gcviewer. Example URL: https://sourceforge.net/projects/gcviewer/files/gcviewer-1.36.jar

hohwille:
```
$ curl https://sourceforge.net/projects/gcviewer/files/gcviewer-1.36.jar
<html>
<head>
<title>301 Moved Permanently</title>
</head>
<body>
<h1>301 Moved Permanently</h1>
[...]

PR #1167, opened 2023-05-22
#1142 wrong checksum in ide urls for npm

moritzLanger:
#1142 
+ fixed the wrong url, which was used for the npm baseURL

+ changed the Version pattern, so to not include non existing npm versions.

### DIFF
--- a/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/dotnet/DotNetUrlUpdater.java
+++ b/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/dotnet/DotNetUrlUpdater.java
@@ -23,8 +23,8 @@ public class DotNetUrlUpdater extends GithubUrlUpdater {
   protected void addVersion(UrlVersion urlVersion) {
 
     String baseUrl = "https://dotnetcli.azureedge.net/dotnet/Sdk/${version}/dotnet-sdk-${version}-";
-    boolean ok1 = doAddVersion(urlVersion, baseUrl + "win-x64.exe", WINDOWS, X64, "");
-    boolean ok2 = doAddVersion(urlVersion, baseUrl + "win-arm64.exe", WINDOWS, ARM64, "");
+    boolean ok1 = doAddVersion(urlVersion, baseUrl + "win-x64.zip", WINDOWS, X64, "");
+    boolean ok2 = doAddVersion(urlVersion, baseUrl + "win-arm64.zip", WINDOWS, ARM64, "");
     if (!ok1 && !ok2) {
       return;
     }


url-updater/src/main/java/com/devonfw/tools/ide/url/updater/npm/NpmUrlUpdater.java
```diff
@@ -18,11 +18,7 @@ protected String getTool() {
   @Override
   protected void addVersion(UrlVersion urlVersion) {
 
-    String baseUrl = "https://nodejs.org/dist/v${version}/node-v${version}";
-    doAddVersion(urlVersion, baseUrl + "-win-x64.zip", WINDOWS, X64);
-    doAddVersion(urlVersion, baseUrl + "-darwin-x64.tar.gz", MAC, X64);
-    doAddVersion(urlVersion, baseUrl + ".pkg", MAC, ARM64);
-    doAddVersion(urlVersion, baseUrl + "-linux-x64.tar.xz", LINUX, X64);
+    doAddVersion(urlVersion,"https://registry.npmjs.org/npm/-/${version}.tgz");
```
hohwille, line 21: Did you test this? IMHO this can not be correct.
e.g. https://registry.npmjs.org/npm/-/8.19.3.tgz will lead to 404. Also Node.js is OS specific so there is no single download that works on every OS...

hohwille (changes requested):
@moritzLanger thanks for your PR. Can you have a look as IMHO you broke the download links.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/npm/NpmUrlUpdater.java
moritzLanger: The 404 happens because the npm- is missing before the version number, but because I changed the VersionPattern, it will be included meaning the version download is https://registry.npmjs.org/npm/-/npm-8.19.3.tgz.

moritzLanger: I also understand that node is os depended, and so might be npm, only I was trying to emulate the mirror repo, which also has only a single version for npm, as their are no other downloads than .tgz

hohwille: But the version number is `8.19.3` and not `npm-8.19.3`. Also if we would change the version this way, we would break existing installations of devonfw-ide with the next release that have configured the version number without the `npm-` prefix. Then simply add the `npm-` prefix instead to the URL pattern.

hohwille: For the record:
https://github.com/devonfw/ide-mirrors/blob/master/npm/urls

PR #1169, opened 2023-05-23
#1162: applied refactorings

jan-vcapgemini:
Addresses: #1162 #1166 

Imlements:
* refactored checkDownloadUrl #1166 
* added isContentValid and isChecksumStillValid methods added new doAddVersion without checksum
* removed emtpy string from old doAddVersion calls using os and architecture
* removed checksum check from doAddVersion
* removed 2nd checkDownloadUrl method
* renamed UpdateManagerIT to JsonUrlUpdaterIT
* replaced base URLs with localhost
* added test and resource for missing checksums
* added test for missing downloads
* renamed tests


url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -190,58 +201,59 @@ protected boolean isSuccess(HttpResponse<?> response) {
...
 
   /**
-   * Checks the download URL and takes the provided checksum into account instead of downloading the file and generating
-   * the checksum from it
+   * Checks if the download file checksum is still valid
    *
-   * @param url the URL of the download to check.
-   * @param urlVersion the {@link UrlVersion} where to store the collected information like status and checksum.
+   * @param url String of the URL to check
+   * @param urlVersion String of the version
```
hohwille, line 207: this change is incorrect and the old Javadoc of `urlVersion` was correct and more appropriate. Was this on accident or on purpose?

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -190,58 +201,59 @@ protected boolean isSuccess(HttpResponse<?> response) {
...
+  private static boolean isChecksumStillValid(String url, UrlVersion urlVersion, OperatingSystem os,
+      SystemArchitecture architecture, String checksum, String tool, String version) {
+
+    UrlDownloadFile urlDownloadFile = urlVersion.getOrCreateUrls(os, architecture);
+    UrlChecksum urlChecksum = urlVersion.getOrCreateChecksum(urlDownloadFile.getName());
+    urlDownloadFile.addUrl(url);
+    urlChecksum.setChecksum(checksum);
+    String oldChecksum = urlChecksum.getChecksum();
+
+    if ((oldChecksum != null) && !Objects.equal(oldChecksum, checksum)) {
```
hohwille, line 224: IMHO this code does not make sense:
The getter will return what has been given to the setter so `checksum` and `oldChecksum` will always refer to the same value.
IMHO the entire method `isChecksumStillValid` does not make sense yet. If we would have a method with that name then either to verify the checksum for the end-user but that code is currently still in bash or in case we would revalidate this on update of existing versions. However, due to the greediness of checksum computation this is not planned.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -190,58 +201,59 @@ protected boolean isSuccess(HttpResponse<?> response) {
...
+  /**
+   * Checks if the content type is valid (not of type text)
+   *
+   * @param url String of the url to check
+   * @param tool String of the tool name
+   * @param version String of the version
+   * @param contentType String of the content type
+   * @return {@code true} if the content type is not of type text, {@code false} otherwise.
+   */
+  private static boolean isContentTypeValid(String url, String tool, String version,
```
hohwille, line 244: maybe `isValidDownload` would be more suitable as `text/html` in general is a valid content-type but not for a binary download. Actually we would expect `application/octet-stream` but not every server is 100% compliant so I better went for a blacklist than a whitelist.

hohwille (changes requested):
@jan-vcapgemini Thanks for your PR. Great that you cleaned the signature change and awesome that you created excellent integration tests. 👍 
Can you have a look at my review comments and update accordingly?

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
hohwille: You could even pass the response as parameter and read the content type within this method, what would make more even sense. Thanks for extracting such smaller methods to make the code better!

hohwille: Maybe even one step further, you could then also do the `isSuccess` check in this method and simplify `checkDownloadUrl` further.

url-updater/src/test/java/com/devonfw/tools/ide/integrationtests/JsonUrlUpdaterIT.java
```diff
@@ -0,0 +1,128 @@
...
+    AndroidStudioUrlUpdaterMock updater = new AndroidStudioUrlUpdaterMock();
+
+    // when
+    updater.update(urlRepository);
+
+    Path androidStudioVersionsPath = tempDir.resolve("android-studio").resolve("android-studio").resolve("2023.1.1.2");
+
+    // then
+    assertThat(androidStudioVersionsPath.resolve("windows_x64.urls.sha256")).exists()
+        .hasContent("de08da1685e537e887fbbe1eb3278fed38aff9da5d112d96115150e8771a0f30");
```
hohwille, line 125: is this the SHA256 sum of "aBody"? If so adding a comment or extracting to a named constant could help the reader of the code.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
jan-vcapgemini: I've implemented all of your mentioned points. Thanks for your constructive feedback. I hope that it looks better now.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -250,37 +262,34 @@ private boolean checkDownloadUrl(String url, UrlVersion urlVersion, OperatingSys
    * @return {@code true} if the download was checked successfully, {@code false} otherwise.
    */
   private boolean checkDownloadUrl(String url, UrlVersion urlVersion, OperatingSystem os,
-      SystemArchitecture architecture) {
+      SystemArchitecture architecture, String checksum) {
 
     HttpResponse<InputStream> response = doGetResponseAsStream(url);
```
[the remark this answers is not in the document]
jan-vcapgemini: I've adjusted it now.

url-updater/src/test/java/com/devonfw/tools/ide/integrationtests/JsonUrlUpdaterIT.java
jan-vcapgemini: I've extracted a named constant now.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
jan-vcapgemini: I've removed the addUrl and setChecksum now. Not sure if we should remove the whole method though.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -250,37 +265,30 @@ private boolean checkDownloadUrl(String url, UrlVersion urlVersion, OperatingSys
...
-      String oldChecksum = urlChecksum.getChecksum();
-      if ((oldChecksum != null) && !Objects.equal(oldChecksum, checksum)) {
-        logger.error("For tool {} and version {} the mirror URL {} points to a different checksum {} but expected {}.",
-            tool, version, url, checksum, oldChecksum);
-        success = false;
-      } else {
-        urlDownloadFile.addUrl(url);
-        urlChecksum.setChecksum(checksum);
+      if (checksum == null || checksum.isEmpty()) {
+        String contentType = response.headers().firstValue("content-type").orElse("undefined");
```
hohwille, line 279: This line is still redundant and could be moved to a private method for reuse.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -250,37 +265,30 @@ private boolean checkDownloadUrl(String url, UrlVersion urlVersion, OperatingSys
    * @return {@code true} if the download was checked successfully, {@code false} otherwise.
    */
   private boolean checkDownloadUrl(String url, UrlVersion urlVersion, OperatingSystem os,
-      SystemArchitecture architecture) {
+      SystemArchitecture architecture, String checksum) {
 
-    HttpResponse<InputStream> response = doGetResponseAsStream(url);
+    HttpResponse<?> response = doCheckDownloadViaHeadRequest(url);
```
hohwille, line 270: It would be even smarter to distinguish if we already have a checksum or not:
* if we already have a checksum given, a HEAD request is sufficient
* otherwise we need the full download and can do only a GET - currently you always do HEAD and then additionally to a GET request what is overhead.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -333,7 +341,7 @@ protected HttpResponse<?> doCheckDownloadViaHeadRequest(String url) {
 
       return this.client.send(request, HttpResponse.BodyHandlers.ofString());
     } catch (Exception e) {
-      logger.error("Failed to preform HEAD request of URL {}", url, e);
+      logger.error("Failed to perform HEAD request of URL {}", url, e);
```
hohwille, line 344: thanks for also fixing typos :)

hohwille (approved):
@jan-vcapgemini thanks for your rework. This PR is now great and so far correct. 👍 
I again left some nice-to-have comments for the future but I want to get this merged to resolve the conflicts due to the signature changes.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
hohwille: We will do this after the merge to not block the merge.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
hohwille: this is a really picky nice-to-have that can be done after the merge.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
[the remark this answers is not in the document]
hohwille: Thanks. See my new comment for further improvements...

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
hohwille: We can discuss this after merge. However, the method still seems pointless to me: if it is called from `addVersion` the version is new and no previous checksum can exist. If we use this method on update of versions this makes sense but in such case we never set the checksum, as we already assume that it cannot change afterwards - a release by design has to be immutable.

PR #1174, opened 2023-05-30
#1142 wrong checksum in ide urls for npm 

moritzLanger:
partially solving #1142 

- fixed the wrong url, which was used for the npm baseURL

- changed the Version pattern, so to not include non existing npm versions.


hohwille (approved):
@moritzLanger thanks for your rework. Now perfect and ready for merge 👍

hohwille:
@moritzLanger can you also create a PR that deletes the entire `npm` folder from `ide-urls`?
I will then merge and in the next night it will be re-created correctly from the github action.

PR #1175, opened 2023-05-31
Sonar start and stop options

Oumal1:


hohwille (approved):
@Oumal1 thanks for your PR. As `StopNTService.bat` does not exist anymore in newer versions this is some improvement and ready for merge 👍 
However, it will not entirely solve the problem and will not work out of the box (if the service is not installed in windows).

Issue #1147, opened 2023-05-08
Separate plugins from software


hohwille:
We should move plugins from `software` to their own top-level `plugins` folder as they are something different to the rest we put in software.

Current layout:
```
/ DEVON_IDE_HOME
└──/ software
   ├──/ az
   ├──/ cobigen-cli
   ├──/ dotnet
   ├──/ eclipse
   ├──/ eclipse-plugins
   ├──/ extra
   ├──/ intellij
   ├──/ intellij-plugins
   ├──/ java
   ├──/ maven
   ├──/ vscode
   └──/ vscode-extensions
```

New layout proposed by this issue:
```
/ DEVON_IDE_HOME
├──/ plugins
|  ├──/ eclipse
|  ├──/ intellij
|  └──/ vscode
└──/ software
   ├──/ az
   ├──/ cobigen-cli
   ├──/ dotnet
   ├──/ eclipse
   ├──/ extra
   ├──/ intellij
   ├──/ java
   ├──/ maven
   └──/ vscode
```

This story affects all commandlets that can have plugins:
* [android-studio](https://github.com/devonfw/ide/pull/1136)
* [eclipse](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/eclipse)
* [intellij](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/intellij)
* [vscode](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/vscode)

maybeec:
I would not agree to that. What is the benefit? You might even run into incompatibility issues if someone does a major upgrade of an ide causing the ide not to start anymore.

hohwille:
> I would not agree to that. What is the benefit? You might even run into incompatibility issues if someone does a major upgrade of an ide causing the ide not to start anymore.

Thanks for your feedback, I try to explain more... not part of this story but for the rationale:

In the future we will enable [shared repository](https://github.com/devonfw/ide/blob/master/documentation/software.asciidoc#shared) by default (for the real solution of #1082).
Then `software` folder will only contain symbolic links to software tool-edition versions from the central shared software repository (except for the `extra` folder that we could also eliminate entirely or move from `software/extra` to `extra-software`).
[...]

PR #1179, opened 2023-06-05
#1147: separate plugins from software

jan-vcapgemini:
Addresses/Fixes: #1147 

Implements:

* changed plugin paths from software/x-plugin to plugins/x
* adjusted documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -6,7 +6,7 @@ This file documents all notable changes to https://github.com/devonfw/ide[devonf
 
 Release with new features and bugfixes:
 
-* https://github.com/devonfw/ide/issues/1158[#1158]: Projects main-branch is always checked out  
+* https://github.com/devonfw/ide/issues/1158[#1158]: Projects main-branch is always checked out
 * https://github.com/devonfw/ide/issues/1118[#1118]: Add milestone versions of Eclipse
 
 The full list of changes for this release can be found in https://github.com/devonfw/ide/milestone/44?closed=1[milestone 2023.05.001].
@@ -56,7 +56,7 @@ The full list of changes for this release can be found in https://github.com/dev
 Release with new features and bugfixes:
 
 * https://github.com/devonfw/ide/issues/1066[#1066]: devon.bat stopped adding PATH
-* https://github.com/devonfw/ide/issues/1043[#1043]: Environment variables defined in devon.properties are not available in devonfw CMD 
+* https://github.com/devonfw/ide/issues/1043[#1043]: Environment variables defined in devon.properties are not available in devonfw CMD
 * https://github.com/devonfw/ide/issues/982[#982]: Add plugin recommendation feature to vscode
 * https://github.com/devonfw/ide/issues/971[#971]: Ability to run "migration script" on update
 * https://github.com/devonfw/ide/issues/991[#991]: Integrate openrewrite feature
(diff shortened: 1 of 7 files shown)


CHANGELOG.asciidoc
```diff
@@ -165,7 +167,7 @@ New release with various improvements and bugfixes:
 * https://github.com/devonfw/ide/issues/692[#692]: Improve Windows Terminal with gitbash
 * https://github.com/devonfw/ide/issues/786[#786]: Improve Windows Terminal with powershell
 * https://github.com/devonfw/ide/issues/849[#849]: Integration of GCViewer
-* https://github.com/devonfw/ide/issues/867[#867]: ERROR: Error opening the file. There may be a disk or file system error.
+* https://github.com/devonfw/ide/issues/867[#867]: ERROR: Error opening the file, there may be a disk or file system error
```
hohwille, line 170: are you using auto-magic AsciiDoc plugins that try to do "auto-correction".
This is the summary of an issue and it comes from an actually error message that is just the way this error message is.
Chaning this title here does not seem to make sense.
Tame your IDE :)

hohwille (commented):
@jan-vcapgemini thanks for your PR. Even though still in draft state this looks fine for me. 👍 
We should then also update #1135 accordingly.
For me it makes sense to ship this change with the next release that brings more of such structural changes. Further we should also add this to our migration script (see #1123).
Could you include the migration script addition to this PR (and also have a look at my review comment)? Thanks

CHANGELOG.asciidoc
jan-vcapgemini: I've tried to manually fix the conflict between end of lines with a dot and a required newline.
It's reverted now.

jan-vcapgemini:
> @jan-vcapgemini thanks for your PR. Even though still in draft state this looks fine for me. 👍 We should then also update #1135 accordingly. For me it makes sense to ship this change with the next release that brings more of such structural changes. Further we should also add this to our migration script (see #1123). Could you include the migration script addition to this PR (and also have a look at my review comment)? Thanks

I've adjusted the CHANGELOG.asciidoc and migration script now.
I was waiting for the changes of: #1135, but we can also update the paths there too.

scripts/src/main/resources/scripts/migrations/2023.06.001
```diff
@@ -47,3 +47,22 @@ find "${DEVON_IDE_HOME}"/workspaces/ -mindepth 2 -maxdepth 2 -name devon.propert
...
+plugins_path="${DEVON_IDE_HOME}/plugins"
+
+if [ ! -d "${plugins_path}" ]
+then
+	mkdir "${plugins_path}"
+else
+	doWarning "Plugins directory already exists!"
+fi
+
+find "${DEVON_IDE_HOME}"/software -maxdepth 1 -type d -name "*-plugins" -print | while read found_folder
```
hohwille, line 60: Great that you added the migration. 👍 
As we only have 2 specific candidates here (`vscode-extensions` and `eclipse-plugins`) it does not make much sense trying to solve this in a generic way esp. as `vscode-extensions` does not match `*-plugins`.
I would simply hardcode these two folders here. In the end all we need is just
```
mv "${DEVON_IDE_HOME}/software/vscode-extensions" "${DEVON_IDE_HOME}/plugins/vscode"
mv "${DEVON_IDE_HOME}/software/intellij-plugins" "${DEVON_IDE_HOME}/plugins/intellij"
```

Android-studio has not yet been released and eclipse is not yet separated so KISS is key.

hohwille: Feel free to check the directories for existence to suppress `mv: cannot stat 'software/vscode-extensions': No such file or directory` error.

jan-vcapgemini: I've applied KISS now and added checks for existing directories.

hohwille (approved):
@jan-vcapgemini excellent. Thanks for the update 👍 
Ready for merge.

Issue #1176, opened 2023-06-01
Add updater for GCloud


alfeilex:
GCloud is already implemented as a commandlet in IDE but it is not present in the URL-Updater. The aim of this issue would be to create an updater for gcloud. 

A.C:

- [x] Find out how to get versions
- [x] Find a valid download link
- [x] Implement `GcloudUrlUpdater.java` in `..\url-updater\src\main\java\com\devonfw\tools\ide\url`
- [x] Update `UpdateManager.java`
- [x] Test the updater run

PS: Gcloud is currently only available for Windows

PR #1185, opened 2023-06-07
#1176 : Add updater for GCloud 

alfeilex:
This PR adds updater for GCloud. 

Related issue: #1176


url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+    return "gcloud";
+  }
+
+  @Override
+  protected void addVersion(UrlVersion urlVersion) {
+
+    String baseUrl = "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${version}-${os}-";
+
+    doAddVersion(urlVersion, baseUrl + "x86_64.zip", WINDOWS);
+    //TODO add Linux and MacOS
```
alfeilex, line 22: We do not provide `gcloud` for Linux and MacOS. The reason for that is that we do not provide python for these OS's and that is a requirement for `gcloud.` Should we still add these with the hope of fixing Python on these?

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  protected Pattern getVersionPattern() {
+
+    return Pattern.compile("(\\d+\\.\\d+\\.\\d+)");
+  }
+
+  @Override
+  protected String mapVersion(String version) {
+
+    if (version.matches("\\d{3,}\\.\\d+\\.\\d+")) {
```
alfeilex, line 40: There are many releases for `gcloud`. In my opinion we should just provide versions higher `100.x.x` or even `200.x.x`. We could also use the regex on the `getVersionpattern()` function and do without mapping

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
hohwille: Yes. If a version is added to ide-urls with only windows and we later change the code to also create linux and mac URLs, those will not be added without deleting the old versions. So we should already add all OS from the start.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
hohwille: In such case, it would be better to require a minimum version. See example here:
https://github.com/devonfw/ide/blob/master/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/node/NodeUrlUpdater.java#L48

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+  @Override
+  protected String getTool() {
+
+    return "gcloud";
+  }
+
+  @Override
+  protected void addVersion(UrlVersion urlVersion) {
+
+    String baseUrl = "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${version}-${os}-";
```
hohwille, line 19: either declare as local variable or use a constant (`private static final String BASE_URL = ...`).

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+
+    doAddVersion(urlVersion, baseUrl + "x86_64.zip", WINDOWS);
+    //TODO add Linux and MacOS
+  }
+
+  @Override
+  protected String getVersionUrl() {
+
+    return "https://cloud.google.com/sdk/docs/release-notes";
```
hohwille, line 28: our goal is to get rid of `WebsiteUrlUpdater` and only use it as last resort.
Please consider using `GitHubUrlUpdater` instead:
https://github.com/twistedpair/google-cloud-sdk/tags

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
alfeilex: What do you mean by local variable? Is it not already declared locally?

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
alfeilex: Done

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gcloud/GCloudUrlUpdater.java
alfeilex: Done

hohwille (approved):
@alfeilex thanks. Looks perfect now 👍

PR #1203, opened 2023-06-20
#1162: added timeout for GitHub actions

jan-vcapgemini:
Addresses/Fixes: #1162

Implements:
* added a timeout handling to UpdateInitiator, UpdateManager and AbstractUrlUpdater


url-updater/src/main/java/com/devonfw/tools/ide/url/UpdateInitiator.java
```diff
@@ -15,22 +17,42 @@ public class UpdateInitiator {
...
     String pathToRepo = args[0];
+    long timeout;
+    Duration expirationTimeInMillis = null;
+
+    if (args.length < 2) {
+      logger.warn("Timeout was not set, setting timeout to infinite instead.");
+    } else {
+      try {
+        timeout = Long.parseLong(args[1]);
+        expirationTimeInMillis = Duration.ofMillis(System.currentTimeMillis() + (timeout * 1000));
```
hohwille, line 40: Sorry, for the misunderstanding: I tried to say that a real duration should be stored as java.time.Duration instead of long (before you hard a `long timeout` what was a duration in seconds and therefore it is better to use `Duration.ofSeconds(parsedLong)`).
However, if you store the date and time that will abort further processing when exceeded this is still a date+time information rather than a duration (a duration is more a delta of two date/time values). You can either keep it easy and stay with `long` and use `System.currentTimeMillis()` what is most efficient or you use `java.time.Instant` if you want to make it most expressive. In the latter case use `Instant.now()` instead of `System.currentTimeMillis()` and make use of methods like `plus(Duration)` and `isAfter(Instant)`... Again sorry, if I was not clear on this in my review suggestion today.

hohwille: For the record; `Duration` even has a parse method. The syntax is standardized by some cross-language ISO format (that might seem a little odd but is quite flexible if you want to specify a real duration as argument instead of just a number:
https://www.geeksforgeeks.org/duration-parsecharsequence-method-in-java-with-examples/

jan-vcapgemini: Adjusted now.

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
```diff
@@ -431,12 +442,30 @@ private void doUpdateStatusJson(boolean success, int statusCode, UrlVersion urlV
...
+
+    return false;
+  }
+
   /**
    * Updates the tool's versions in the URL repository.
    *
    * @param urlRepository the {@link UrlRepository} to update
    */
-  @Override
```
hohwille, line 439: why did you remove this override? The method is still overridden. The annotation is used to ensure this.

hohwille (approved):
@jan-vcapgemini thanks for your PR. Great.
Ready for merge

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/AbstractUrlUpdater.java
jan-vcapgemini: Was by mistake. Re-added it now.

Issue #1184, opened 2023-06-07
correct python configuration in ide-urls


CREITZ25:
Python is configured as system independent in IDE-URLS. In fact, the packages are system-dependent. Instead of the urls file, the files linux_x64.urls, mac_x64.urls and windows_x64.urls and, if necessary, the corresponding files for the arm64 architecture must be created.

PR #1206, opened 2023-06-22
 #1184: Correct python configuration in ide-urls

alfeilex:
- switch from website url updater to github url updater
- add urls for mac and linux

Related issue: #1184


hohwille (commented):
@alfeilex looks already good to me 👍 
Are you still testing or why is this in draft?

alfeilex:
> Are you still testing or why is this in draft?

The current problem is that python is installed on windows with embed files. These files are not present in every version and so there are no `windows.urls` provided for these versions.

hohwille (approved):
@alfeilex Great. Thanks for this PR. Looks good to me 👍

PR #1209, opened 2023-06-23
fix gcloud toolname

alfeilex:
fix gcloud tool name


hohwille (approved):
@alfeilex Thanks for this fix. 👍  I should have seen it in the review ;)

Issue #1216, opened 2023-06-26
Refactor checksum verification


alfeilex:
There is an problem on verify checksums on Windows since the command `shasum -a 256` is not available there. 

on windows and linux:
`sha256sum`

on macos:
`shasum -a 256`

PR #1217, opened 2023-06-26
#1216: Refactoring checksum verification

alfeilex:
A new function for checking the checksum is created. Further the sha256 generation is fixed for each os


hohwille (approved):
@alfeilex thanks for your analysis and fix of the checksum issue. Also great that you found that the code is redundant and extracted a common function. 👍
You accidentally added 3 yaml files. Can you delete them and then move out of draft mode so that I can merge?

hohwille:
PR was replaced by #1220

PR #1221, opened 2023-06-28
Fix SONAR_HOME path variable for Windows

alfeilex:
This PR is a quick fix for a path variable in `sonar` command script

https://github.com/devonfw/ide/blob/015e8fae3c6a0c474f6064c708b8a89685b45392/scripts/src/main/resources/scripts/command/sonar#L42-L45


hohwille (approved):
OK, now I got it. In our script we already changed the variable to `SONAR_HOME` but forgot to also update other occurrences referencing the variable.

Thanks for fixing

Issue #1218, opened 2023-06-27
gradle download broken


hohwille:
It seems that gradle downloads got broken - maybe the URLs have been changed by the tool vendor.
Our github action already noted that the downloads broken immediately after the metadata was created as you can see in the status.json files.
It seems we have to fix the GradleUrlUpdater and also delete the entire gradle folder from ide-urls so it will be recreated in the night.

p.s.: I hope such effects will not recur often as this sucks and causes us maintenance effort that we actually do not want to have. The idea of #941 was actually to reduce the maintenance effort as much as possible.

hohwille:
Example for current download link leading to error:
https://services.gradle.org/distributions/gradle-9.95-bin.zip

hohwille:
It seems to be different. The download link structure is OK.
https://github.com/devonfw/ide-urls/blob/master/gradle/gradle/8.1.1/urls
https://services.gradle.org/distributions/gradle-8.1.1-bin.zip works

As you can see on the [download website of gradle](https://gradle.org/releases/) there are no 9.x release versions listed. So just tracking git tags does not seem to be the proper approach here. Still I guess that deleting the gradle folder and rebuilding it over night will not rebuild the broken releases.
My assumption is that the broken versions have been created from the very old code of url-updater that did not follow redirects. So as a first step, I will simply delete gradle folder.

hohwille:
Even the latest official version failed:
```
$ devon gradle setup
Updating gradle from version 8.0.2 to version 8.1.1...
Success: run command mkdir
Starting installation of gradle in version 8.1.1 to /d/projects/test/software/gradle
Deleting old installation as it is a symlink: /d/projects/test/software/gradle
Trying to download gradle-8.1.1.zip from https://services.gradle.org/distributions/gradle-8.1.1-bin.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
 18  118M   18 22.3M    0     0   188k      0  0:10:44  0:02:01  0:08:43     0
curl: (56) Recv failure: Connection was reset
Failed to download https://services.gradle.org/distributions/gradle-8.1.1-bin.zip with exit code 56
```

hohwille:
This is nothing we can change. The vendor of gradle has to provide a reliable service for downloading its tool.

jan-vcapgemini:
[...]

PR #1222, opened 2023-06-29
#1218: optimized gradle url-updater

jan-vcapgemini:
Addresses/Fixes: #1218 

Implements:
* made version regex pattern more distinct
* added hashsum param to doAddVersion
* added doGetHashSumForVersion method (can extract a valid hashsum for a given version)
* moved regex patterns and URLs into constants
* added lazy loading of responseBody
* added extra check for empty checksum
* moved download Url into constant

### DIFF
--- a/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gradle/GradleUrlUpdater.java
+++ b/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gradle/GradleUrlUpdater.java
@@ -22,6 +22,10 @@ public class GradleUrlUpdater extends WebsiteUrlUpdater {
 
   private final static String HASH_VERSION_URL = "https://gradle.org/release-checksums";
 
+  private final static String DOWNLOAD_URL = "https://services.gradle.org/distributions/gradle-${version}-bin.zip";
+
+  private String responseBody;
+
   @Override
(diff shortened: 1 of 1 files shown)


url-updater/src/main/java/com/devonfw/tools/ide/url/updater/gradle/GradleUrlUpdater.java
```diff
@@ -30,8 +43,52 @@ protected String getTool() {
   @Override
   protected void addVersion(UrlVersion urlVersion) {
 
-    doAddVersion(urlVersion, "https://services.gradle.org/distributions/gradle-${version}-bin.zip");
+    String responseBody = doGetResponseBodyAsString(HASH_VERSION_URL);
```
hohwille, line 46: functionally correct. But we are requesting this HTML for every version what seems to be waste.
Can we do that only once? The easiest way would be to use a lazy getter with a member variable for the HTML (currently `responseBody`).

jan-vcapgemini: I've adjusted it now.

hohwille (approved):
@jan-vcapgemini Excellent. Thanks for the final polishing. Now everything is perfect.

PR #1234, opened 2023-07-05
Quick-Fix: Remove python and pip integration-test

alfeilex:
Since we can not offer a stable python solution, this PR removes the integration-tests for python and pip.


hohwille (approved):
@alfeilex thanks for creating the PR. 👍 
I hope we can stabilize python & co to later turn this on again...

Issue #1084, opened 2023-03-28
Adapt scripts to use ide-urls


hohwille:
For epic #941 we also need to adapt the devonfw-ide scripts itself such that it uses `ide-urls` instead of `ide-mirrors`.
This story is about all the required changes to the scripts:
* [x] adapt `doDownload` function and all related functions to use `ide-urls` with the new layout. We can drop all variable replacements and do a lot of simplification but also need to verify checksums (see #1065).
* [x] also functions to list versions and determine the latest version have to be rewritten (`available-versions` does not exist anymore and a new challenge is that we have to be able to sort the versions properly that we get from the list of version folders and for some tools like Java there are some specific sorting challenges).
* [x] adapt commandlets (at least `mvn`, `node`, `cobigen`, `docker`, `intellij`, `eclipse`, and `java`) to fit to the new structure - so remove all functions `doResolveDownloadUrl` (see `java`), and change additions according to #1087. See also #1085
* [x] define the following defaults for variables ~~(unfortunately we can not do this in our own [devon.properties](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/devon.properties) as then it can not be overridden in `~/devon.properties` according to the [configuration order](https://github.com/devonfw/ide/blob/master/documentation/configuration.asciidoc)~~:
```
DOCKER_EDITION=rancher
```

ATTENTION: This story depends on story #1087 to be implemented/merged before this story to work properly.

Final acceptance criteria is that all tools in all editions with all versions can still be downloaded and installed from `urls` after removing `mirrors`.

hohwille:
github feature to link issues with PRs is still buggy and partially broken. I have to reopen to connect with PR.

PR #1237, opened 2023-07-06
#1084: omit default edition in the download filename at pip

CREITZ25:
Addition to https://github.com/devonfw/ide/issues/1084:
Omit the default edition to keep behavior of previous releases of devonfw-ide.
Small change, big impact: This allows to reuse the download cache and prevents that devonfw-ide will download all the existing files again and create duplicates on the disc.

This is coded in pip commandlet, and must be fixed, too


hohwille (approved):
@CREITZ25 Aweseome. I missed this out. Thanks for house-keeping. 👍

Issue #1245, opened 2023-07-10
tomcat is not working on MacOSX


jan-vcapgemini:
### Expected behavior

As a devonfw-ide MacOS user I want to install tomcat, so that I can use it for my projects.

### Actual behavior
After installation following error occurs:

```
/Users/username/devonfw-test-vanilla/scripts/command/../functions: line 84: /Users/username/devonfw-test-vanilla/software/tomcat/bin/catalina.sh: Permission denied
Exit code was 126
```

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Run `devon tomcat version set 9.0.76`
2. run `devon tomcat setup`

### Comments/Hints:
Could possibly be fixed with setting the permissions inside the tomcat command script.

### Affected version:

* OS: Mac
* devonfw-ide: 2023.07.001-SNAPSHOT

hohwille:
I should have properly stopped merging new features like tomcat until we had completed our release.
Now, we should fix this before releasing... Pity.

hohwille:
Should be fixed after my commits and when the updater runs again tonight.

hohwille:
Seems to be OK, as I checked X-flags manually in tar.gz and URLs are updated.

PR #1249, opened 2023-07-10
#1245 tomcat is not working on MacOSX

CREITZ25:
#1245 tomcat is not working on MacOSX

### DIFF
--- a/scripts/src/main/resources/scripts/command/tomcat
+++ b/scripts/src/main/resources/scripts/command/tomcat
@@ -29,7 +29,7 @@ function doSetup() {
   doInstall "tomcat" "${TOMCAT_VERSION}" "${1}"
   if doIsMacOs
   then
-    find "${TOMCAT_HOME}"/bin -name *.sh -exec chmod +x {} \;
+    find "${TOMCAT_HOME}"/bin -name '*.sh' -exec chmod +x {} \;
   fi
   "${TOMCAT}" version
   return 0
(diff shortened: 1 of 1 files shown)


hohwille (commented):
@CREITZ25 Excellent that you analyzed the problem on MacOS and that you found a solution.
However, looking at this my first question is why would we have to do the homework of a software tool provider?
The answer seems to be that we just use the wrong packaging format:
https://github.com/devonfw/ide/blob/fb7be0434dd2889a5dab2edbb52bdb78d8f9236e/url-updater/src/main/java/com/devonfw/tools/ide/url/updater/tomcat/TomcatUrlUpdater.java#L22

ZIP files are too stupid and cannot express file permissions like executable flags.

https://github.com/devonfw/ide-urls/blob/8cd93670b7d747e641467a8480e761eac2c4ac50/tomcat/tomcat/9.0.76/urls#L1
contains this URL:
https://archive.apache.org/dist/tomcat/tomcat-9/v9.0.76/bin/apache-tomcat-9.0.76.zip

If I replace `zip` with `tar.gz`, I get a proper archive that has correct executable flags and we can skip such workaround and avoid hacks.
I know it is a pain how much effort it is to get our updater correct and it seems as if this is rather an expert domain that can not be easily mastered by developers without longer experience in devonfw-ide. Anyhow my suggestion would be to fix it at the root of the problem.
My first guess would be that we also have the same issue for linux and your PR is only fixing macos...

hohwille:
Can be closed. See review comment and issue #1245.

Issue #1197, opened 2023-06-16
Integrate Java Mission Control


hohwille:
Java Mission Control (jmc) is a tool to analyse performance and runtime data of Java applictions. It works perfect with Java Flight Recorder (a feature available in Java that can be activated via `-XX:StartFlightRecording`).
JMC does not come with Adoptium (OpenJDK) but is available here:
https://adoptium.net/jmc/

It would be awesome if devonfw-ide could support users to install and run it (`devon jmc`).

hohwille:
There is no jmc here: https://github.com/devonfw/ide-urls/tree/master/
We can not merge this without the download metadata being present.
https://github.com/devonfw/ide/pull/1251/files only added the `JmcUrlUpdater` without adding it so it also gets executed.

PR #1251, opened 2023-07-11
#1197 added UrlUpdater for java mission control

moritzLanger:
Both the official website and GitHub repo did not work with the updater, because at least one version would be missing. Sadly there wasn't a Json I could find but the Mirrored Repo of adoptium delivers every version for the GithubUrlUpdater.


url-updater/src/main/java/com/devonfw/tools/ide/url/updater/jmc/JmcUrlUpdater.java
```diff
@@ -0,0 +1,39 @@
...
+  @Override
+  protected String getGithubOrganization() {
+
+    return "adoptium";
+  }
+
+  @Override
+  protected String getGithubRepository() {
+
+    return "jmc-build";
```
hohwille, line 26: latest tag is  9.0.0-SNAPSHOT but there is no such release. Can you filter out SNAPSHOT versions?

hohwille (changes requested):
@moritzLanger thanks for your PR. Looks great.
Can you check my comment regarding SNAPSHOT versions. After all the cleanups, where I had to remove content from ide-urls manually, I want to get it right from the start...

url-updater/src/main/java/com/devonfw/tools/ide/url/updater/jmc/JmcUrlUpdater.java
moritzLanger: While testing locally the SNAPSHOT version already got filtered, or do you mean something else? 
image

hohwille: Thanks for checking. Resolved.

hohwille (approved):
My concern was wrong. All great 👍 
Ready for merge.

PR #1257, opened 2023-07-11
#1260: fix migration script 2023.07.001

### LINKED ISSUE #1260: Migration 2023.07.001 buggy
With #1087 we have created a migration script for `2023.07.001` release.
Unfortunately this was buggy.
E.g. it changed `INTELLIJ_EDITION_TYPE=U` to `INTELLIJ_TYPE=U` but correct would have been `INTELLIJ_TYPE=ultimate`.
Even worse it did replace every `u` (lower case) with `ultimate` so in my devon properties the java version `8u292b10` was changed to `8ultimate292b10` (including other occurrences like `u` in hostnames).

Usually we need to properly test a migration script as it cannot be changed or fixed after it did already run.
However, shit happens and as we noticed quickly after the release, the plan is to change the existing migration script and quickly push out another release `2023.07.002` in the hope that most users will skip the release `2023.07.001` and get the fixed migration script applied instead of the buggy one.

If we can adopt this very easily it would be an option to change the logic for `INTELLIJ_EDITION_TYPE` and `ECLIPSE_EDITION_TYPE` such that we do not replace this properties with the new ones, but instead add the new ones for better downward compatibility as described in #1242. This only applies for these two explicitly stated properties and nothing else in the existing migration.
So if we find a `devon.properties` containing one of these two properties, we would add the corresponding property without the `_TYPE` and the new value unless it is already present.
Example before migration:
```
INTELLIJ_EDITION_TYPE=U
ECLIPSE_EDITION_TYPE=java
```
And after the migration:
```
INTELLIJ_EDITION_TYPE=U
ECLIPSE_EDITION_TYPE=java
INTELLIJ_EDITION=ultimate
ECLIPSE_EDITION=eclipse
```
However, if the migration finds the latter file for whatever reason, it will not change by the migration - even if it is inconsistent like:
```
INTELLIJ_EDITION_TYPE=C
ECLIPSE_EDITION_TYPE=java
INTELLIJ_EDITION=ultimate
ECLIPSE_EDITION=cpp
```
Especially it would be a bug to produce something like this by the migration:
```
INTELLIJ_EDITION_TYPE=C
ECLIPSE_EDITION_TYPE=java
INTELLIJ_EDITION=ultimate
ECLIPSE_EDITION=cpp
INTELLIJ_EDITION=intellij
ECLIPSE_EDITION=eclipse
```

Important: Fixing this quickly wins. If we can not implement the additional requirement that I suggested here this week, it is better to only fix the real bug and do the release this week so less people end up with the mess produced by that bug.

CREITZ25:
Fixes #1260 
I just created and linked that issue that I also added to our CHANGELOG.


hohwille (commented):
@CREITZ25 thanks for your fast PR.
However, you did not get what I was suggesting.
We cannot easily undo the chaos created by the old migration script replacing every `u` with `ultimate`, etc.
So here you added a new migration as a corrected "copy" of the 2023.07.001 migration.

Instead of adding a new migration that will run after the old and broken one, we need to change the existing `2023.07.001` migration. This will not help for people who have already installed 2023.07.001 but it will help for all users that upgrade from <= 2023.04.001 starting from the day when we have released the fix.
If we hurry up, we can ship a 2023.07.002 release with no other change than this fix until Thursday or Friday this week what would help a lot.

CREITZ25:
> @CREITZ25 thanks for your fast PR. However, you did not get what I was suggesting. We cannot easily undo the chaos created by the old migration script replacing every `u` with `ultimate`, etc. So here you added a new migration as a corrected "copy" of the 2023.07.001 migration.
> 
> Instead of adding a new migration that will run after the old and broken one, we need to change the existing `2023.07.001` migration. This will not help for people who have already installed 2023.07.001 but it will help for all users that upgrade from <= 2023.04.001 starting from the day when we have released the fix. If we hurry up, we can ship a 2023.07.002 release with no other change than this fix until Thursday or Friday this week what would help a lot.

Yes, your are right. 
Now we have the corrected file 2023.07.001. 
And I added 2023.07.002 which repairs also errors caused by 2023.07.001.

hohwille (approved):
@CREITZ25 thanks for the excellent fix. 👍 
Ready for merge and release.

Issue #1259, opened 2023-07-11
Android-Studio does not separate plugins


hohwille:
As just described in #1147 the android-studio was added during #1147 was implemented and I did not notice that it does not follow the new requirement to separate plugins from the software. This requirement is needed for stories like #1178 and making the shared software repo the new default. Hence, this is actually a bug.

For implementation see PR #1179 - we have to do the analogue changes we had done for intellij as android-studio is more or less a fork or branch of intellij.

hohwille:
What even confused me is that I did install FolderCompact Plugin and it was installed directly inside the `software/android-studio/plugins` folder what causes the actual problem.
However, in the commandlet there is this code:
https://github.com/devonfw/ide/blob/7a41895b2fb9cadb08931b2bbbfc05c7880b00e9/scripts/src/main/resources/scripts/command/android-studio#L68

PR #1261, opened 2023-07-12
#1259: changed android-studio plugins directory

jan-vcapgemini:
Addresses/Fixes: #1259

Implements:
* changed android-studio plugins directory from software/android-studio-plugins to plugins/android-studio


hohwille (approved):
@jan-vcapgemini thanks for the fast fix. Looks good 👍

Issue #1258, opened 2023-07-11
Error opening android-studio: Could not find templates path


hohwille:
```
$ devon android-studio
Success: run command mkdir
Starting installation of android-studio in version 2023.1.1.11 to /d/projects/XXX/software/android-studio
Trying to download android-studio-2023.1.1.11-windows.zip from https://redirector.gvt1.com/edgedl/android/studio/ide-zips/2023.1.1.11/android-studio-2023.1.1.11-windows.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   541  100   541    0     0    959      0 --:--:-- --:--:-- --:--:--   962
100 1074M  100 1074M    0     0  3751k      0  0:04:53  0:04:53 --:--:-- 5342k
Download of android-studio-2023.1.1.11-windows.zip from https://redirector.gvt1.com/edgedl/android/studio/ide-zips/2023.1.1.11/android-studio-2023.1.1.11-windows.zip succeeded.
Checksum 77d50b0d982faedd98b73e08c922f4e20407a4a507eff2dcb3c6a1ce01b208ac is correct
Success: run command unzip
Successfully extracted archive android-studio-2023.1.1.11-windows.zip to /d/projects/XXX/updates/extracted/android-studio
Success: run command mv
Successfully installed android-studio
The software android-studio has been added. You need to rerun 'devon' command without arguments or restart your terminal to update your PATH so the newly installed software will be found.
{
  "name": "Android Studio",
  "version": "dev build",
  "versionSuffix": "EAP",
  "buildNumber": "231.9161.38.2311.__BUILD_NUMBER__",
  "productCode": "AI",
  "dataDirectoryName": "AndroidStudioPreview2023.1",
Success: verify installation of android-studio (head)

******** ATTENTION ********
Could not find templates path at /d/projects/XXX/settings/android-studio/workspace
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
```

We need to make our workspace update more robust. If there is no workspace configuration in settings, then the IDE (android-studio) should open with the default settings instead of raising an error and preventing the user to launch the IDE.

PR #1262, opened 2023-07-12
#1258: Fix error opening android-studio: Could not find templates path 

alfeilex:
This PR fixes the case that android studios is launched without existing `settings/android-studio/workspace` folder

Related PR: #1258


scripts/src/main/resources/scripts/command/android-studio
```diff
@@ -109,7 +109,13 @@ function doConfigureAndroidStudio() {
   fi
   if [ -n "${mode}" ]
   then
-    doConfigureWorkspace "${SETTINGS_PATH}/android-studio/workspace" "${WORKSPACE_PATH}" "${mode}"
+    local settings_workspace="${SETTINGS_PATH}/android-studio/workspace"
+    if [ -d "${settings_workspace}" ]
+    then
+      doConfigureWorkspace "${settings_workspace}" "${WORKSPACE_PATH}" "${mode}"
+    else
+      doWarning "There is no workspace folder in ${settings_workspace}. Launch Android Studio with default settings..."
```
alfeilex, line 117: In case the folder `settings/android-studio/workspace` is deleted after the first launch of android studio, the tool will be started with settings defined in `workspaces/.../.android-studio` folder, if it is exists. So the the warning that android-studio will launch with default settings is not correct, is it? Should we just change the warning or add another check?

hohwille: I think this is fine. It is an odd edge-case. Technically what we want to say is that we do not manipulate/manage the settings and that is correct. If you want you can change the wording of the message but we should not change the logic of the code.

hohwille (approved):
@alfeilex thanks for the fast fix/PR. Looks good to me 👍 I just improved the warning message after reading your comment.  Now ready for merge.

Issue #1263, opened 2023-07-13
In Release 2023.07.001 the file DEVON_IDE_HOME/.devon.software.version will not created with setup


CREITZ25:
You have installed IDE Release 2023.04.001 and want to update your scripts with the command devon ide update scripts.
In that case the migration-script will be processed and the file .devon.software.version will be created in ${DEVON_IDE_HOME}.

If you install 2023.07.001 directly with the setup command, this file won't be created.

We have to create this file in this case too, to prevent executing unnecessary old migration scripts

PR #1266, opened 2023-07-13
#1263 devon.software.version will not be created with setup or setup.bat

CREITZ25:
#1263 devon.software.version will not be created with setup or setup.bat


scripts/src/main/resources/scripts/command/ide
```diff
@@ -575,4 +575,6 @@ else
 fi
 
 echo
+# shellcheck disable=SC2007
+echo "$[devon_ide_version]" > "${DEVON_IDE_HOME}"/.devon.software.version
```
hohwille, line 579: This has to go to the end of the `doSetup()` function.
Otherwise it will always overwrite the version into that file whenever the `ide` commandlet is called.

hohwille (changes requested):
@CREITZ25 your `echo` line is the correct fix for the issue.
However, IMHO it has to be moved to `doSetup()` function to avoid undesired side-effects.

scripts/src/main/resources/scripts/command/ide
CREITZ25: Of course, done

CREITZ25 (commented):
done

hohwille (approved):
@CREITZ25 excellent. That was fast. Thanks. 👍

CREITZ25 (commented):
done

Issue #1246, opened 2023-07-10
yarn is not installing properly on MacOSX and Windows


jan-vcapgemini:
### Expected behavior

As a devonfw-ide Windows and MacOS user I want to install yarn, so that I can use it for my projects.

### Actual behavior
devon yarn is not installed automatically (just with devon yarn setup)

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1. Run `devon yarn`

### Affected version:

* OS: Mac, Windows
* devonfw-ide: 2023.07.001-SNAPSHOT

hohwille:
Please try to include logs, as I could not get what the actual problem was and could not reproduce the bug in the first place.
So here is my log:
```
hohwille@CE49454 MSYS /d/projects/test2
$ devon yarn
Running: yarn install
/d/projects/test2/scripts/command/yarn: line 84: yarn: command not found
sed: can't read package.json: No such file or directory
No build script is present in package.json - skipping to run build script.

hohwille@CE49454 MSYS /d/projects/test2
$ devon yarn setup
Running: npm install -g yarn@latest

added 1 package in 1s

hohwille@CE49454 MSYS /d/projects/test2
$ devon yarn
Running: yarn install
yarn install v1.22.19
info No lockfile found.
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
success Saved lockfile.
Done in 0.05s.
sed: can't read package.json: No such file or directory
No build script is present in package.json - skipping to run build script.
```

hohwille:
I am slightly confused why this is happening.
If you call `devon yarn` with no argument, it will do this:
https://github.com/devonfw/ide/blob/a66c6261fd7c144adedf264edbe97d96775b8bff/scripts/src/main/resources/scripts/command/yarn#L118

This is calling the following:
https://github.com/devonfw/ide/blob/a66c6261fd7c144adedf264edbe97d96775b8bff/scripts/src/main/resources/scripts/command/yarn#L89

And that first calls this:
https://github.com/devonfw/ide/blob/a66c6261fd7c144adedf264edbe97d96775b8bff/scripts/src/main/resources/scripts/command/yarn#L78

The actual bug seems to come from this condition:
https://github.com/devonfw/ide/blob/a66c6261fd7c144adedf264edbe97d96775b8bff/scripts/src/main/resources/scripts/command/yarn#L34

[...]

PR #1270, opened 2023-07-14
#1246: Fix yarn installation

alfeilex:
This PR fixes the installation of yarn

Related issue: #1246


hohwille (approved):
@alfeilex Excellent. Thank you for the fix! 👍 
Ready for merge.

scripts/src/main/resources/scripts/command/yarn
```diff
@@ -31,7 +31,7 @@ function doSetup() {
     then
       doEcho "Version ${installed_version} of yarn is already installed at $(command -v yarn)"
     fi
-  elif [ "${1}" != "silent" ] || [ "${version}" != "latest" ]
+  elif [ "${version}" != "latest" ] || [ -z "${installed_version}" ]
```
hohwille, line 34: Why do we have this `[ "${version}" != "latest" ]` here?
If `[ -z "${installed_version}" ]` makes perfectly sense as then there is no yarn installed so we have to install it.
However, if `${version}"` is `latest` then we technically do not know if the installation is up-to-date or not.
IMHO the intention of the old `if` condition was to omit this re-installation of `latest` version in `silent` mode but still have it on an explicit call of `devon yarn setup . This does not work anymore after this change so IMHO it is still incorrect.

So should we also add `|| [ "${1}" != "silent" ]`?
I think then we would have all cases covered or am I missing something?

hohwille: I simply edited this so this is resolved.

PR #1272, opened 2023-07-18
#1144: fixed maven installation bug

### LINKED ISSUE #1144: Maven ide tests affect devon-ide installations
### Expected behavior

As a devonfw-ide developer, I want to run the mvn tests on my machine within an isolated environment of devonfw-ide.

### Actual behavior
Currently the tests overwrite my `devon` command inside my $USER_HOME/.devon folder.

If I have introduced an error within my `devon` command script all of my devon-ide's will be affected.

1. Run `devon mvn clean install` or `devon build` inside the root of the **ide** project
2. Check your $USER_HOME/.devon/devon script

### Comments/Hints:
This part of the `devon` script seems to be the cause:
https://github.com/devonfw/ide/blob/fa4de78f2aaca0d48b5e36f3d7b7e279e230cf9e/scripts/src/main/resources/scripts/devon#L73-L78

### Affected version:

* OS: Windows

NicolasHauf:
This pull request fixes a bug where a new maven installation overwrites a file in the users home folder.
Closes #1144

### DIFF
--- a/scripts/src/main/resources/scripts/command/ide
+++ b/scripts/src/main/resources/scripts/command/ide
@@ -49,13 +49,15 @@ function doUpdateScripts() {
   doUpgradeMavenArtifact "${DEVON_IDE_HOME}" "${DEVON_IDE_REPO_URL}" "devonfw-ide-scripts" "${target_version}" ".tar.gz" "$[devon_ide_version]"
   if [ "${?}" = 0 ]
   then
-    if [ "${target_version}" = "LATEST" ]
+    doVersionCompare "${my_ide_version}" "${current_devon_version}"
+    if [ "${?}" = 1 ]
     then
       cp "${DEVON_IDE_HOME}/scripts/devon" ~/.devon/devon
     fi
     if doIsWindows
     then
-      if [ "${target_version}" = "LATEST" ]
+    doVersionCompare "${my_ide_version}" "${current_devon_version}"
+    if [ "${?}" = 1 ]
       then
         cp "${DEVON_IDE_HOME}/scripts/devon.bat" "${USERPROFILE}/scripts/devon.bat"
       fi


scripts/src/main/resources/scripts/command/ide
```diff
@@ -49,13 +49,15 @@ function doUpdateScripts() {
   doUpgradeMavenArtifact "${DEVON_IDE_HOME}" "${DEVON_IDE_REPO_URL}" "devonfw-ide-scripts" "${target_version}" ".tar.gz" "$[devon_ide_version]"
   if [ "${?}" = 0 ]
   then
-    if [ "${target_version}" = "LATEST" ]
+    doVersionCompare "${my_ide_version}" "${current_devon_version}"
```
hohwille, line 52: `my_ide_version` is a local variable of another function (`doUpdateCliScript`) so here it will always be empty.
IMHO all this code here is a broken duplication of existing code.
We can replace it with just a call of `doUpdateCli` and should be done.

hohwille (changes requested):
@NicolasHauf thanks for your PR. You found the right spot. IMHO with your change it will never update. Can you have a look at my review comments and consider a rework?
Please also sign our CLA as otherwise I cannot merge any of your contribution.

scripts/src/main/resources/scripts/command/ide
```diff
@@ -44,27 +44,6 @@ function doUpdateMetaFile() {
...
-    then
-      if [ "${target_version}" = "LATEST" ]
-      then
-        cp "${DEVON_IDE_HOME}/scripts/devon.bat" "${USERPROFILE}/scripts/devon.bat"
-      fi
-      doWarning "To prevent windows file locking errors, we update async in background now.\nPlease wait a minute and then rerun 'devon ide update' to complete."
-      exit 0
-    fi
-  fi
-}
```
hohwille, line 66: This function is called in multiple places.
As you removed it, you broke the code as there are remaining places where it is still called that will stop working.

hohwille (changes requested):
@NicolasHauf thanks for your rework. The way you fixed it should now work.
However, you broke the code as the function that you removed is still called in another place.

hohwille (changes requested):
Thanks for your rework. The change looks good now from the code PoV. I would have expected that this would solve our problem.

However, I have just pulled your changes locally and retested the bug but it is still present.
Did you ever reproduce the bug locally? Then you can more easily analyze what is going on and find a real fix for it.

p.s.: If you want to work on a different story and hand this issue over to somebody with deeper insights of our bash magic, feel free to go for it (e.g. announce in daily and remove yourself from assignee of the issue).

hohwille:
As discussed the problem is not as simple as I assumed and it is rather tricky to debug and trace down.
Hence, I am closing this PR so someone else can take over the story.

Issue #1164, opened 2023-05-16
devon jenkins stop fails


hohwille:
```
$ devon jenkins stop
java -jar /d/projects/test/software/jenkins/jenkins-cli.jar -s http://localhost:9999 -auth admin:admin shutdown
Error: Unable to access jarfile D:/projects/test/software/jenkins/jenkins-cli.jar
```

The `jenkins-cli.jar` is missing in `software/jenkins`.
However, see this:
https://github.com/devonfw/ide/blob/eed1f40356b8114358a74b6ab34c6b2ddb326726/scripts/src/main/resources/scripts/command/jenkins#L34
and:
https://github.com/devonfw/ide/blob/eed1f40356b8114358a74b6ab34c6b2ddb326726/scripts/src/main/resources/scripts/command/jenkins#L37

hohwille:
Indeed the `jenkins-cli.jar` that was present in `jenkins.war/WEB-INF` is not present anymore in recent versions.
I can not spot it in the WAR anymore (also not in `WEB-INF/lib`.

PR #1281, opened 2023-07-25
#1164: fix jenkins stop command

jan-vcapgemini:
Adresses/Fixes: #1164 

Implements:
* added new JENKINS_HOME environment variable to devon.properties
* renamed JENKINS_HOME in jenkins script to JENKINS_CLI_HOME
* added jenkins home folder creation to doSetup
* added override of java user.home to each java command of jenkins
* added first version of admin password retrieval
* added new param target_filename to doInstall
* added extra condition to doInstall to provide a filename for doDownload
* replaced doDownload in jenkins tool command with doInstall
* adjusted extraction of jenkins-cli.jar in doSetup of jenkins (jenkins-cli.jar was removed from WEB_INF folder in later jenkins versions)
* replaced password with auth file

### DIFF
--- a/scripts/src/main/resources/scripts/command/jenkins
+++ b/scripts/src/main/resources/scripts/command/jenkins
@@ -67,10 +67,12 @@ function doJenkinsCli() {
   then
     admin_password="$(cat "${JENKINS_HOME}"/secrets/initialAdminPassword)"
   fi
+  local auth_file="${JENKINS_HOME}/auth_file"
+  echo "admin:${admin_password}" > "${JENKINS_HOME}/auth_file"
   local jenkins_cli_command="java -Duser.home=${JENKINS_HOME} -jar '${JENKINS_CLI_HOME}/jenkins-cli.jar'"
-  echo "${jenkins_cli_command}" -s "http://localhost:${JENKINS_PORT:-9999}" -auth "${JENKINS_ADMIN_LOGIN:-admin}:${JENKINS_ADMIN_PASSWORD:-${admin_password}}" "${@}"
+  echo "${jenkins_cli_command}" -s "http://localhost:${JENKINS_PORT:-9999}" -auth "@${auth_file}" "${@}"
   # shellcheck disable=SC2145
-  doRunCommand "${jenkins_cli_command} -s 'http://localhost:${JENKINS_PORT:-9999}' -auth '${JENKINS_ADMIN_LOGIN:-admin}:${JENKINS_ADMIN_PASSWORD:-${admin_password}}' ${@}"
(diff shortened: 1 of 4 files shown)


scripts/src/main/resources/scripts/command/jenkins
```diff
@@ -12,48 +12,67 @@ fi
 
 # shellcheck source=scripts/functions
 source "$(dirname "${0}")"/../functions
-JENKINS_HOME="${DEVON_IDE_HOME}/software/jenkins"
-TOOL_VERSION_COMMAND="java -jar '${JENKINS_HOME}/jenkins.war' --version"
+JENKINS_CLI_HOME="${DEVON_IDE_HOME}/software/jenkins"
```
hohwille, line 15: mhm, this `«TOOL»_HOME` environment variables seems to cause some confusion.
Typically this refers to the installation folder of that tool.
This applies for `JAVA_HOME`, `MAVEN_HOME`, `GRADLE_HOME`, etc.
For CobiGen we seemed to have the first excuse where somebody (I guess you @jan-vcapgemini) explained that CobiGen treats `COBIGEN_HOME` as the directory where the user specific configuration for CobiGen is stored that is usually located somewhere in the users home directory.
I still think that such things are odd and it would be better if all tools would follow a common scheme how to name environment variables (and I am not saying that `«TOOL»_HOME` was really a great choice but it became more or less a de-facto standard and not following such de-facto standard is typically not great).
However, I did some research and found [this one link](https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/Jenkins-Home-Directory-Location-Change-JENKINS_HOME-windows-linux-ubuntu) that says you are absolutely correct that `JENKINS_HOME` means the `.jenkins` folder that is by default in the users home directory. On the other hand `JENKINS_CLI_HOME` indicates to be the directory where the `jenkins-cli` is "at home". However, the main thing here is the `jenkins.war` file and if something represents jenkins then it is this war file.
Anyhow, I have no better clue so we should just follow your suggestion and rename it this way.
(I guess when we go to Java such internal variables (e.g. `JENKINS_CLI_HOME` will disappear anyhow.)

hohwille (approved):
@jan-vcapgemini Thanks for your PR. This is nice work and will fix the issue. Also you aim for clean code e.g. by reusing `JENKINS_COMMAND`. 👍 
My comment is just to document my thoughts but nothing to change.

CREITZ25:
@jan-vcapgemini  function doInstall has a new parameter, this should also documented in function.asciidoc

CREITZ25 (commented):
@jan-vcapgemini function doInstall has a new parameter, this should also documented in function.asciidoc

PR #1291, opened 2023-07-26
#1197 Integrated JavaMissionControl

moritzLanger:
#1197 Integrate JavaMissionControl

Added the JmcUrlUpdater to UpdateManger function.

Added the jmc script.

Testing on a "real" Linux Computer still needs to be Done.


tobka777 (changes requested):
Thank you for your PR. Can you please still go through the [DoD](https://github.com/devonfw/ide/blob/master/documentation/DoD.asciidoc) and add the missing files like the documentation of the tool or LICENCE if needed.

moritzLanger:
> Thank you for your PR. Can you please still go through the [DoD](https://github.com/devonfw/ide/blob/master/documentation/DoD.asciidoc) and add the missing files like the documentation of the tool or LICENCE if needed.

Thanks for the reminder, I added the missing documentation.

scripts/src/main/resources/scripts/command/jmc
```diff
@@ -0,0 +1,55 @@
...
+  then
+    doSetup silent
+  fi
+  doRunCommand "${JMC_HOME}/jmc ${*} &"
+}
+
+# JDK Mission Control installation
+function doSetup() {
+  local install_dir="${DEVON_IDE_HOME}/updates/install/jmc"
+  doInstall "jmc" "${JMC_VERSION}" "${1}" "" "${install_dir}"
```
hohwille, line 31: why dont you use `${JMC_HOME}` instead of `${install_dir}`?
You are installing to a "temporary" directory and after that you are moving individual files from there to the final directory. The inner structure of the JMC download archive may change in the future what will then break this code.
If we want to have the `JDK Mission Control` content on the toplevel directory so the PATH gets computed correctly, you could afterwards move all the contents of the `JDK Mission Control` folder (if it exists) to its parent (`${JMC_HOME}`).
So if one day they change their download archive to a more common structure (e.g. with toplevel `bin` folder), our code will still work. Also you do not have to move `legal` and `.devon.software.version` then.

hohwille (changes requested):
@moritzLanger Thanks for your PR. All looks fully functional and complete. Great work! 👍 
I still do have one suggestion for improvement of maintainability. Could you have a look?

scripts/src/main/resources/scripts/command/jmc
moritzLanger: Thank you for the suggestion, I initially build it that way, as it was done so in other scripts, I changed it accordingly now and it works the same way

scripts/src/main/resources/scripts/command/jmc
```diff
@@ -0,0 +1,51 @@
...
+  then
+    doSetup silent
+  fi
+  doRunCommand "${JMC_HOME}/jmc ${*} &"
+}
+
+# JDK Mission Control installation
+function doSetup() {
+  doInstall "jmc" "${JMC_VERSION}" "${1}" "" "${JMC_HOME}"
+  doRunCommand "mkdir -p ${JMC_HOME}"
```
hohwille, line 31: this is kind of pointless

scripts/src/main/resources/scripts/command/jmc
```diff
@@ -0,0 +1,51 @@
...
+  doRunCommand "mv -f '${JMC_HOME}/JDK Mission Control/' ${JMC_HOME}"
```
hohwille, line 32: this will move/rename the folder `JDK Mission Control` in `${JMC_HOME}` itself.
I expected it to do nothing but as I tested it, it renamed the folder from `JDK Mission Control` to `'JDK Mission Control'`.
You instead need to move the contents of that folder.

hohwille (approved):
@moritzLanger thanks for your rework. 👍 
IMHO it is now broken. I added some review comments.
However, to avoid further ping-pongs, I go for a constructive review and change the two lines directly so we can merge. Everything else if perfect already.

scripts/src/main/resources/scripts/command/jmc
hohwille: Done.

scripts/src/main/resources/scripts/command/jmc
hohwille: Done.

tobka777 (approved):

hohwille:
I wrote into the issue instead of the PR:
> There is no jmc here: https://github.com/devonfw/ide-urls/tree/master/
We can not merge this without the download metadata being present.
https://github.com/devonfw/ide/pull/1251/files only added the JmcUrlUpdater without adding it so it also gets executed.

OK. So the change to add the updater is included in this PR. I now remember that you already told this in advance.
The problem with this approach is that we cannot properly test this before merging.
However, I already approved and will merge then but I tend to first release our bugfix version 2023.08.001 and then merge this PR so we have some time left to test it before the next release will be made.
Thanks again for your work. It all seems fine and I am quite confident that it will work 👍

hohwille:
@moritzLanger I wanted to merge this now but unfortunately we have this meanwhile:
> Conflicting files
url-updater/src/main/java/com/devonfw/tools/ide/url/updater/UpdateManager.java 

Can you merge & resolve this?

hohwille:
Replaced by PR #1313, hence closing.

Issue #1186, opened 2023-06-07
install.asciidoc file referenced by software.asciidoc is missing


jan-vcapgemini:
The install.asciidoc file is missing.

https://github.com/devonfw/ide/blob/43bc42f6e7cb44535dadd72d63d5ee2917185077/documentation/software.asciidoc?plain=1#L8

Can the link be removed?
What has to be described on the install page?

PR #1292, opened 2023-07-26
#1186 install.asciidoc file referenced by software.asciidoc is missing

CREITZ25:
#1186 install.asciidoc file referenced by software.asciidoc is missing


jan-vcapgemini (approved):

tobka777 (approved):
Thank you. That works!

Issue #1320, opened 2023-08-10
Documentation entirely broken


hohwille:
### Expected behavior

As a devonfw-ide user, I want to be able to read the offline documentation from the included PDF so that I can lookup some information without going to the online documentation. Also the online documentation only reflects the latest state and if I do have an older version, the documentation might be incorrect.

### Actual behavior

Since many releases the documentation is broken.
See e.g. https://repo1.maven.org/maven2/com/devonfw/tools/ide/devonfw-ide-doc/2023.08.001/devonfw-ide-doc-2023.08.001.pdf
Chapter `1.4` is the last one in the index, after that only asciidoc source-code is displayed in the PDF.
On 3. Juli 2023 I posted this in our chat but nobody cared:
> Any ideas what we have broken in our documentation?
```
[INFO] Copying 23 resources 
[INFO] asciidoctor: ERROR: ionic.asciidoc: line 20: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: jasypt.asciidoc: line 20: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: java.asciidoc: line 19: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: jenkins.asciidoc: line 18: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: kotlinc.asciidoc: line 15: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: kotlinc-native.asciidoc: line 15: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: tomcat.asciidoc: line 20: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: kubectl.asciidoc: line 28: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: lazydocker.asciidoc: line 25: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: mvn.asciidoc: line 25: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: ng.asciidoc: line 20: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: node.asciidoc: line 21: table missing leading separator; recovering automatically 
[INFO] asciidoctor: ERROR: npm.asciidoc: line 24: table missing leading separator; recovering automatically 
[...]

PR #1322, opened 2023-08-10
#1320: Fix offline generated documentation

alfeilex:
This PR fixes the missing table close in the intellij docs which lead to an error.

Related issue: https://github.com/devonfw/ide/issues/1320


hohwille (approved):
@alfeilex thanks for the fast fix. 👍

Issue #1330, opened 2023-08-14
GcViewer DonwloadUrl not Working


moritzLanger:
### Expected behavior

As a User, I want to be able to downlaod and use GCviewer 

### Actual behavior

the current gc viewer download url while working manually, does not download the .jar but a .zip containing meta data. This leads to the software not being able to be started and used.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1.  use devon gcviewer
2. Error message ` Error: Unable to access jarfile gcviewer*.jar` appears

### Affected version:

* OS: Windows/Linux/Mac?
* Browser: Chrome/Firefox/Safari?

PR #1331, opened 2023-08-14
#1330 GcViewer DonwloadUrl not Working

moritzLanger:
#1330 GcViewer DonwloadUrl not Working 

The URL added in #1134 only works, when manually downloading, when using with the ide tool, it downloads meta data and not the desired .jar. Therefore the "old" url must be used, despite being one that will be redirected.


CREITZ25 (approved):
nice

Issue #1299, opened 2023-07-31
gcViewer version change/update not working


moritzLanger:
### Expected behavior

As a user, I want to be able to upgrade and downgrade my gcViewer version so that i can use the desired versions.

### Actual behavior

The current [GCviewer Script ](https://github.com/devonfw/ide/blob/master/scripts/src/main/resources/scripts/command/gcviewer#L36)is not deleting the unused versions, leading to the first installed version being the one opened/used regarless of the version devon claims to be using.

### Steps to reproduce (bug) / Use Case of feature request (enhancement)

1.  install any version of gcViewer (`devon gcviewer setup`)
2.  use `devon gcviewer verstion set 1.XX` to change the version 
3. redo `devon gcviewer setup`
4. launching gcviwer and going to help revelas the diffrent version than `devon gcviewer version get`

### Comments/Hints:

<img width="389" alt="image" src="https://github.com/devonfw/ide/assets/44358321/40d2b5f9-1e07-4d41-b3c8-0ec163692706">

<img width="251" alt="MicrosoftTeams-image" src="https://github.com/devonfw/ide/assets/44358321/c8454ba9-a395-4d68-8e0c-6379c9d7c790">

### Affected version:

* OS: Windows/Linux/Mac

hohwille:
* https://github.com/devonfw/ide/blob/35770460cc08b8b5aba9c86170987a1ba5843a14/scripts/src/main/resources/scripts/command/gcviewer#L36
* https://github.com/devonfw/ide/blob/35770460cc08b8b5aba9c86170987a1ba5843a14/scripts/src/main/resources/scripts/command/gcviewer#L22

PR #1333, opened 2023-08-15
#1299 gcViewer version change/update not working 

moritzLanger:
#1299 
fixed gcviewer script, to delete the old gcviewer software folder before installing new version


scripts/src/main/resources/scripts/command/gcviewer
```diff
@@ -16,9 +16,16 @@ GCVIEWER_HOME="${DEVON_IDE_HOME}/software/gcviewer"
 TOOL_VERSION_COMMAND="-"
 # shellcheck source=scripts/commandlet-cli
 source "$(dirname "${0}")"/../commandlet-cli
+
 # gcviewer installation
 function doSetup() {
   doDevonCommand java setup silent
+  local current_version
+  current_version=$(doDevonCommand gcviewer version get | sed -n '2 p')
+  if [[ "${GCVIEWER_VERSION}" != "${current_version}" ]]
```
hohwille, line 25: avoid double brackets that may not work on MacOS due to old bash version.
Further consider the following cases:
* `GCVIEWER_VERSION` may not be defined.
* `GCVIEWER_VERSION` could be set to a version pattern (#893)

In both cases the existing installation would be deleted and the same version would be downloaded again.
However, after all this may still be the easiest fix as we otherwise would have to resolve the version to be installed duplicating some logic that is already done inside `doInstall`.

A smarter approach could also be to add a tweak to the if in line 32 that checks if there are multiple gcviewer JAR files now in `${GCVIEWER_HOME}` and delete all but the latest one (according to mtime - in case the mtime is pointing to the date+time when the file was downloaded/copied).

moritzLanger: i implemented this suggestion in the latest commit

hohwille (approved):
@moritzLanger Excellent solution with your file loop and the `-nt` comparison. 👍 
I fixed the indentation and initialized `temp` as local variable with empty value to prevent side-effects if the user may have set `temp="foo"` before your code gets invoked.
Ready for merge.

Issue #1325, opened 2023-08-10
Add GraalVM JDK17 and 20 versions


alfeilex:
In the `ide-urls` we currently use the version up to 22.3.3. Since Oracle handed over the Community Edition to OpenJDK, there are two new distributions that we should include in `ide-urls` for `graalvm`.

These versions may also be relevant if we use the Maven plugin to create GraalVM Native Image. I am not sure, but it could be that the latest versions of this plugin just works with the latest GraalVM JDK17 and 20 versions.

- [x] Add JDK17 and 20 versions to url-updater 

Versions:
- https://www.graalvm.org/downloads/
- https://www.oracle.com/java/technologies/downloads/#graalvmjava17-windows
- There are also the other platform download links

alfeilex:
Missing releases (the first and second release): 
https://www.graalvm.org/release-notes/

PR #1336, opened 2023-08-22
Add GraalVM JDK 17 and 20 versions

alfeilex:
This PR adds the new releases of GraalVM JDK versions to `ide-urls`

Related issue: #1325


ide/src/main/java/com/devonfw/tools/ide/url/updater/graalvm/GraalVmUrlUpdater.java
```diff
@@ -21,13 +22,20 @@ protected String getVersionPrefixToRemove() {
...
   @Override
   protected void addVersion(UrlVersion urlVersion) {
-
-    String baseUrl = "https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-${version}/graalvm-ce-java11-";
-    doAddVersion(urlVersion, baseUrl + "windows-amd64-${version}.zip", WINDOWS, X64);
-    doAddVersion(urlVersion, baseUrl + "linux-amd64-${version}.tar.gz", LINUX, X64);
-    doAddVersion(urlVersion, baseUrl + "darwin-amd64-${version}.tar.gz", MAC, X64);
-    doAddVersion(urlVersion, baseUrl + "darwin-aarch64-${version}.tar.gz", MAC, ARM64);
-
+    // if (urlVersion.getName().startWith("jdk")
```
alfeilex, line 25: What is the best condition to use here? Is not it easier to use the `urlVersion.getName()` method?

hohwille: Yes that would be better. You could also try something like `urlVersion.getStart().getLettersString()` to get the potential `jdk` prefix but the out-commented code is maybe the one most easy to understand by different developers.

alfeilex: Done

alfeilex:
I chose `community` and `oracle` to distinguish the editions

hohwille (approved):
@alfeilex thanks for your rework, now functionally correct 👍 
Ready for merge...
However, an optimization for the future could consider:
* the two non-abstract updaters should filter out the versions they are unable to process and that will only lead to download errors.
* Instead of reading the tagged versions twice from github we could follow a similar approach like we did for intellij and handle both editions in a single updater.

Issue #1308, opened 2023-08-04
Run missed out migrations automatically


hohwille:
As a devonfw-ide user, I want to be able to update my very old IDE installations without problems so that everything is working.

However, when my very old IDE installations are older than `2023.02.001` the migration scripts are not run as the feature to run them was only added in `2023.02.001`.
But there is a simple solution to add a "self-healing" feature:
If the file `$DEVON_IDE_HOME/.devon.software.version` is not present, the migration has been missed out and we can simply run `$DEVON_IDE_HOME/scripts/migration` to heal this.
This workaround can simply be added somewhere [here](https://github.com/devonfw/ide/blob/3e5813b3d57a0e3b062bc7725ad097cf2b7a3197/scripts/src/main/resources/scripts/environment-project#L8).
However, it has to be taken into account that this should not happen on initial setup (`devon ide setup`) that also will trigger the linked code-spot (in that case initially there is no `software` folder or `workspaces` and if the ide setup creates the `.devon.software.version` early enough before invoking `devon` recursive then all should work fine).

hohwille:
Suggested test:
* build a SNAPSHOT release with your changes (`mvn clean install`)
* extract the resulting `scripts/target/devonfw-ide-scripts-*.tgz` to a new empty folder (e.g. "C:\projects\test").
* run the setup (e.g. `setup.bat` on windows)
* Check all the logs and ensure that migrations did NOT run
* after this all worked and your IDE is setup correctly without running migrations then delete the `.devon.software.version` file from your IDE_HOME (e.g. "C:\projects\test\.devon.software.version").
* Then run a `devon` command with some arguments (e.g. `devon java setup`) and verify that now the migrations are run.

PR #1352, opened 2023-09-06
Fix/1308 trigger migration for old versions

jan-vcapgemini:
Addresses/Fixes: #1308

Implements:
* simple check for existing .devon.software.version and software/workspaces


alfeilex:
I tested your commit with following tasks:

- setup an older ide instance 
- run `devon ide update scripts`
- add your code to `scripts/environment-project`
- run a command starts with `devon <...>` or just `devon`
- the migration should be start

hohwille (approved):
@jan-vcapgemini thanks for your PR and the final perfect fix. 👍

Issue #1153, opened 2023-05-11
improve offline capability


hohwille:
As a devonfw-ide user, I want to be able to work offline so that I remain productive in environments with bad or no internet connection (e.g. on business trips).

Obviously, I can not setup new tools or upgrade tools when I am offline.
However, when all has been setup but some setting has changed, I cannot run tools like eclipse when I am offline and get blocked.

```
$ devon eclipse
Updating eclipse from version 2023-03 to version 2023‑03...
Success: run command mkdir
Starting installation of eclipse in version 2023‑03 to /d/projects/MusicDoc/software/eclipse
fatal: unable to access 'https://github.com/devonfw/ide-mirrors.git/': Could not resolve host: github.com

******** ATTENTION ********
Can not update git repository: /d/projects/MusicDoc/mirrors
See above error for details - check your network connectivity and retry.
We are sorry for the inconvenience. Please check the above errors, resolve them and try again.
```

I see two issues here:

1. why is he "updating" from version `2023-03` to version `2023‑03` as both versions seem identical?
2. if we are offline we should allow the user a fallback option to continue. My suggestion would be to ask for confirmation by default in such case and add a CLI option (`-o` or `--offline`) that can suppress the interactive confimation/question (one could also use batch mode but somehow I think our approach with that is kind of flawed. Do we want to ask all questions with `yes` or with `no` in batch mode? If there is a question like "failed to update download urls - maybe you are offline. Do you want to skip the update?" the answer might be different than in case of "your settings folder does not have a .git folder and seems corrupted. Do you want to delete it and do a fresh clone?". Ideally we could ask the operating system if we are offline or not but that seems to be a general problem that is not so easy as you would expect it.

hohwille:
For the record: my eclipse installation was still fully present and I could manually run `devon` and then `eclipse` to launch eclipse but first of all a user has to know such workaround and second this way you are losing features as we are doing more than just calling `eclipse` without args.

PR #1363, opened 2023-10-16
#1153 - improve offline capability

diiinesh:
Added function: checkNetworkStatus()

adjusted function: doInstall()
- specifically the if conditions where these functions are called:
1. doGitPullOrClone
2. doUpdateUrls
3. doDownload

If system is offline, skip these functions.


diiinesh:
Closed because of required clean up of fork

PR #1364, opened 2023-10-16
#1153 - improve offline capability

diiinesh:
Fixes #1363 
Added function: checkNetworkStatus()

adjusted function: doInstall()

    specifically the if conditions where these functions are called:

    doGitPullOrClone
    doUpdateUrls
    doDownload

If system is offline, skip these functions.

### DIFF
--- a/scripts/src/main/resources/scripts/functions
+++ b/scripts/src/main/resources/scripts/functions
@@ -1257,6 +1257,15 @@ function doGitPullOrClone() {
   fi
 }
 
+# checks network connectivity status
+function checkNetworkStatus() {
+  if curl -s --head --fail https://www.google.com > /dev/null; then
+    return 0 # Online
+  else
+    return 1 # Offline
+  fi
+}
+
 #  $1: name of software
 #  $2: version of software
 #  $3: optional silent flag ('silent' to suppress output if already up-to-date or empty for version output)
@@ -1279,6 +1288,10 @@ function doInstall() {
   local repository="${8}"
   local url="${9}"
   local target_filename="${10}"
+
+  checkNetworkStatus
+  local isOnline=$?
+
   if [ -z "${target_path}" ]
   then
     target_path="${DEVON_IDE_HOME}/software/${software}"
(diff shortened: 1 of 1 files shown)


jan-vcapgemini (approved):
Looks good to me.

scripts/src/main/resources/scripts/functions
```diff
@@ -1291,12 +1304,15 @@ function doInstall() {
     doFail "Missing software argument for doInstall: ${*}"
   fi
   local result=0
-  if [ "${url/.git/}" != "${url}" ]
+  if [ "${url/.git/}" != "${url}" ] && [ "$isOnline" = 0 ]
   then
     doGitPullOrClone "${target_path}" "${url}"
     return
   fi
```
hohwille, line 1311: You introduced a bug here:
if we have a git repo we want to pull or clone it and then return.
In case we are offline, you will now skip the return causing followup errors.

scripts/src/main/resources/scripts/functions
```diff
@@ -1381,7 +1397,7 @@ function doInstall() {
     install_path="${DEVON_SOFTWARE_PATH}/${repo}/${software}/${version}"
     version_file="${install_path}/.devon.software.version"
   fi
-  if [ ! -d "${install_path}" ] || [ "${install_path}" == "${target_path}" ]
+  if [ ! -d "${install_path}" ] || [ "${install_path}" == "${target_path}" ] && [ "$isOnline" = 0 ]
```
hohwille, line 1400: if a tool is not installed at all and I will run devon command to install it, with this change, the command will succeed but do nothing. This is not the expected behavior for the end-user.

hohwille (changes requested):
@diiinesh thanks for your PR. Your function `doCheckNetworkStatus` looks perfect and you also properly documented it.
However, for the integration of the checks we need to adjust. I added comments to explain why the current approach is not working correctly.
My suggestion would be:
* `doGitPullOrClone`: if repo is already there and pull fails, then add a `doCheckNetworkStatus` and if offline log a warning that the repo could not be updated because we are offline and return (continue flow)
* `doInstall`: if software is already installed (maybe in a different version), then add a `doCheckNetworkStatus` and if offline log a warning that we cannot install update because we are offline and return (if `${version}` was given per arg and matches `.ide.software.version` we should even suppress the check and the warning and then there is nothing to do anyways and also no problem if we are offline).

hohwille (approved):
@diiinesh thanks for your update. Looks sane to me now. 👍 
It is still a little different to the behavior of IDEasy but fine to be merged.

PR #1375, opened 2024-01-04
#1367: added fix to java version compare

### LINKED ISSUE #1367: Detection of latest Java version is buggy
```
$ devon java version set 21
21.0.1_12  21_35

$ devon java version set 21*
JAVA_VERSION=21* has been set in /d/projects/mmm/settings/devon.properties
To install that version call the following command:
devon java setup

$ devon java setup
Updating java from version 17.0.8_7 to version 21_35...
```

Vising https://adoptium.net/ confirms that this is a bug and `21.0.1_12` is newer than `21_35`.
Indeed an edge-case from the version compare function but with Java a very relevant one.
The problem is that semantics like `.` vs. `_` are actually undefined.
However, if both versions have an underscore (`_`) it seems kind of natural to first compare the prefix part before the underscore and only if equal then compare the suffix part so:
* `21.0.1` <=> `21` --> `21.0.1` > `21`
* only if it would be equal then compare `12` <=> `35` --> `12` < `35`.

A remaining question would be the comparison of `21.0.1.12` with `21_35` and IMHO there can not be a clear answer as it may depend on the project/product creating these versions.

I guess for Java the suffix starting with the `_` is some kind of build number.

MattesMrzik:
Fixes #1367. Added small clause to `doVersionCompare()` to make version like `21_35` smaller than `21.0.1_12`.

### DIFF
--- a/scripts/src/test/bash/test-version-compare
+++ b/scripts/src/test/bash/test-version-compare
@@ -59,5 +59,10 @@ doTestVersionCompare 3.0.0-beta17 '>' 3.0.0-beta11-SNAPSHOT
 doTestVersionCompare 3.0.0.11 '>' 3.0.0-beta11-SNAPSHOT
 doTestVersionCompare 2020.04.001 '>' 3.3.1
 doTestVersionCompare "11*" '>' "11u0"
+doTestVersionCompare 21.0.1_12 '>' 21_35
+doTestVersionCompare 21_35 '<' 21.0.1_12
+doTestVersionCompare 21_35 '<' 21
+doTestVersionCompare 21_35 '<' 21_36
+doTestVersionCompare 21_35 '<' 21_36.2
 
 exit "${exitcode}"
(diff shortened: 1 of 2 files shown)


scripts/src/test/bash/test-version-compare
```diff
@@ -59,5 +59,10 @@ doTestVersionCompare 3.0.0-beta17 '>' 3.0.0-beta11-SNAPSHOT
 doTestVersionCompare 3.0.0.11 '>' 3.0.0-beta11-SNAPSHOT
 doTestVersionCompare 2020.04.001 '>' 3.3.1
 doTestVersionCompare "11*" '>' "11u0"
+doTestVersionCompare 21.0.1_12 '>' 21_35
+doTestVersionCompare 21_35 '<' 21.0.1_12
+doTestVersionCompare 21_35 '<' 21
```
hohwille, line 64: I am unsure if this is correct. I assume not. However, it is an edge-case that we do not have. So why not merge and fix only if that ever turns out to be a problem, as I assume it will never ever happen.

hohwille (approved):
@MattesMrzik thanks for your PR and this solid fix. Great job👍 
Ready for merge.

PR #1393, opened 2024-11-13
Fix: Global arguments of find command should be set first

alfeilex:
This issue fixes: #1394

After that change:
https://github.com/devonfw/ide/commit/402fbad9c659ed2d7ff9a5c85bd9b820a444e874

the order of arguments was wrong. That fix reorders the `type d` argument after the global arguments like `-maxdepth` and 
`-mindepth`


hohwille (approved):
@alfeilex thank you so much for the quick fix 🥇

